PR #13, opened 2016-04-11
fix #12: build: provide BOB_STEP_NAME as environment variable

### LINKED ISSUE #12: buildStep doesn't know it's name
At the moment there is no possibility to get the complete step name (e.g app/module1/module2) in the build or package step. 
This can be useful, if you create module-wise-project files from bob...

rahu:


pym/bob/cmds/build.py
```diff
@@ -412,6 +412,7 @@ def _runShell(self, step, scriptName):
             stepEnv["PATH"] = ":".join(step.getPaths() + step.getSandbox().getPaths())
         stepEnv["LD_LIBRARY_PATH"] = ":".join(step.getLibraryPaths())
         stepEnv["BOB_CWD"] = step.getExecPath()
+        stepEnv["BOB_STEP_NAME"] = "/".join(self.__currentPackage.getStack())
```
jkloetzke: Referencing ``self.__currentPackage`` looks wrong, ``step.getPackage()`` is more reasonable. In any case the variable should be named 'BOB_PACKAGE_PATH' because that's what it is in the end. In BOB_STEP_NAME I would expect something like "checkout", "build" and "package".

I have some mixed feelings though because a common package might be built only once but is actually found under several package paths.

BTW, the Jenkins implementation is missing too.

rahu: This is needed to generate Eclipse Project Files. I used Makefile projects and call bob dev <module> instead of make. Now it's quiet easy to generate a project file for the application with all dependencies linked as shared folders. But the goal is to generate individual projectfiles for both, application(root recipe) and dependencies, because this is the historical workflow.... 

For dependencies project's I need to call bob dev $BOB_PACKAGE_PATH since I haven't found another way to get this information in the buildStep.
Since I don't need projectFiles on Jenkins I haven't added this there.

Maybe it's better to move this line to our customized NameFormatter since we know the disadvantages of using this variable. In this case I again need to modify the StepEnv from there. (-> #19)

PR #25, opened 2016-07-23
add (probably incomplete) CVS support

stefanreuther:
The eighties called and want their SCM back. Use:
```
     checkoutSCM:
        scm: cvs
        cvsroot: ...
        module: ...
        dir: ...
        rev: tag-or-branch
```

### DIFF
--- a/pym/bob/input.py
+++ b/pym/bob/input.py
@@ -603,6 +603,89 @@ fi
     def hasJenkinsPlugin(self):
         return True
 
+
+class CvsScm(BaseScm):
+    # Checkout using CVS
+    # - mandatory parameters: cvsroot, module
+    # - optional parameters: rev, dir (dir is required if there are multiple checkouts)
+    def __init__(self, spec):
+        super().__init__(spec)
+        self.__cvsroot = spec["cvsroot"]
+        self.__module = spec["module"]
+        self.__rev = spec.get("rev")
+        self.__dir = spec.get("dir", ".")
+
+    def resolveEnv(self, env):
+        # Resolve macros in CVS config
+        super().resolveEnv(env)
+        self.__cvsroot = env.substitute(self.__cvsroot, "cvs::cvsroot")
+        self.__module = env.substitute(self.__module, "cvs::module")
+        if self.__rev:
+            self.__rev = env.substitute(self.__rev, "cvs::rev")
+        if self.__dir:
+            self.__dir = env.substitute(self.__dir, "cvs::dir")
+
+    def asScript(self):
+        # If given a ":ssh:" cvsroot, translate that to CVS_RSH using ssh, and ":ext:"
+        # (some versions of CVS do that internally)
+        m = re.match('^:ssh:(.*)', self.__cvsroot)
+        if m:
+            prefix="CVS_RSH=ssh "
+            rootarg=":ext:" + m.group(1)
+        else:
+            prefix=""
+            rootarg=self.__cvsroot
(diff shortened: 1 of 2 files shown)


pym/bob/input.py
```diff
@@ -603,6 +603,72 @@ def isDeterministic(self):
...
+        self.__module = env.substitute(self.__module, "cvs::module")
+        if self.__rev:
+            self.__rev = env.substitute(self.__rev, "cvs::rev")
+        if self.__dir:
+            self.__dir = env.substitute(self.__dir, "cvs::dir")
+
+    def asScript(self):
+        revarg = "-r {rev}".format(rev=self.__rev) if self.__rev != None else "-A"
+        return """
+# Validate that we're talking to the correct repository. If not, ditch it.
```
jkloetzke: AFAICT this check is unnecessary. Bob will always check the digests that are returned by getDirectories() an will move reconfigured directories to an "attic" directory next to the workspace. The idea is that switching branches/tags in a working copy is hard to implement for any SCM. The safe choice is to move away offending directories and make a clean checkout with the new configuration. In release mode (bob build ...) Bob will always use a fresh directory anyway.

Therefore the checkout script may ignore any changes made to the recipe. The only thing that should be handled gracefully are manual changes in the working copy by the user without modifying the recipe.

jkloetzke:
Despite the inline comment it's looking good. Would you mind to add some documentation too in doc/manual/manual/configuration.rst in the "checkoutSCM" chapter?

pym/bob/input.py
stefanreuther: OK, didn't know about the "attic" feature yet.

I was mainly worrying about users having messed with the workspace directory in unexpected ways. In this case CVS will happily do a ``cvs up`` without knowing that it may not be the module we expect. But then, it's probably a case of "you broke it, keep both halves", and nuking the directory is not the nicest thing to do.

stefanreuther:
ok, updated a little.

I'm still wondering how to deal with ``CVS_RSH``. What works is to specify it in default.yaml and define it as a checkoutVar in each recipe that needs it. It would probably be cool if that were somehow automatic, but I'm not sure. My current application only uses locally-mounted repos.

stefanreuther:
To satisfy the coverage annoybot, here's a test case for CVS SCM: <https://github.com/stefanreuther/bob/tree/test_cvs_scm>. But I'm not sure whether that is a good idea.

jkloetzke:
That's looking good.

I'm thinking hard about the `CVS_RSH` variable and I have the feeling that this is something that needs to be specified separately for each CvsScm instance. I think most people will use SSH these days but that does not need to be true for all repositories throughout the recipes. Especially when pulling in some external dependencies.

I would either expose this as a separate property (e.g. `rsh: ssh`) or parse the `cvsroot` in an URL-like syntax (e.g. `cvsroot: ssh://foo@bar:...` -> `CVS_RSH=ssh cvs -d :ext:foo@bar:...`). I'm not sure if the latter is a good idea, though.

The black box test, as it is now, might blow up on the CI infrastructure because cvs is certainly missing. A better idea would be to either resort to some classic unit test for the CsvScm class (see test/*.py for some examples) or fully mock the cvs command by some script that checks the parameters in the test environment. The part to fix `bob clean` to accept `-D` is certainly worth merging, though it should be a separate commit.

jkloetzke:
Looks good to me. Could you squash everything into one commit before the merge?

stefanreuther:
Done. I had to use ``--force``, I hope that's ok :)

jkloetzke:
Sure. Thanks. :smile:

PR #26, opened 2016-08-23
[RFD] query-path and query-scm commands

stefanreuther:
Two new commands.

(a) `query-scm`: solves the problem that your SCM hook tells you "url xyz changed", and you want to know what to rebuild.

(b) `query-path`: allows you to find out where bob put the build result so you can postprocess that automatically (like: upload generated reports to a webserver). Disclaimer: I'm not sure I understood everything that makes up this command :-)

### DIFF
--- a/pym/bob/cmds/build.py
+++ b/pym/bob/cmds/build.py
@@ -14,7 +14,7 @@
 # You should have received a copy of the GNU General Public License
 # along with this program.  If not, see <http://www.gnu.org/licenses/>.
 
-from ..errors import BuildError
+from ..errors import BuildError, ParseError
 from ..input import RecipeSet, walkPackagePath
 from ..state import BobState
 from ..tty import colorize
@@ -288,16 +288,19 @@ esac
         return os.path.join("work", base.replace('::', os.sep), step.getLabel())
 
     @staticmethod
-    def releaseNamePersister(wrapFmt, persistent=True):
+    def releaseNamePersister(wrapFmt):
 
         def fmt(step, props):
             return BobState().getByNameDirectory(
                 wrapFmt(step, props),
                 asHexStr(step.getVariantId()),
-                step.isCheckoutStep(),
-                persistent)
+                step.isCheckoutStep())
 
         return fmt
+
+    @staticmethod
+    def releaseNameInterrogator(step, props):
+        return BobState().getExistingByNameDirectory(asHexStr(step.getVariantId()))
 
     @staticmethod
     def developNameFormatter(step, props):
@@ -338,7 +341,7 @@ esac
                     ret = os.path.join(baseDir, wrapFmt(step, props))
                 else:
                     ret = os.path.join("/bob", asHexStr(step.getVariantId()))
-            return os.path.join(ret, "workspace")
+            return os.path.join(ret, "workspace") if ret is not None else None
 
         return fmt
 
@@ -1039,9 +1042,7 @@ would get removed without acutally deleting that already.
     recipes.defineHook('releaseNameFormatter', LocalBuilder.releaseNameFormatter)
     recipes.parse()
 
-    nameFormatter = recipes.getHook('releaseNameFormatter')
-    nameFormatter = LocalBuilder.releaseNamePersister(nameFormatter, False)
-    nameFormatter = LocalBuilder.makeRunnable(nameFormatter)
+    nameFormatter = LocalBuilder.makeRunnable(LocalBuilder.releaseNameInterrogator)
 
     # collect all used paths (with and without sandboxing)
     usedPaths = set()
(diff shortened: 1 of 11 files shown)


jkloetzke:
I think the query-path command totally makes sense. OTOH I'm not sure that the query-scm command is the right approach for you, though I do see some value having it. If some URL changes you would normally map that onto a Jenkins server that will do the right thing.

Some remarks about the query-paths feature:
* The output format could be specified by some format string. That would be extensible and flexible. I would directly use the format string syntax. When specifying a single package the default could be "{path}". On multiple packages that could be "{pkg-name} {path}" instead.
* The implementation should be moved to the bob.cmds.build module as it relies on its internals anyway
* Currently works only for release mode. Support for dev mode should be implemented too. That implies a "--dev/--release" option.
* Getting the src and build directories would also be nice

About the query-scm feature:
* Supporting a format string would be equally nice, e.g. to print the "dir" attribute of a SCM too? Not sure if it makes sense.
* Maybe the query could be restricted to a sub-tree by optionally accepting a package path. (just an idea)
* Traverse by using getDirectDepSteps() instead of getAllDepSteps(). Otherwise forwarded dependencies (e.g. tools) will be visited over and over again.

It looks like it may be a good idea to factor out the "-D \<var\>=\<value\>", "-c \<file\>" and "--[no-]sandbox" handling into some common code. If you don't want to do it I'm ok with doing it myself after merging your stuff.

stefanreuther:
> OTOH I'm not sure that the query-scm command is the right approach for you, though I do see some value having it. If some URL changes you would normally map that onto a Jenkins server that will do the right thing.

I'm trying to get along without a Jenkins server; the idea was to hook a post-commit script directly to bob (ok, plus some debouncing logic). However, the more I think about it, the more problems I see with query-scm. (a) if someone does `svn rm http://a/b/c1; svn cp http://a/b/c2 http://a/b/c1`, that would report a change on `http://a/b/c1`, but not be registered by `bob query-scm` for a project that checks out `http://a/b/c1/trunk`. (b) if `bob query-scm` reports a change in module `a/b/c/d`, I'd have to trigger a rebuild on `a`. But then I could as well just have a cron-job rebuild everything every five minutes and let bob figure out what changed.

> The output format could be specified by some format string.

I tried to stay minimalist here, for easy scriptability: `a=$(bob query-path --show=path my-package)` to get a single path, `bob query-path --show=name $(bob ls)` to ask what has already been built. Using a format string would be an option, but using a different output format for the special case of one argument would not (for me).

The other three comments (move to build module, dev mode, src/build directories) make total sense. Same thing goes for `-D` processing; I think we already discussed that earlier.

jkloetzke:
> I'm trying to get along without a Jenkins server; the idea was to hook a post-commit script directly to bob (ok, plus some debouncing logic). However, the more I think about it, the more problems I see with query-scm. (a) if someone does svn rm http://a/b/c1; svn cp http://a/b/c2 http://a/b/c1, that would report a change on http://a/b/c1, but not be registered by bob query-scm for a project that checks out http://a/b/c1/trunk.

I think you can only trigger per repository. This may be a bit coarse in case of SVN but the check should be cheap if nothing has changed.

> (b) if bob query-scm reports a change in module a/b/c/d, I'd have to trigger a rebuild on a. But then I could as well just have a cron-job rebuild everything every five minutes and let bob figure out what changed.

Maybe a better approach would be to extend the "build only" switch. Something like "bob build -B https://... -B git@github.com:foo/bar.git root/package". The idea of "-B \<url\>" is that Bob will treat all packages in build only mode except the ones where these URLs are matching. In that sense the existing "-b" switch is the same as "-B" without URL. That should come close to your use case, right? Basically this is what a Jenkins would do in case it gets the SCM trigger...

stefanreuther:
Reworked according to discussion: "query-path" now moved to the build module and given a format string. Removed "query-scm" for now because it's hard and I don't need it anymore.

Open points: (a) I added "--develop"/"--release" but these do not seem to have any effect? (b) My format string ignores the format_spec, conversion parameters; my Python-fu is not good enough to make something useful of them. Right now, a missing directory causes the line to not be shown at all; it would make sense to have some kind of "show me an empty field instead" flag.

pym/bob/cmds/build.py
```diff
@@ -943,3 +943,103 @@ def doClean(argv, bobRoot):
...
+    recipes.parse()
+
+    # Find the name formatter
+    if args.dev:
+        nameFormatter = recipes.getHook('developNameFormatter')
+        developPersister = recipes.getHook('developNamePersister')
+        nameFormatter = developPersister(nameFormatter)
+    else:
+        nameFormatter = recipes.getHook('releaseNameFormatter')
+        nameFormatter = LocalBuilder.releaseNamePersister(nameFormatter)
```
jkloetzke: Here you should actually use a completely separate function. The `releaseNameFormatter` hook is not needed as we just want to query the persistent state in the release case. The key is to replace LocalBuilder.releaseNamePersister() by something along these lines (not tested):

```python
    @staticmethod
    def releaseNameInterrogator(step, props):
        return BobState().getExistingByNameDirectory(asHexStr(step.getVariantId()))
```

That way the normal path formatting machine should be usable.

pym/bob/cmds/build.py
```diff
@@ -943,3 +943,103 @@ def doClean(argv, bobRoot):
...
+        showPackage = True
+        packageText = ''
+        for (text, var, spec, conversion) in Formatter().parse(args.f):
+            packageText = packageText + text
+            if var == None:
+                pass
+            if var == 'name':
+                packageText = packageText + p
+            elif var == 'src':
+                dir = state.getByNameDirectoryMaybe(asHexStr(package.getCheckoutStep().getVariantId()))
```
jkloetzke: Just query the package directly:

```python
if package.getCheckoutStep().isValid():
    packageText = packageText + package.getCheckoutStep().getWorkspacePath()
else:
    showPackage = False
```

You have to check first via isValid(), otherwise you would get a deliberately non-exiting path.

pym/bob/state.py
```diff
@@ -91,6 +91,12 @@ def getByNameDirectory(self, baseDir, digest, persistent):
                 self.__save()
             return res
 
+    def getByNameDirectoryMaybe(self, digest):
```
stefanreuther: Nobody likes my "Maybe" function names :'-(

pym/bob/cmds/build.py
stefanreuther: It does not seem to work as outlined in this (and the previous) comment. I added a version that works as far as I understand.

* developer mode does not (seem to) persist file names. Thus, use the regular name builder and use that. Since we only want to show existing path names, verify that.
* for release mode, query state as before, but stub out the name builder.

jkloetzke:
I've pushed a branch to my fork (https://github.com/jkloetzke/bob/tree/query_scm) with two fixes. AFAICS it works as expected in develop as well as in release mode. I would propose to squash the commits to build.py but keep the changes to state.py as separate commits, though.

Looks like it could be merged in the near future...

stefanreuther:
Final version?

Merged your changes, squashed everything together and added a black-box test which promptly found a typo in the original code.

pym/bob/cmds/build.py
```diff
@@ -1068,3 +1069,107 @@ def doClean(argv, bobRoot):
...
+        help="Use config File")
+
+    group = parser.add_mutually_exclusive_group()
+    group.add_argument('--sandbox', action='store_true', help="Enable sandboxing")
+    group.add_argument('--no-sandbox', action='store_false', dest='sandbox', help="Disable sandboxing")
+    parser.set_defaults(sandbox=None)
+
+    group = parser.add_mutually_exclusive_group()
+    group.add_argument('--develop', action='store_true',  dest='dev', help="Use developer mode")
+    group.add_argument('--release', action='store_false', dest='dev', help="Use release mode", default=True)
```
jkloetzke: I guess your intention was that developer mode is the default, right? It is also my preference. Unfortunately argparse applies the default only when a variable is defined for the first time. The way it is now the release mode is the default. So even though it looks silly the `default=True` must be one line above.

pym/bob/cmds/build.py
```diff
@@ -1068,3 +1069,107 @@ def doClean(argv, bobRoot):
         if not args.dry_run:
             removePath(d)
 
+def doQueryPath(argv, bobRoot):
+    # Local imports
+    from string import Formatter
+
+    # Configure the parser
+    parser = argparse.ArgumentParser(prog="bob query-path",
+                                     description="Lists all packages and their last build directory. If a package has never been built, no output is shown for that package.")
```
jkloetzke: It would really be nice to have some short paragraph in the help text explaining the available output formats. Otherwise nobody knows the options without reading the source code. There are some expamples in `doClean()` and `doQuerySCM()` how to that.

pym/bob/cmds/build.py
```diff
@@ -1068,3 +1069,107 @@ def doClean(argv, bobRoot):
...
+    parser.add_argument('packages', metavar='PACKAGE', type=str, nargs='+',
+        help="(Sub-)package to build")
```
jkloetzke: s/build/query/

pym/bob/cmds/build.py
```diff
@@ -1068,3 +1069,107 @@ def doClean(argv, bobRoot):
...
+            elif var == 'name':
+                state.appendText(p)
+            elif var == 'src':
+                state.appendStep(package.getCheckoutStep())
+            elif var == 'build':
+                state.appendStep(package.getBuildStep())
+            elif var == 'dist':
+                state.appendStep(package.getPackageStep())
+            else:
+                raise KeyError("Unknown field '{" + var + "}'")
```
jkloetzke: You should raise a ParseError here. Only exceptions derived from bob.errors.BobError are caught gracefully:

```
jkloetzke@tuxedo:~/src/bob/playground/recipes$ bob query-path -f "{name} {foobar}" foo/bar 
  File "/home/jkloetzke/src/bob/bob/pym/bob/scripts.py", line 103, in bob
  File "/home/jkloetzke/src/bob/bob/pym/bob/scripts.py", line 59, in __querypath
    doQueryPath(*args, **kwargs)
  File "/home/jkloetzke/src/bob/bob/pym/bob/cmds/build.py", line 1172, in doQueryPath
    raise KeyError("Unknown field '{" + var + "}'")
KeyError: "Unknown field '{foobar}'"
```

pym/bob/cmds/build.py
stefanreuther: fixed

pym/bob/cmds/build.py
stefanreuther: fixed

jkloetzke:
:+1: Thanks for the effort. :smile:

PR #28, opened 2016-09-18
bob project: added project command to bob

rahu:
This new command basicly does nothing. It registers a new plugin
hook called 'projectGenerator'. This hook is called with 3 arguments
(packages, destination, package) making it possible to write individual plugins
for project file generation of different IDE's.

As an example there is a qt-creator generator added.

I do not know much about qt-creator project files so maybe the template can be shorten a lot. :)


jkloetzke:
Great idea. I think there could even be some standard support in Bob directly in the long run for major IDEs. Some things come immediately to my mind:

* It should be possible to support multiple project types. I guess something like "bob project qt-creator ..." where the first argument names the project type.
* pass extra args to the project creator plugin for further option processing

rahu:
Hi Jan,
please have a look on the updated MR.

bob project:
- I splitted argument parsing. At first the plugin name is parsed to be able to support multiple project generators. All remaining arguments are passed to the plugin, so each plugin can define it's own args. I lost the help from plugin args. If I call `bob project qt-project root -h` bob project help text is shown. Any ideas?

qt-generator-plugin:
- I defined  some common used sets of bob features ( -f, -b, -n,  -nb). You should be able to select them in QTcreator
- after `bob dev` a second `bob project` command is executed to update .files and .includes file.

TODO's: 
- look for binaries in the dist folder and add them as debug targets.
- ???

contrib/plugins/qt-creator-project.py
```diff
@@ -0,0 +1,231 @@
...
+from bob.errors import ParseError
+from bob.input import PluginState, PluginProperty, RecipeSet
+from collections import OrderedDict
+from bob.cmds.build import LocalBuilder
+
+def getCheckOutDirs(packages, dirs):
+    # first scan all packages recursive for their dependencies and build a list of checkout dirs
+    for p in packages:
+        deps = p.getDirectDepSteps()
+        path = p.getCheckoutStep().getWorkspacePath()
```
jkloetzke: Use `p.getCheckoutStep().isValid()` here instead of looking for 'invalid' in the path.

contrib/plugins/qt-creator-project.py
```diff
@@ -0,0 +1,231 @@
...
+       return -1
+    generateQtProject(roots, args.dev, args.destination, args.package, args.update)
+
+
+
+manifest = {
+    'apiVersion' : "0.1",
+    'hooks' : {
+        'qt-project' : projectGenerator,
+    }
```
jkloetzke: Hooks are for well-known internal extension points. In this case I would introduce a new section:

```python
manifest = {
  'apiVersion' : "0.3",
  'projectGenerators' : {
    'qt-project' : projectGenerator,
  }
}
```

This way the plugin could also provide more generators.

contrib/plugins/qt-creator-project.py
```diff
@@ -0,0 +1,231 @@
...
+            with open(templateFile) as template:
+                for line in template:
+                    if 'plugin:id' in line:
+                        sharedFile.write(line.replace("<!-- plugin:id -->", projectid ))
+                    elif 'plugin:buildstep' in line:
+                        addBuildSteps(sharedFile, 'dev' if devMode else 'build', project, destination)
+                    else:
+                        sharedFile.write(line)
+
+def projectGenerator(argv, bobRoot):
```
jkloetzke: The signature should look like: `generator(package, argv)`.

The parsing of the recipes and the package selection must be done outside of the plugin. They are not part of the stable plugin API. It would be done by all generators anyway. Just spill the extra aguments into `argv` here (see below).

If `bobRoot` will be used in the future is also not clear. Better not pass it.

contrib/plugins/qt-creator-project.py
```diff
@@ -0,0 +1,231 @@
...
+    parser = argparse.ArgumentParser(prog="bob project", description='Generate Project Files')
+    parser.add_argument('package', type=str, nargs='?',
+                        help="Sub-package to start listing from")
+    devBuild = parser.add_mutually_exclusive_group()
+    devBuild.add_argument('-d', '--dev', default=True, action='store_true',
+                        help="Use bob dev mode in project files")
+    devBuild.add_argument('-b', '--build', default=False, action='store_false', dest='dev',
```
jkloetzke: Release mode makes no sense for IDE projects. If the recipes are updated the location of the files and results might change. I don't think this makes sense in an IDE and will be quite surprising.

contrib/plugins/qt-creator-project.py
```diff
@@ -0,0 +1,231 @@
...
+                        help="Use bob build in project files")
+    parser.add_argument('-u', '--update', default=False, action='store_false',
+                        help="Update project files (.files, .includes)")
+    parser.add_argument('--destination', metavar="DEST",
+        help="Destination of project files")
+
+    group = parser.add_mutually_exclusive_group()
+    group.add_argument('--sandbox', action='store_true', default=True,
```
jkloetzke: Sandboxing will not work with an IDE. The debug information will point to files in `/bob/...` which the IDE cannot find.

pym/bob/cmds/misc.py
```diff
@@ -18,6 +18,22 @@
...
 import sys
 
+
+def doProject(argv, bobRoot):
+    def createProject(packages):
+        print ("you need to register a plugin.")
+
+    parser = argparse.ArgumentParser(prog="bob project", description='Generate Project Files')
+    parser.add_argument('project', type=str, nargs=1,
+                        help="plugin to use for project file generation")
```
jkloetzke: You should get the package name here. Anything after that can go to the plugin:

```python
    parser.add_argument('packages', metavar='PACKAGE',
        help="(Sub-)package to build")
    parser.add_argument('args', nargs=argparse.REMAINDER,
                        help="Arguments for project generator")
args = parser.parse_args(argv)
```

Then pass `args.args` to the plugin.

contrib/plugins/qt-creator-project.py
rahu: With this signature I'll need a convention like package[0] is the root-package and all other packages are dependencies?

contrib/plugins/qt-creator-project.py
rahu: :+1:  but I'd like to use `generators` instead of `projectGenerators`, thinking of a generator to generate tags of your actual used sources or something like this.

rahu: and maybe change `bob project` to `bob generator` ??

contrib/plugins/qt-creator-project.py
jkloetzke: Hmm, AFAIU the "project" command will take one package. What get's passed to the plugin is the `bob.input.Package`. This is the root of the project and the dependencies are traversed from there by `getDirectDepSteps`...

contrib/plugins/qt-creator-project.py
jkloetzke: Could be an idea though I don't have a good idea what could be another use case.

rahu:
I did some refactoring. Is this matching your comments? 

Defines (-D..), `--download=`, config files, `-c ` can be passed as additional arguments to `bob project`. (still todo). I'm unsure if I should introduce a new keyword like `--dev_args='-D.. -D.. --download=no'` for this or just lead over all remaining arguments to `bob dev`? 

But at the moment  there are 5 different build profiles -> the user has to change 10 build steps manually in QtCreator. This is not much comfortable and will result in strange behavior if he misses one. Maybe I should generate a shell script with all these user-settings and use this in the qt-creator-build-step to pass only bob dev args? This script would be the singe point of information and the user can easily add/remove settings?

jkloetzke:
Not quite yet. :-D

The problem is that the plugin still uses internal APIs ob Bob. I will make a patch how I think it should look like...

jkloetzke:
I have pushed https://github.com/jkloetzke/bob/tree/project that should come close. Worked for me (tm).

rahu:
Starting from your work I added two more features:
- the plugin now generates a shell script calling `bob dev ` and `bob project`. The advantage of this is a single point where you can change defines, configs and env-vars. All build-configs in qt-creator are using the script and add dev-args (-b, -n, ..)
The build script is added to .files File so you can see it in QTCreator.
- the plugin generates run targets in QT-creator. Therefor it looks for x86 executables in packageDir.  To be able to do this `bob project` needs to do a `bob dev` first, because binaries have to be build before generator can pick them up. For file type identification I'm using [python-magic](https://github.com/ahupp/python-magic) 
   ` pip3 install python-magic `

contrib/bash-completion
```diff
@@ -237,6 +237,11 @@ __bob_jenkins()
    __bob_subcommands "add export ls prune push rm set-url set-options" "jenkins"
 }
 
+__bob_project()
+{
+   __bob_complete_path "-d --dev -b --build --destination --sandbox --no-sandbox"
```
jkloetzke: These options do not fit anymore. Also, the command first has to complete the "generator" and then a "package". For that we'll need a "bob project --list" to get all available generators.

contrib/plugins/qt-creator-project.py
```diff
@@ -0,0 +1,229 @@
...
+
+    dirs = []
+    getCheckOutDirs(package, dirs)
+    projectDir = project.replace('/','_')
+    if projectDir[-1:] == '_':
+        projectDir =projectDir[:-1]
+
+
+    if not destination:
+        destination = os.path.join(os.getcwd(), projectDir)
```
jkloetzke: Maybe use `os.path.join(os.getcwd(), "project", projectDir)` to group them by default in a sub directory?

I would also suggest to use just `package.getName()` for `projectDir`. Right now the name will get quite long if you build something deep down in the hierarchy.

rahu: I agree with the first point. Maybe not use "project" but "qt-project" or "project/qt-project". Otherwise we'll probably mix up project files.

For the second: If I use `package.getName()` for `projectDir` I'll lose the possibility to have the same package with different settings open in parallel. This will happen in case you have a app cross-compiled for the target `bob project target/whatever/stuff/my_app` and also for the host `bob project host/whatever/stuff/my_app`. On the other hand to long project name does not disturb me. A good tradeoff is maybe another parameter e.g. `bob project --name=host_myapp`?

rahu:
- [x] added docu
- [x] added test
- [x] moved to projects-dir & added  `--name`
- [x] cleanup

pym/bob/cmds/build.py
```diff
@@ -906,6 +907,97 @@ def doDevelop(argv, bobRoot):
...
+
+    if args.list:
+        for g in generators:
+            print(g)
+        return 0
+    else:
+        if not args.package or not args.projectGenerator:
+            print("bob project: error: the following arguments are required: projectGenerator, package, args")
+            print(args.package)
+            print(args.projectGenerator)
```
jkloetzke: I guess these are debug leftovers...

pym/bob/cmds/build.py
```diff
@@ -906,6 +907,97 @@ def doDevelop(argv, bobRoot):
...
+        extra.append(e)
+    if args.preserve_env: extra.append('-E')
+
+    package = walkPackagePath(rootPackages, args.package)
+
+    # execute a bob dev with the extra arguments to build all executables.
+    # This makes it possible for the plugin to collect them and generate some runTargets.
+    if args.execute_prebuild:
+        devArgs = extra
+        extra.append(args.package)
```
jkloetzke: This accidentally changes `extra` that is passed to the generator below. I guess you wanted to do the following:

```python
devArgs = extra.copy()
devArgs.append(args.package)
```

rahu: Ok - fixed these points .

pym/bob/cmds/build.py
rahu: ..removed ... rebase can be nasty sometimes ... :-/

jkloetzke:
:+1: Thanks.

PR #29, opened 2016-09-22
debug bob: added debug main script

rahu:
With this it's possible to debug bob in a IDE like the PhyDev Eclipse plugin.
It needs to be in a separate package to get rid of the relative path imports.

Phydev debug configuration setup:
 - Main Module: debug.py
 - Workspace:   recipe repo, e.g. sandbox
 - Arguments:   bob args, e.g. dev vexpress

### DIFF
--- a/pym/debug.py
+++ b/pym/debug.py
@@ -1,0 +1,32 @@
+# Bob build tool
+#
+# This program is free software: you can redistribute it and/or modify
+# it under the terms of the GNU General Public License as published by
+# the Free Software Foundation, either version 3 of the License, or
+# (at your option) any later version.
+#
+# This program is distributed in the hope that it will be useful,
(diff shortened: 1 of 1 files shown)


pym/debug.py
```diff
@@ -0,0 +1,6 @@
+from bob.scripts import bob
```
jkloetzke: Trailing white space

pym/debug.py
```diff
@@ -0,0 +1,6 @@
+from bob.scripts import bob 
+import sys,os
+
+if __name__ == '__main__':
+    rootDir = os.getcwd()
+    bob(rootDir)
```
jkloetzke: No newline at end of file

rahu:
updated: added comments + newline

PR #33, opened 2016-09-27
Add boolean funcs

kolewu:
Add `or` and `and` to string substitution functions.


jkloetzke:
:+1:

Would you mind adding some documentation for the new functions in doc/manual/manual/configuration.rst in the "String substitution" chapter? That would be perfect.

jkloetzke:
Thank you.

PR #39, opened 2016-10-06
test: run coverage only from the current (v)env

### LINKED ISSUE #36: blackbox test not working inside venv
When starting `test/run-tests.sh` from inside a venv and the system lacks some dependencies, the test fails.  This looks like the test is somehow not started inside the venv.

```
pip list | grep python-magic
```
is empty. Then entering the venv
```
source /.../pyvenv/bob/bin/activate
pip list | grep python-magic
```
returns the package name and version as expected. Now running `test/run_tests.sh` gives this error:
```
Run test/generator
  File "/.../bob/pym/bob/scripts.py", line 98, in bob
  File "/.../bob/pym/bob/scripts.py", line 47, in __project
    doProject(*args, **kwargs)
  File "/.../bob/pym/bob/cmds/build.py", line 956, in doProject
    from ..generators.QtCreatorGenerator import qtProjectGenerator
  File "/.../bob/pym/bob/generators/QtCreatorGenerator.py", line 22, in <module>
    import magic
ImportError: No module named 'magic'
--- /dev/fd/63	2016-10-05 18:38:58.416804654 +0200
+++ output-plugin.txt	2016-10-05 18:19:31.212152069 +0200
@@ -0,0 +1,3 @@
+PLUGIN: root
+PLUGIN: []
+PLUGIN: ['--download=no']
test/generator failed
```

kolewu:
If running the tests inside a venv but coverage is only installed
outside, it was used and thus cannot import packages that are only
installed inside the venv.

This patch tests, if coverage is installed inside the current (v)env and
only uses this and not the outside one.

fix #36

### DIFF
--- a/test/run-tests.sh
+++ b/test/run-tests.sh
@@ -4,8 +4,16 @@ cd "${0%/*}/.."
 export PYTHONPATH="${PWD}/pym"
 BOB_ROOT="$PWD"
 
+USE_COVERAGE=0
 if type -fp coverage3 >/dev/null; then
-	RUN="coverage3 run --source $PWD/pym  --parallel-mode"
+    # make sure coverage is installed in the current environment
+    if python3 -c "import coverage" 2>/dev/null; then
+	    RUN="coverage3 run --source $PWD/pym  --parallel-mode"
+        USE_COVERAGE=1
+    else
+        RUN=python3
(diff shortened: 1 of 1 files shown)


jkloetzke:
Good idea to catch this case this way. Thanks.

PR #40, opened 2016-10-06
test: suppress confusing logmessage

kolewu:
Without this change the test output is disturbed by a confusing log
message. Since we expect it, we test for it and suppress it in doing so.

### DIFF
--- a/test/test_utils_dirhasher.py
+++ b/test/test_utils_dirhasher.py
@@ -20,6 +20,7 @@ from unittest.mock import MagicMock, mock_open, patch
 import binascii
 
 import os
+import sys
 from bob.utils import hashFile, hashDirectory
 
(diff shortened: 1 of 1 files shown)


jkloetzke:
Cool, I wasn't aware that one could even catch log message in the unittest module.

Could you please squash the commits and describe in the commit message why the test for 3.4 is done? It's better to not have intermediate partially broken commits...

kolewu:
Squashed.

Is there a reason you use `assert` and not the assert-functions from `unittest`?  I don't know whether this makes a difference technically -- if not for the convenience of some of the assert-functions.

jkloetzke:
Thanks.

Actually there is no good reason. I never bothered to dive into the unit testing framework and the assert statement was sufficient so far. :-)

PR #41, opened 2016-10-09
bob project: add eclipseProjectGenerator

rhubert:
This adds Eclipse CDT support to Bob. All dependencies are added as virtual folders
pointing to the src-directory of the package. For all x86 executables launch configs
will be generated.

What do you think? Shall this also be a build-in generator?

### DIFF
--- a/pym/bob/cmds/build.py
+++ b/pym/bob/cmds/build.py
@@ -954,7 +954,8 @@ def doProject(argv, bobRoot):
     touch(sorted(rootPackages.values(), key=lambda p: p.getName()))
 
     from ..generators.QtCreatorGenerator import qtProjectGenerator
-    generators = { 'qt-creator' : qtProjectGenerator }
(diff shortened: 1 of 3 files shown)


jkloetzke:
Do you have anything to add or can I merge?

rhubert:
Not sure. During testing I hit https://bugs.eclipse.org/bugs/show_bug.cgi?id=400073 (boost)
As a workaround it is be possible to exclude linked folders from build. They won't get indexed in this case but you can work with the rest. 

It the moment you have to be quicker than the indexer to exclude big folders. I'm thinking of a --exclude=<package> argument for eclipseGenerator. What's you opinion?

jkloetzke:
The "--exclude" options sounds like a viable workaround. Maybe let it take package names (with globbing)?

rhubert:
--exclude added. You can use regex, e.g. --exclude .*boost.* . Now it's ready to merge from my side.

jkloetzke:
:+1:

PR #42, opened 2016-10-15
add alias property

rhubert:
With this you can use a alias to reduce typing effort for deep packages like a/b/c/d. This alias must be defined in default.yaml (or a file included from there) and can be used for all bob cmds.

### DIFF
--- a/pym/bob/input.py
+++ b/pym/bob/input.py
@@ -1504,6 +1504,9 @@ class Package(object):
         """Name of the package"""
         return self.__name
 
+    def setName(self, name):
+        self.__name = name
+
     def getStack(self):
         """Returns the recipe processing stack leading to this package.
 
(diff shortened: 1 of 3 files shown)


pym/bob/input.py
```diff
@@ -2537,6 +2545,15 @@ def generatePackages(self, nameFormatter, envOverrides={}, sandboxEnabled=False)
                 except ParseError as e:
                     e.pushFrame(root.getPackageName())
                     raise e
+            for alias in self.__packageAliasList:
+                realName = self.__packageAliasList[alias]
+                try:
+                    p = walkPackagePath(result, realName)
+                    p = copy.copy(p)
+                    p.setName(alias)
```
jkloetzke: This is where the internal structure falls apart. The package name will not match the recipe name. Also the Steps back-link to their Package and this is broken too. All in all I think this is the wrong approach.

rhubert:
not only the ls command. also 'project' and maybe 'query-*' ...

PR #45, opened 2016-10-17
bob project: add --resume switch

### LINKED ISSUE #44: bob project --resume
1) run bob project without -n to let bob execute bob dev
2) if there is a error you'll get  a message: `You may resume at this point with '--resume' after fixing the error.`

But there is no --resume arg available for bob project.

rhubert:
Bob project runs 'bob dev'. Since this step can fail it should be possible to
use the --resume argument for bob project also.

Fixes #44

### DIFF
--- a/pym/bob/cmds/build.py
+++ b/pym/bob/cmds/build.py
@@ -926,6 +926,8 @@ def doProject(argv, bobRoot):
         help="Preserve environment variable")
     parser.add_argument('-E', dest="preserve_env", default=False, action='store_true',
         help="Preserve whole environment")
+    parser.add_argument('--resume', default=False, action='store_true',
+        help="Resume build where it was previously interrupted")
     parser.add_argument('-n', dest="execute_prebuild", default=True, action='store_false',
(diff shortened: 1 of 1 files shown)


jkloetzke:
I think this is wrong. That way you will always build via '--resume' if the project file was generated with that switch. This should only be added to the `devArgs` some lines below.

I'm not even sure if it is a good idea to expose this switch as a profile in the standard qt generator. You can easily shoot yourself in the foot when you activate this profile because Bob will never examine a package again once it was built.

rhubert:
You are right. I'll move it to the devArgs. 

I don't want to add another profile for the resume build. Is there a special return value for 'build failed but can be resumed'? If so we can use the to add the `--resume` to the next bob dev inside the buildme file.

jkloetzke:
No, there is no special return value. With "--resume" Bob will just skip all steps that were successful in previous run.

rhubert:
OK - I think using the --resume after a failed build from qtCreator this is a optimization for the future. I modified PR to enable --resume for bob project's build step.

PR #47, opened 2016-10-17
input: remove shelve file

### LINKED ISSUE #46: cache db in empty resp. invalid directories
I accidentally executed bob in the recipes directory and got a cache-db afterwards (which shouldn't be created in that error case):

martin@kriegel ~/mib3/tsd-recipes-mib3/recipes (git)-[master] % bob dev zr3
Parse error: No recipes directory found.
martin@kriegel ~/mib3/tsd-recipes-mib3/recipes (git)-[master] % git status
                                                                                                                                                          
Auf Branch master
Ihr Branch ist auf dem selben Stand wie 'origin/master'.
Unbeobachtete Dateien:
  (benutzen Sie "git add <Datei>...", um die Änderungen zum Commit vorzumerken)

        .bob-cache.shelve.db

nichts zum Commit vorgemerkt, aber es gibt unbeobachtete Dateien (benutzen Sie "git add" zum Beobachten)
martin@kriegel ~/mib3/tsd-recipes-mib3/recipes (git)-[master] %

rhubert:
If bob is executed in a directory without recipes it should leave this
directory without any leftovers.

Fixes #46

### DIFF
--- a/pym/bob/input.py
+++ b/pym/bob/input.py
@@ -2346,6 +2346,8 @@ class RecipeSet:
             return default
 
     def parse(self):
+        if not os.path.isdir("recipes"):
+            raise ParseError("No recipes directory found.")
         self.__cache.open()
         try:
             self.__parse()
@@ -2372,9 +2374,6 @@ class RecipeSet:
         self.__parseUserConfig("default.yaml")
 
         # finally parse recipes
-        if not os.path.isdir("recipes"):
-            raise ParseError("No recipes directory found.")
-
         for root, dirnames, filenames in os.walk('classes'):
             for path in fnmatch.filter(filenames, "*.yaml"):
                 try:


pym/bob/input.py
```diff
@@ -2373,6 +2373,10 @@ def __parse(self):
 
         # finally parse recipes
         if not os.path.isdir("recipes"):
+            self.__cache.close()
```
jkloetzke: No. That way YamlCache.close() is called twice because it is guarded by a try/finally statement in parse() above! Also, instead of cleaning up the mess left behind what about moving the test just above self.__cache.open() on line 2349? That way the file would not be created in the first place...

rhubert: Moving the test above self.__cache.open() would also result in closing a file never opened?

jkloetzke: No, the exception would be thrown before the try/finally is entered. I mean by moving

```python
        # finally parse recipes
        if not os.path.isdir("recipes"):
            raise ParseError("No recipes directory found.")
```

above `self.__cache.open()` will not create the file at all.

rhubert: What about moving this test to Line 2196?
```
    def __init__(self):
        if not os.path.isdir("recipes"):
            raise ParseError("No recipes directory found.")
```

to be safe(er) on futher addings?

pym/bob/input.py
```diff
@@ -2544,6 +2541,8 @@ def generatePackages(self, nameFormatter, envOverrides={}, sandboxEnabled=False)
 
 class YamlCache:
     def open(self):
+        if not os.path.isdir("recipes"):
```
jkloetzke: This is actually the wrong location. The test should be in RecipeSet.parse() before calling `self.__cache.open()`.

PR #50, opened 2016-10-18
project: adapt to different magic modules

stefanreuther:
There are two "magic" modules with similar functionality. Find out which one
we got and adapt.

Discussion: http://stackoverflow.com/questions/25286176/how-to-use-python-magic-5-19-1

Debian has the "wrong one" as python3-magic, causing the blackbox tests to bomb out with

>     Run test/generator
>     An internal Exception has occured. This should not have happenend.
>     Please open an issue at https://github.com/BobBuildTool/bob with the following backtrace:
>     Traceback (most recent call last):
>       File "/home/stefan/extsrc/bob/pym/bob/scripts.py", line 103, in bob
>         availableCommands[verb][1](argv, bobRoot)
>       File "/home/stefan/extsrc/bob/pym/bob/scripts.py", line 47, in __project
>         doProject(*args, **kwargs)
>       File "/home/stefan/extsrc/bob/pym/bob/cmds/build.py", line 1008, in doProject
>         generator(package, args.args, extra)
>       File "/home/stefan/extsrc/bob/pym/bob/generators/QtCreatorGenerator.py", line 321, in qtProjectGenerator
>         generateQtProject(package, args.destination, args.update, args.name, extra)
>       File "/home/stefan/extsrc/bob/pym/bob/generators/QtCreatorGenerator.py", line 295, in generateQtProject
>         ftype = magic.from_file(os.path.join(root, filename))
>     AttributeError: 'module' object has no attribute 'from_file'

Disclaimer: I do not use project file generation and didn't test this other than running the testsuite.

### DIFF
--- a/pym/bob/generators/EclipseCdtGenerator.py
+++ b/pym/bob/generators/EclipseCdtGenerator.py
@@ -19,12 +19,12 @@ import argparse
 import sys
 import os
 import re
-import magic
 import shutil
 import stat
 import uuid
 from os.path import expanduser
 from os.path import join
+from bob.utils import summonMagic
 from collections import OrderedDict
 from pipes import quote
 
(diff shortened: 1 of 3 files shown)


rhubert:
it's magic...  :+1: I tested the fix with eclipse and QtCreator generator after removing magic from pip and using apt-get python-magic.

PR #53, opened 2016-10-19
input: allow "+" in package/class names

stefanreuther:
Trivial change to schema: I had been using "+" in package/class names which the new schema verifier refuses. I see no technical reason to forbid that. Plus, Portage allows it :-)


jkloetzke:
I'm ok with adding '+' to the allowed characters. For consistency I would add it to the allowed tool names too.

So you have to adapt two more regex: toolGlobSchema and recipeFilterSchema.

stefanreuther:
That was a pedagogical "I would", right? :-)

jkloetzke:
Certainly. :stuck_out_tongue_winking_eye:

PR #55, opened 2016-10-20
bash-completion: query-path query-scm query-recipe

rhubert:
This adds completion for query-path, query-scm and query-recipe.
On bash it's not working if you use a unquoted or unescaped format sting.

@stefanreuther I think the brace expansion in bash breaks completion. Can we use another character for the braces in the format strings?

### DIFF
--- a/contrib/bash-completion
+++ b/contrib/bash-completion
@@ -269,6 +269,25 @@ __bob_project()
     __bob_project_generators "$($bob project --list)"
 }
 
+__bob_query_scm()
+{
+    __bob_complete_path "-f --default -r"
+}
+
+__bob_query_path()
+{
+   if [[ "$prev" = "-c" ]] ; then
+      COMPREPLY=( $(compgen -a -f -o dirnames "$cur" | cut -f 1 -d '.' ) )
+   else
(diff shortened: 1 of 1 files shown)


stefanreuther:
I'm not particularly attached to those braces, but it's what the Python standard library offers so I went with it for some sort of consistency.

jkloetzke:
Hmm, it's just too convenient to use. For query-scm you can even use some of the advanced formatting knobs like fixed field widths. Somehow the git completion must have solved this. There the `stash` command has the same problem (and it works):

    git stash show stash@{<tab><tab>

jkloetzke:
Should I merge or do you still want to test something with the brace handling?

rhubert:
I had a look into git completion. They have the same problem but you don't see it cause there are no more completion possibilities after `git stash show stash@{0}` Completion is working until the closing bace. After this git completion isn't called anymore. I verified this by adding a debug print to https://github.com/git/git/blob/master/contrib/completion/git-completion.bash#L2745 .

So from my side the is ready to merge since completion is working as log as used with format string in quotes
``` 
bob query-path -f "{build} {dist}" <tab>
vexpress/
```

jkloetzke:
Ok, thanks. :+1:

PR #59, opened 2016-10-28
Adding function match for regex based conditions.

karheinz:
The first commit adds a function called 'match' for regex based conditions.
The second commit fixes a typo in the docs.

### DIFF
--- a/doc/manual/manual/configuration.rst
+++ b/doc/manual/manual/configuration.rst
@@ -275,6 +275,10 @@ The following built in string functions are supported:
 
 * ``$(eq,left,right)``: Returns ``true`` if the expansions of ``left`` and
   ``right`` are equal, ``false`` otherwise.
+* ``$(match,string,pattern[,flags])``: Returns ``true`` if ``pattern`` is found
+  in ``string``, ``false`` otherwise. Quoting the pattern is recommended. Flags
+  are optional. The only supported flag by now is ``i`` to ignore case while
+  searching.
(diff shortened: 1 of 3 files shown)


test/test_input_recipe.py
```diff
@@ -149,3 +149,29 @@ def testNestedFwd(self):
         self.cmpEntry(res[3], "d", fwd=True)
         self.cmpEntry(res[4], "e")
 
+    def testConditions(self):
```
jkloetzke: That unit test looks odd. If at all it should be in a separate test suite and in a file called `test_input_env.py`. All these cases simply test the `funMatch` function that is already tested in `test_input_stringparser.py`. Instead a dedicated test for bob.input.Env.evaluate should be created without relying on a particular string function.

karheinz:
OK, I removed the test.

jkloetzke:
Looks good. Could you please squash the last commits with the previous?

karheinz:
Done ;)

rhubert:
I think someone should increment bob version after this commit...

jkloetzke:
Thanks. I would increment the version number with the next release. Otherwise we have to increment it basically with every commit...

rhubert:
I don't think so. Someone, lets call him Philipp, may want to use `match` tomorrow and in this case he needs to make sure the other team members update to a bob version supporting `match` using `bobMinimumVersion`.  :) 

Other commits (like `bob status`) do not need a new bob version...

jkloetzke:
Good point. I've pushed v0.7.0 with the changes so far...

PR #60, opened 2016-10-29
Project additional includes

rhubert:
With this you can add additional included directories (-I) to eclipse and
QtCreator projects. This may be usefull if some compiling for host where /usr/include
is unkown for bob.
For QtCreator you can now also add aditional files (e.g. *.ui) to project using a regex (-f <regex>).

### DIFF
--- a/pym/bob/generators/EclipseCdtGenerator.py
+++ b/pym/bob/generators/EclipseCdtGenerator.py
@@ -89,8 +89,13 @@ def createLaunchFile(outFile, exe, project):
     outFile.write('<stringAttribute key="process_factory_id" value="org.eclipse.cdt.dsf.gdb.GdbProcessFactory"/>\n')
     outFile.write('</launchConfiguration>')
 
+def addIncludes(cProjectFile, includeDirs):
+    cProjectFile.write('<option id="org.eclipse.cdt.build.core.settings.holder.incpaths.'+getId()+'" name="Include Paths" superClass="org.eclipse.cdt.build.core.settings.holder.incpaths" valueType="includePath">\n')
+    for i in includeDirs:
+        cProjectFile.write(' <listOptionValue builtIn="false" value="'+i+'"/>\n')
+    cProjectFile.write('</option>\n')
 
-def addCConfig(cProjectFile, excludePackages, buildName, id, buildArgs, buildMeFile):
+def addCConfig(cProjectFile, excludePackages, includeDirs, buildName, id, buildArgs, buildMeFile):
     cProjectFile.write('<cconfiguration id="' + id + '">\n')
     cProjectFile.write('<storageModule buildSystemId="org.eclipse.cdt.managedbuilder.core.configurationDataProvider" id="'
                         + id + '" moduleId="org.eclipse.cdt.core.settings" name="' + buildName + '">\n')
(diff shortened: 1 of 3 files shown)


pym/bob/generators/EclipseCdtGenerator.py
```diff
@@ -249,7 +266,10 @@ def generateEclipseProject(package, destination, updateOnly, projectName, exclud
         buildMe = []
         buildMe.append("#!/bin/sh")
         buildMe.append("bob dev $1 " + args + " " + project )
-        buildMe.append("bob project -n eclipseCdt " + project + " -u --destination " + destination + ' --name ' + projectName)
+        projectCmd = "bob project -n eclipseCdt " + project + " -u --destination " + destination + ' --name ' + projectName
```
jkloetzke: Shouldn't all parameters be quoted (via pipes.quote)? Otherwise the shell script might split arguments at white spaces or expand special characters.

pym/bob/generators/QtCreatorGenerator.py
```diff
@@ -194,7 +205,14 @@ def generateQtProject(package, destination, updateOnly, projectName, args):
         buildMe = []
         buildMe.append("#!/bin/sh")
         buildMe.append("bob dev $1 " + args + " " + project )
-        buildMe.append("bob project -n qt-creator " + project + " -u --destination " + destination + ' --name ' + projectName)
+        projectCmd = "bob project -n qt-creator " + project + " -u --destination " + destination + ' --name ' + projectName
+        for i in includeDirs:
+            projectCmd += " -I " + i
+        if filter:
+            projectCmd += " --filter " + filter
```
jkloetzke: Same here. Regex strings are especially sensitive and must be quoted correctly...

rhubert:
for cross compiling: is there a possibility to get all tools provided to a package? if so I would simply add all directories from all tools used to build to the includes?

rhubert:
Thinking a little bit longer about this crosscompiling thing - I think you found a bug in the actual implementation. I generated a QtCreator project for the bob-tutorials/vexpress and opened busybox/shell/ash.c. Then I jumped into <paths.h> which is resolved to /usr/include/paths.h  -> a header from my host. :-(

jkloetzke:
We would have to add a method to `Step` to get all the tools. But the only thing that you could do is to add the whole workspace of the the tool(s). That may be too coarse, though.

rhubert:
For QTCreator all directories and subdirectories of all `src`-dirs are added to includes-files. I don't think a few more directories make a big difference. Maybe for tools-includes we can also add some kind of heuristic ( e.g. search for `<tools-dist>/../usr/includes`  and only this directory. ) But, no matter how I look at it there is always a situation where the wrong include will be used. Selecting the include files in a multiarch project in QtCreator where two ore more toolchains are used will be a unsolvable task. :-1:  

So maybe the more correct way would be to remove (if possible) the search path for host includes and let the used decide which include dir's to add? 

I crosscheck this in eclipse. Standard includes like <stdio.h> are not resolve without adding `-I /usr/include` to `bob project`

rhubert:
Goolge brought me to a [nice tool](https://github.com/danimo/qt-creator/blob/master/src/tools/sdktool/README.txt) which can be used to add a new Kit to QtCreators profiles files:

Something like
```
./sdktool addTC --id "bob:toolchain" \
 --name "Bob Toolchain" \
 --path "/home/biera/projects/bob/bob" \
 --abi unknown
./sdktool addKit --id "bob_kit" \
 --name "Bob Kit" \
 --debuggerengine 1 \
 --debugger :auto \
 --devicetype Desktop \
 --toolchain "bob:toolchain"
```
can be used to add a new Kit to QtCreator without any compiler settings. Thus has the advantage that QtCreator doesn't have host includes or other preset settings which is the same behavior like eclipse. With this PR users can add include directories to get completion for SDK headers. 

Disadvantage is that bob's qtcreatorgenerator needs to know the path of the Qt installation. I have no idea how to get this path except using a env-var (like smartgit generator).

rhubert:
I reworked the handling of kits and profiles for QtCreator. I think this is the way we have to go to support cross compiling with QtCreator and Bob.

rhubert:
pushed another commit to this branch. This prevents *.files and *.includes to be regenerated with the same content and in result of this QTCreator from reindexing after each build.

jkloetzke:
Ready for merge?

rhubert:
Yes. I added a small fix in the generated buildme file. Instead of `$1 $@` should be used. Just stumbled over this today after adding `-vvv` to the build step in QtCreatorGui and wondering why the output is still the same :-1: Shall I squash this into a5eb0f0 or leave a separate commit?

Also wrapping my brain around calling `bob project` after `bob dev`. This took some seconds and is not necessary in most cases. Maybe it can also be done in background. But it will take some time to do some more investigation about this - so I'll come back with another PR in case I found a solution.

jkloetzke:
I would like to leave it as a separate commit. That's fine.

PR #62, opened 2016-10-29
bob status and --clean-checkout

rhubert:
As discussed on the mailing list this is my first attempt for bob status. 

One question that needs to be discussed is: What should --force-checkout do for url scms or if error is returned from Svn/Git? (Nothing / move to attic / delete old sources without moving to attic )? As a user I would assume that after --force-checkout the sources are in a clean state so moving to attic would the best?

### DIFF
--- a/contrib/bash-completion
+++ b/contrib/bash-completion
@@ -91,7 +91,7 @@ __bob_cook()
    elif [[ "$prev" = "-c" ]] ; then
       COMPREPLY=( $(compgen -a -f -o dirnames "$cur" | cut -f 1 -d '.' ) )
    else
-      __bob_complete_path "-e -f --force -n --no-deps -b --build-only --clean --incremental --resume -q --quiet --upload --download --sandbox --no-sandbox -v --verbose"
+      __bob_complete_path "-e -f --force --clean-checkout -n --no-deps -b --build-only --clean --incremental --resume -q --quiet --upload --download --sandbox --no-sandbox -v --verbose"
    fi
 }
 
--- a/contrib/bash-completion
+++ b/contrib/bash-completion
@@ -283,6 +283,15 @@ __bob_query_recipe()
     __bob_complete_path
 }
 
+__bob_status()
+{
+   if [[ "$prev" = "-c" ]] ; then
+      COMPREPLY=( $(compgen -a -f -o dirnames "$cur" | cut -f 1 -d '.' ) )
+   else
+      __bob_complete_path "-r --recursive -D -c --build --release -e -E -v"
+   fi
+}
+
 __bob_subcommands()
 {
    local i c command completion_func
@@ -318,7 +327,7 @@ __bob()
     local parse_pos=1 bob="$1" cur="$2" prev="$3"
     local sandbox=""
 
-   __bob_subcommands "build clean dev ls jenkins project query-path query-scm query-recipe"
+   __bob_subcommands "build clean dev ls jenkins project query-path query-scm query-recipe status"
 }
 else
 # Top level completion function for bash.
@@ -329,7 +338,7 @@ __bob()
 
    _get_comp_words_by_ref -n : cur prev
 
-   __bob_subcommands "build clean dev ls jenkins project query-path query-scm query-recipe"
+   __bob_subcommands "build clean dev ls jenkins project query-path query-scm query-recipe status"
 
    __ltrim_colon_completions "$cur"
 }
--- a/pym/bob/cmds/build.py
+++ b/pym/bob/cmds/build.py
@@ -362,6 +362,7 @@ esac
         self.__downloadDepth = 0xffff
         self.__bobRoot = bobRoot
         self.__cleanBuild = cleanBuild
+        self.__cleanCheckout = False
 
     def setArchiveHandler(self, archive):
         self.__doDownload = True
@@ -378,6 +379,9 @@ esac
 
     def setUploadMode(self, mode):
         self.__doUpload = mode
+
+    def setCleanCheckout(self, clean):
+        self.__cleanCheckout = clean
 
     def saveBuildState(self):
         # save as plain dict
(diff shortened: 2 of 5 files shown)


jkloetzke:
For UrlScm the status command could check if the checksum still matches. If no checksum was specified it cannot do anything sensible AFAICS.

Regarding forced checkout: if some kind of hard error is encountered I would stop with a `BuildError`. Instead of blindly moving something to the attic the user should be responsible to get the directory back into a  sane state. That's how Bob handles most of these errors anyway: only handle known cases and let the user fix the rest...

rhubert:
UrlScm: in case of a archive this will only check if the archive is not modified(?). What if someone has modified the extracted files?

for the forced checkout I'll modify this PR.

pym/bob/cmds/build.py
```diff
@@ -611,12 +623,20 @@ def _cookCheckoutStep(self, checkoutStep, done, depth):
                         .format(prettySrcPath), "33"))
                 else:
                     self._info("   CHECKOUT  skipped due to --build-only ({})".format(prettySrcPath))
-            elif (self.__force or (not checkoutStep.isDeterministic()) or
+            elif ((self.__force or (not checkoutStep.isDeterministic()) or
```
jkloetzke: I think the logic here and below has to be changed. The status of a directory must only be checked if the scmDigest of `oldCheckoutState` matches the one in `checkoutState`. Otherwise the directory content is not known. Anything else will be moved away anyway. All in all the logic should look like this unless I forgot something:

```python
if self.__buildOnly and (BobState().getResultHash(prettySrcPath) is not None):
  ...
else:
  if self.__forceCheckout:
    # check state of SCMs and invalidate if the directory is unclean
    stats = {}
    for scm in checkoutStep.getScmList():
      stats.update({ dir : scm for dir in scm.getDirectories.keys() })
    for (scmDir, scmDigest) in oldCheckoutState.copy().items():
      if scmDir is None: continue
      if scmDigest != checkoutState.get(scmDir): continue
      if stats[scmDir].status(checkoutStep.getWorkspacePath(), scmDir, False) == 'unclean':
        oldCheckoutState[scmDir] = None
  if (self.__force or (not checkoutStep.isDeterministic()) or
    (BobState().getResultHash(prettySrcPath) is None) or
    (checkoutState != oldCheckoutState)):
    ...
  else:
    self._info("   CHECKOUT  skipped (fixed package {})".format(prettySrcPath))
```

pym/bob/input.py
```diff
@@ -576,6 +577,67 @@ def isDeterministic(self):
     def hasJenkinsPlugin(self):
         return True
 
+    def callGit(self, workspacePath, *args):
+        cmdLine = ['git']
+        for arg in args:
+            cmdLine.append(arg)
+        proc = subprocess.Popen(cmdLine, cwd=os.path.join(os.getcwd(),workspacePath, self.__dir),
+                universal_newlines=True, stdout=subprocess.PIPE, stderr=subprocess.PIPE)
+        stdout, stderr = proc.communicate()
```
jkloetzke: Please don't throw away stderr. It might be valuable to the user. You may have a look at `subprocess.check_output()`. That should be better suited thatn `Popen` for the job I guess.

pym/bob/input.py
```diff
@@ -576,6 +577,67 @@ def isDeterministic(self):
...
+        return stdout
+
+    def status(self, workspacePath, dir, verbose):
+        if not os.path.exists(os.path.join(os.getcwd(), workspacePath, self.__dir)):
+            return
```
jkloetzke: Is `None` a valid return value?

contrib/bash-completion
```diff
@@ -91,7 +91,7 @@ __bob_cook()
    elif [[ "$prev" = "-c" ]] ; then
       COMPREPLY=( $(compgen -a -f -o dirnames "$cur" | cut -f 1 -d '.' ) )
    else
-      __bob_complete_path "-e -f --force -n --no-deps -b --build-only --clean --incremental --resume -q --quiet --upload --download --sandbox --no-sandbox -v --verbose"
+      __bob_complete_path "-e -f --force --force-checkout -n --no-deps -b --build-only --clean --incremental --resume -q --quiet --upload --download --sandbox --no-sandbox -v --verbose"
```
rhubert: I'm free of emotions at this point. ;-)

rhubert:
Ok - I modified the commits according to your comments. 

For UrlSCM (and CVS) it may be a option to print a warning like `unable to get SCM state.`?

pym/bob/input.py
```diff
@@ -576,6 +577,62 @@ def isDeterministic(self):
...
+        for arg in args:
+            cmdLine.append(arg)
+        try:
+            output = subprocess.check_output(cmdLine, cwd=os.path.join(os.getcwd(),workspacePath, self.__dir),
+                universal_newlines=True, stderr=subprocess.STDOUT)
+        except subprocess.CalledProcessError as e:
+            raise BuildError("git error: '{}' '{}'".format(" ".join(cmdLine), e.output))
+        return output
+
+    def status(self, workspacePath, dir, verbose):
```
jkloetzke: I would like to introduce at least another level of verbosity:

* 0: don't print anything
* 1: just print that a repo was touched, e.g. "dir: on different commit"
* 2: full details, e.g. with `git status` output what files are dirty

For the status command the default should be 1 because the user typically wants to know if some working copy needs to be taken care of. With the current implementation it is 2. If there are multiple repositories that have changes the output will get confusing.

pym/bob/input.py
```diff
@@ -576,6 +577,62 @@ def isDeterministic(self):
...
+        if not os.path.exists(os.path.join(os.getcwd(), workspacePath, self.__dir)):
+            return 'error'
+
+        ret = 0
+        if self.__commit:
+            output = self.callGit(workspacePath, 'rev-parse', 'HEAD').rstrip()
+            if output != self.__commit:
+                ret += 1
+                if verbose:
+                    print(colorize("    ! should have commitId {} but commitID is {}".format(self.__commit, output), "33"))
```
jkloetzke: All these prints need to add the directory. Otherwise the user doesn't know the exact working copy if multiple SCMs are checked out in a recipe.

pym/bob/input.py
```diff
@@ -576,6 +577,62 @@ def isDeterministic(self):
...
+                    print('      '+line.rstrip())
+            ret += 1
+
+        output = self.callGit(workspacePath, 'log', '--all', '--not', '--remotes')
+        if len(output):
+            if verbose:
+                print(colorize("    ! has unpushed commit(s):", "33"))
+                for line in output.split('\n'):
+                    print('      '+line.rstrip())
+            ret += 1
```
jkloetzke: Hmm, I think this check does not qualify the repository as "unclean" in the sense that it must be moved to the attic. The user might have just tested something previously and might have switched back to the upstream branch.

pym/bob/input.py
```diff
@@ -576,6 +577,62 @@ def isDeterministic(self):
     def hasJenkinsPlugin(self):
         return True
 
+    def callGit(self, workspacePath, *args):
+        cmdLine = ['git']
+        for arg in args:
+            cmdLine.append(arg)
+        try:
+            output = subprocess.check_output(cmdLine, cwd=os.path.join(os.getcwd(),workspacePath, self.__dir),
+                universal_newlines=True, stderr=subprocess.STDOUT)
```
rhubert: Redirecting stderr to stdout give stdout back as `e.output`
```python
 except subprocess.CalledProcessError as e:
            raise BuildError("git error: '{}' '{}'".format(" ".join(cmdLine), e.output))
```
This way the output is colorized like the build error and prefixed with `Build error: git error: 'git status --porcelain' 'error: <stderr>`

Otherwise stderr is white and BuildError is output after git's stderr.

pym/bob/input.py
rhubert: this information is given in build.py - showStatus (arround line 1100). 
```python
for (dir, hash) in scm.getDirectories().items():
    print(colorize("   STATUS {}".format(os.path.join(package.getCheckoutStep().getWorkspacePath(), dir)), "32"))
    scm.status(package.getCheckoutStep().getWorkspacePath(), dir, True)
```
I would move this to input.py and print this STATUS message only in unclean case for verbose=1? This would remove worthless output and focus on the unclean dirs.

pym/bob/input.py
rhubert: Depends on how you define the clean-checkout option I think. This check shows commits you haven't pushed to upstream and you build from other sources then the rest of your team, CI server or whatever. IMO this should be resolved by the clean-checkout option.

pym/bob/input.py
jkloetzke: Ok. Makes sense.

pym/bob/input.py
jkloetzke: Sorry, did't saw that. Agreed, a status message should only be printed if the user should pay attention...

pym/bob/input.py
```diff
@@ -684,6 +746,57 @@ def isDeterministic(self):
...
+            ret = printSCMStatus("    ! has modifications:", scmdir, verbose, ret)
+            if verbose >= 2:
+                for line in svnoutput.split('\n'):
+                    print('      '+line.rstrip())
+
+        _url = re.compile(r"^URL: {}")
+        _revision = re.compile(r"^Revision:")
+        svnoutput = self.callSubversion(os.path.join(os.getcwd(), workspacePath, m['dir']), 'info')
+        for line in svnoutput.split('\n'):
+            u = re.search('^URL: (.*)', line)
```
jkloetzke: This is dangerous. The output format is dependent on the locale. You should use the `--xml` option of subversion and get the information from there.

test/status/recipes/git.yaml
```diff
@@ -0,0 +1,19 @@
...
+environment:
+    TestBranch: "master"
+
+depends:
+    - svn
+
+checkoutSCM:
+    - 
+      scm: git
+      url: https://github.com/BobBuildTool/bob.git
```
jkloetzke: The test setup should prepare a local git repository with some commits. That makes it more deterministic and it does not rely on the network.

test/status/recipes/svn.yaml
```diff
@@ -0,0 +1,15 @@
+checkoutSCM:
+    - 
+      scm: svn
+      url: svn://svn.code.sf.net/p/pc1d/code/trunk
```
jkloetzke: The `svnserve` command should be available. The test case could spawn a server (after creating a repo) and the test could import some files. At the end the test must reap the server though.

```sh
svnadmin create foo
cat >foo/conf/svnserve.conf <<EOF
[general]
anon-access = write
EOF
svnserve -d --foreground -r .
```

pym/bob/cmds/build.py
```diff
@@ -1095,22 +1097,20 @@ def doStatus(argv, bobRoot):
 
     def showStatus(package, recurse, verbose, done):
         checkoutStep = package.getCheckoutStep()
-        if not checkoutStep.getVariantId() in done:
+        if checkoutStep.isValid() and (not checkoutStep.getVariantId() in done):
             done.add(checkoutStep.getVariantId())
             print(">>", colorize("/".join(package.getStack()), "32;1"))
             oldCheckoutState = BobState().getDirectoryState(checkoutStep.getWorkspacePath(), {})
             if not os.path.isdir(checkoutStep.getWorkspacePath()):
```
jkloetzke: Note that `checkoutStep.getWorkspacePath()` may return `None` in release mode if the step was never run!

pym/bob/input.py
rhubert: ahh - now I got you :+1: Didn't thought of this.

rhubert:
Beside the tests I've hopefully fixed all your findings without adding too much new faults... 

Do you want me to remove the tests from this PR? I'll add them anyway and create a new PR once they are ready to merge.

pym/bob/input.py
```diff
@@ -684,6 +757,54 @@ def isDeterministic(self):
...
+        svnoutput = self.callSubversion(os.path.join(os.getcwd(), workspacePath, m['dir']), 'status')
+        first = True
+        if len(svnoutput):
+            ret = printSCMStatus("    ! has modifications:", scmdir, verbose, ret)
+            if verbose >= 2:
+                for line in svnoutput.split('\n'):
+                    print('      '+line.rstrip())
+
+        _url = re.compile(r"^URL: {}")
+        _revision = re.compile(r"^Revision:")
```
jkloetzke: `first`, `_url` and `_revision` are not used.

rhubert: [ x ]

jkloetzke:
In principle I'm ok with merging it. I would like to give it one day or two of testing to see if no regressions were introduced.

Before finally merging it the commits should be squashed roughly into the following ones:
* input: add infrastructure to query SCM state
* build: add --clean-checkout option
* bob: add status command
* doc: add short documentation

I would keep the tests for a separate merge request and not merge them now.

rhubert:
While updating the docu for the new status format I noticed the check for the url in gitscm.status is missing. I'll add this - please do not merge.. :-|

rhubert:
I added the url-test for GitSCM.status. I leaved this as a separate commit  so it's easier to review. If it's ok I'll squash this in `input: add infrastructure to query SCM state`

pym/bob/input.py
```diff
@@ -593,27 +593,33 @@ def status(self, workspacePath, dir, verbose=0):
...
-                status += "S"
-                longStatus += colorize("    > branch: configured: {} actual: {}\n".format(self.__branch, output), "33")
-            else:
-                output = self.callGit(workspacePath, 'rev-list', 'origin/'+self.__branch+'..HEAD')
-                if len(output):
-                    status += "U"
-                    # do not print detailed status this point.
-                    # git log --branches --not --remotes --decorate will give the same informations.
+
+        output = self.callGit(workspacePath, 'ls-remote' ,'--get-url').rstrip()
```
jkloetzke: Looking at the man-page it seems that the remote name was mandatory until recently: https://git-scm.com/docs/git-ls-remote/2.6.2 vs. https://git-scm.com/docs/git-ls-remote/2.8.0 . If this is true then I guess most stable distros won't work...

rhubert: hmm... not sure but maybe it's a documentation issue? I one of these stable distros:
```
$ git --version
git version 1.9.1
 $ git ls-remote -?
usage: git ls-remote [--heads] [--tags]  [-u <exec> | --upload-pack <exec>]
                     [-q|--quiet] [--exit-code] [--get-url] [<repository> [<refs>...]]
$ git ls-remote --get-url 
https://github.com/rhubert/bob.git
```

Doing some archaeological research in git I found this behavior was added with https://github.com/git/git/commit/9c00de5a3135c8f7273668d4013c225d48d47861#diff-f6a8f8e2f3b2ef7d425d564cddebf989 and first tagged with git 1.7.2-rc0.

jkloetzke: Ok, then it seems to be just a documentation issue. As suggested, squash it into the first commit and unless you want to add something more I'm ok with merging it...

rhubert:
I squashed the commits together. At the moment there is nothing I want to add...

jkloetzke:
Thanks. :+1: I'll merge it...

PR #68, opened 2016-11-08
input: handle git errors smoothly

rhubert:
In case of a aborted checkout git may be in a unclean state. For this reason some
git commands are returning a error. GitSCM.status has been modified to print the
error without raising a exception.


jkloetzke:
I would simplify the logic quite a bit. Keep the BuildError exception in callGit and make a big try/except in the status method. If a build error happens print it and return 'error'...

Svn would be nice to fix in the same way too...

rhubert:
:+1: using a exception is too simple :( ... I thought about removing all these 
```python
if ret != 0:
    return 'error'
```
but no idea how to do this...

rhubert:
updated PR...

PR #70, opened 2016-11-11
Replaced the old StringParser with a new one.

karheinz:
The new StringParser is actually a suite of parsers and tokenizers.
Which parser and tokenizer is used depends on the actual context.
The behaviour is controlled by rules configured within each tokenizer.
After the tokenizer has finished, the corresponding parser checks, if
the detected tokens match the grammar of the actual context.  If not, a
ParseError with the detected tokens and their textual values is raised.

The hierarchical structure allows solid nesting of different contexts.
If required, tokens and grammar of each context can easily be modified.
The evaluation of the parse result is done in a seperate function.

This new implementation adds some more parameter expansion capabilities:

    * substring removal (%, %%, #, ##)
    * search and replace (//pattern/string, /pattern/string)
    * case modifications (^/^^/,/,,)

Within the patterns character classes ([...]) and wildcards (*, ?)
can be used. Variable substitution is also done within patterns.

### DIFF
--- a/pym/bob/parser.py
+++ b/pym/bob/parser.py
@@ -38,6 +38,7 @@ class TokenType:
     CMOD  = 'm'
     RMVL  = 'n'
     REPL  = 'o'
+    WHSP  = 'p'
 
     # virtual types
     TEXT  = '(' + '|'.join( [ REG, BS, ECH ] ) + ')'
@@ -59,6 +60,7 @@ TokenDesc = {
     TokenType.CMOD:  'CASE_MODIFICATION',
     TokenType.RMVL:  'REMOVAL_RULE',
     TokenType.REPL:  'REPLACEMENT_RULE',
+    TokenType.WHSP:  'WHITESPACE',
     }
 
 class Token:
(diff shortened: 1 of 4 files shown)


rhubert:
I did some tests with the new parser. 
This condition:
```yaml
depends:
   -  name: foo
      if: "$(and,$(is-sandbox-enabled)),$(not,$(is-tool-defined,bla)))"
```
evaluates to `true` without a sandbox enabled. The reason is the wrong second `)` after `is-sandbox-enabled`.  The new parser (the old one...) returns `false,true)` in this case witch is not false in `_isFalse`.

Maybe the error handling can be improved so that unbalanced braces are detected?

rhubert:
Not sure if this is a task for the string parser.

```yaml
if: $(and, false,true)
```
is `true`. Problem is the space before false. This space can be intended, but IMO it should be quoted if it's intended. All other spaces should be removed by the parser or throw a exception.

 Maybe @jkloetzke has a suggestion how to handle this..

karheinz:
The parser works correct in both cases! Maybe we should introduce a grammar (parser class) for if-conditions.

re.compile( '^{cmd}|{var}|{text}$'.format(...) )

Command args could be stripped before evaluation. This would solve the first case.

karheinz:
... solve the second case.

karheinz:
Copy and paste. Suggestion?

Am 15. November 2016 20:21:34 MEZ, schrieb Ralf <notifications@github.com>:

pym/bob/parser.py
```diff
@@ -0,0 +1,534 @@
+# Bob build tool
+# Copyright (C) 2016  TechniSat Digital GmbH
```
rhubert: I used 'BobBuildTool Team' last time... or your name?

jkloetzke:
The problem that Ralf raised is actually a problem of the "and" and "or" implementations. For example the "if-then-else" function get's it right by stripping white spaces in boolean context. Stripping all arguments is wrong and certainly backwards incompatible.

In general I'm suspicious about the heavy reliance on regular expressions. I have to delve more into the implementation but I guess it is considerably slower. As the string functions are in the hot spot of the package creation we will first have to implement effective caching before slowing down Bob on large projects...

karheinz:
I have removed stripping of command arguments. It should be done in the called function, if necessary.

The restricted grammar for if conditions is something to think about.

Last but not least:
You are right, the parser is too slow (at least for large projects).
Although i think the implementation has some advantages regarding maintainability and robustness ...

jkloetzke:
I would like to close the PR because I don't think it can be merged. Apart from needing a rebase it is pretty clear that the string parser is still in the hot path of the parsing. Slowing it down will still be very much noticable even after the parsing has been improved may times in the last releases.

For the future I would like to go two other ways:

* Maybe add an optional C-helper for the string parser to accelerate the substitution
* Explore if the boolean `if` fields can be extened with real expressions, e.g. instead of writing `if: "$(or,$(eq,${FOO},bar),${BAZ})` you can write `if: !expr '"${FOO}" == "bar" || "${BAZ}"'` like you can do already in the bob paths.

Nontheless, many thanks for the effort.

PR #71, opened 2016-11-12
Fix qt kits

rhubert:


pym/bob/generators/QtCreatorGenerator.py
```diff
@@ -169,28 +171,36 @@ def generateQtProject(package, destination, updateOnly, projectName, includeDirs
...
-        if os.path.exists(i):
-            for root, directories, filenames in os.walk(i):
-                hList.append(os.path.join(i,root))
+                if not hasInclude and include.match(filename):
+                    hasInclude = True
+            if hasInclude:
+                oldPath = ""
+                # need to recursively add all directories from the include up to cwd to get includes like
+                # <a/b/c.h> resolved even if there is no include in 'a'
+                root = root[len(os.getcwd()):]
```
jkloetzke: Isn't `os.getcwd()` wrong here? The base directory is `newPath` which could be a totally different directory if the user specified "--destination"...

pym/bob/generators/QtCreatorGenerator.py
```diff
@@ -169,28 +171,36 @@ def generateQtProject(package, destination, updateOnly, projectName, includeDirs
...
+                    if not newPath in hList:
+                        hList.append(newPath)
+                    oldPath = os.path.join(oldPath, path)
 
     # use default kit "Desktop" if no kit is given
     if kit is None:
-        kit = "Desktop"
+        _kit = re.compile(r".*Desktop.*")
+    else:
+        _kit = re.compile(r""+kit)
```
jkloetzke: Might not be an error. Is it intended that the user gives a regex in "--kit"?

rhubert: yes it was my intend to allow a regex for --kit. Even if this is probably never used...

pym/bob/generators/QtCreatorGenerator.py
rhubert: :+1: you're right...

rhubert:
Please do not merge. I'd like to do some more testing tomorrow...

pym/bob/generators/QtCreatorGenerator.py
jkloetzke: Ok, fine for me.

rhubert:
successful tested on several pc's. ready to merge from my side

PR #72, opened 2016-11-12
test: add git status test

rhubert:
As promised this is the test for GitSCM.status.

### DIFF
--- a/test/test_input_gitscm_status.py
+++ b/test/test_input_gitscm_status.py
@@ -1,0 +1,98 @@
+# Bob build tool
+# Copyright (C) 2016 BobBuildTool team
+#
+# This program is free software: you can redistribute it and/or modify
+# it under the terms of the GNU General Public License as published by
+# the Free Software Foundation, either version 3 of the License, or
+# (at your option) any later version.
+#
+# This program is distributed in the hope that it will be useful,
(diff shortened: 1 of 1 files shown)


test/test_input_gitscm_status.py
```diff
@@ -0,0 +1,112 @@
...
+        removePath(self.__class__.repodir)
+        removePath(self.__class__.repodir_local)
+
+    def setUp(self):
+        self.__class__.repodir = tempfile.mkdtemp()
+        self.__class__.repodir_local = tempfile.mkdtemp()
+
+        try:
+            cmdLine = 'git init'
+            output = subprocess.check_output(cmdLine.split(' '), cwd=self.__class__.repodir,
```
jkloetzke: What about adding a helper function? Untested idea:

```python
def callGit(self, *args, **kwargs):
  try:
    subprocess.check_output(*args, **kwargs, shell=True, universal_newlines=True, stderr=subprocess.STDOUT)
  except subprocess.CalledProcessError as e:
    self.fail("git error: '{}'".format(e.output))
```

Then you can simply call git with one line:

```python
self.callGit('git config user.email "bob@bob.bob"', cwd=self.repodir)
```

PR #73, opened 2016-11-13
Shell command archive

jkloetzke:
Refactors the binary artifact handling and provides a "shell" backend that might use any command for the actual up- or download.

### DIFF
--- a/pym/bob/archive.py
+++ b/pym/bob/archive.py
@@ -214,7 +214,7 @@ class SimpleHttpArchive(BaseArchive):
             cd $WORKSPACE
             BOB_UPLOAD_URL="{URL}/$(hexdump -e '2/1 "%02x/" 14/1 "%02x"' {BUILDID}).tgz"
             if ! curl --output /dev/null --silent --head --fail "$BOB_UPLOAD_URL" ; then
-                curl -sSg -T {RESULT} "$BOB_UPLOAD_URL" || echo Upload failed: $?
+                curl -sSg -T {RESULT} "$BOB_UPLOAD_URL"
             fi""".format(URL=self.__url, BUILDID=quote(buildIdFile), RESULT=quote(tgzFile)))
 
     def download(self, step, buildIdFile, tgzFile):
--- a/pym/bob/archive.py
+++ b/pym/bob/archive.py
@@ -34,10 +34,16 @@ class DummyArchive:
     def wantUpload(self, enable):
         pass
 
-    def canDownload(self):
+    def canDownloadLocal(self):
         return False
 
-    def canUpload(self):
+    def canUploadLocal(self):
+        return False
+
+    def canDownloadJenkins(self):
+        return False
+
+    def canUploadJenkins(self):
         return False
 
     def uploadPackage(self, buildId, path):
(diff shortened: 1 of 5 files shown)


pym/bob/archive.py
```diff
@@ -0,0 +1,389 @@
...
+        (tmpFd, tmpName) = mkstemp()
+        try:
+            os.close(tmpFd)
+            with tarfile.open(tmpName, mode="w:gz") as tar:
+                tar.add(path, arcname=".")
+
+            env = { k:v for (k,v) in os.environ.items() if k in self.__whiteList }
+            env["BOB_LOCAL_ARTIFACT"] = tmpName
+            env["BOB_REMOTE_ARTIFACT"] = self._makeUrl(buildId)
+            ret = subprocess.run(["/bin/bash", "-ec", self.__uploadCmd],
```
rhubert: Maybe I'm wrong but this limits uploadCmd to a cmd given in default.yaml? I'd prefer to put my commands into a separate shell script and/or have the full string functions. E.g.:
```yaml
archive:
       -
           backend: shell
           upload: "$<<scripts/myscript>> upload ${BOB_LOCAL_ARTIFACT} ${BOB_REMOTE_ARTIFACT}"
           upload: "$<<scripts/myscript>> download ${BOB_LOCAL_ARTIFACT} ${BOB_REMOTE_ARTIFACT}"
```

rhubert:
I'm thinking about changing the behavior for the download mode to not break working recipes through updating bob. Maybe all archives can be used for download unless tagged with:
```yaml
use: [upload, no-download]
```
?

pym/bob/archive.py
```diff
@@ -0,0 +1,389 @@
...
+                                 stdout=subprocess.DEVNULL, cwd="/tmp", env=env)
+            if ret.returncode == 0:
+                removePath(path)
+                os.makedirs(path)
+                with tarfile.open(tmpName, "r:gz", errorlevel=1) as tar:
+                    tar.extractall(path)
+                print(colorize("ok", "32"))
+                success = True
+            else:
+                print(colorize(str(e.reason), "33"))
```
jkloetzke: Correct. Will fix it...

pym/bob/archive.py
jkloetzke: Interesting idea. On the other hand you can write a whole script directly in default.yaml if you want as block text (like you do e.g. in `buildScript`). I thought about string functions but these commands have no real context. They are just there to transfer tar files to/from a remote server.

jkloetzke:
The default, if no `use` key is given, is `[download, upload]`. This is the same as before where this key did not exist. Do I overlook something? I'm not sure I understood your comment correctly...

rhubert:
I must have overlooked this default. This is what I wanted to say..

pym/bob/archive.py
rhubert: Sure you can, but do you really want to write complex scripts in yaml? 
Maybe for doing more complex thinks like uploading to different servers depending on the package or something like this it's better to have a plugin interface for uploading?

jkloetzke:
Updated to work with Python 3.3 and 3.4 too...

PR #75, opened 2016-11-16
Alias [RFC]

rhubert:
A different approach for the 'alias' feature. Just to be sure this is the way to go before going further...

What do you think about using the alias in bob's output? for example if there is a alias `my_lib: root/lib` you can now call `bob dev my_lib` but the output is
```
$ bob dev my_lib
>> root/lib
```
IMO this shouldn't be changed. It's more reasonable what's build with this alias.

### DIFF
--- a/pym/bob/cmds/misc.py
+++ b/pym/bob/cmds/misc.py
@@ -39,22 +39,22 @@ except UnicodeEncodeError:
 def doLS(argv, bobRoot):
     def showTree(packages, showAll, prefix=""):
         i = 0
-        for p in packages:
+        for n,p in sorted(packages.items()):
             last = (i >= len(packages)-1)
-            print("{}{}{}".format(prefix, LS_SEP_1 if last else LS_SEP_2, p.getName()))
+            print("{}{}{}".format(prefix, LS_SEP_1 if last else LS_SEP_2, n))
             deps = p.getAllDepSteps() if showAll else p.getDirectDepSteps()
-            showTree([d.getPackage() for d in deps], showAll,
+            showTree({ d.getPackage().getName():d.getPackage() for d in deps }, showAll,
                      prefix + (LS_SEP_3 if last else LS_SEP_4))
             i += 1
 
     def showPrefixed(packages, recurse, showAll, stack, level=0):
-        for p in packages:
+        for n,p in sorted(packages.items()):
             newStack = stack[:]
-            newStack.append(p.getName())
+            newStack.append(n)
             print("/".join(newStack))
             if recurse:
                 deps = p.getAllDepSteps() if showAll else p.getDirectDepSteps()
-                showPrefixed([d.getPackage() for d in deps], recurse, showAll,
+                showPrefixed({ d.getPackage().getName():d.getPackage() for d in deps }, recurse, showAll,
                              newStack, level+1)
 
     parser = argparse.ArgumentParser(prog="bob ls", description='List packages.')
@@ -77,21 +77,13 @@ def doLS(argv, bobRoot):
     recipes.parse()
 
     showAll = args.all
-    roots = sorted(recipes.generatePackages(lambda s,m: "unused", sandboxEnabled=args.sandbox).values(), key=lambda a:a.getName())
+    roots = recipes.generatePackages(lambda s,m: "unused", sandboxEnabled=args.sandbox)
     stack = []
     if args.package:
-        steps = [ s for s in args.package.split("/") if s != "" ]
-        for s in steps:
-            roots = [ r for r in roots if r.getName() == s ]
-            if not roots:
-                print("{}: not found in '{}'".format(s, "/".join(stack)),
-                      file=sys.stderr)
-                sys.exit(1)
-            if showAll:
-                roots = [ d.getPackage() for d in roots[0].getAllDepSteps() ]
-            else:
-                roots = [ d.getPackage() for d in roots[0].getDirectDepSteps() ]
-            stack.append(s)
+        package = walkPackagePath(roots, args.package)
+        stack = steps = [ s for s in args.package.split("/") if s != "" ]
+        deps = package.getAllDepSteps() if showAll else package.getDirectDepSteps()
+        roots = { d.getPackage().getName():d.getPackage() for d in deps }
     else:
         steps = ["/"]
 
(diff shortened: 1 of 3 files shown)


jkloetzke:
This is definitely going in the right direction. How about handling the aliases completely in `generatePackages()` already? Something like:

```python
def generatePackages(...):
  ...
  for i in self.__aliases:
    try:
      p = walkPackagePath(result, i)
      result[i] = p
    except BuildError as e:
      printWarning("Bad alias: " + str(e))
```

That would make the handling in the rest of the code considerably easier...

rhubert:
This is also working. Not sure how much performance impact this has. (from a simple lookup to a package walk..)??

jkloetzke:
The `walkPackagePath` is almost for free. I'm thinking about a way to get even rid of `getPackageAlias`. That will require to modify `doLS` to use `walkPackagePath` that need modification too. I'll have to think about it...

rhubert:
I think in `doLS` this can be replaced by 
```python
  package = walkPackagePath(roots, args.package)
  steps = [ s for s in package.getStack() if s != "" ]
```
and some more minor modifications...

jkloetzke:
The changes were more involved than I thought. I'll give it a bit of testing and push the changes to master in an hour or two...

rhubert:
Now I'm confused. Which changes? Are they already on a private branch?

rhubert:
I pushed my new solution without `getPackageAlias`.

Is there a one line expression for these loops?
```python
 roots = {}
 for d in package.getAllDepSteps():
   roots[d.getPackage().getName()] = d.getPackage()
```

jkloetzke:
I didn't manage to push my prototype before dinner. :laughing:

pym/bob/cmds/misc.py
```diff
@@ -39,22 +40,22 @@
 def doLS(argv, bobRoot):
     def showTree(packages, showAll, prefix=""):
         i = 0
-        for p in packages:
+        for n,p in packages.items():
```
jkloetzke: The order will be undefined. Better iterate over `sorted(packages.items())`.

pym/bob/cmds/misc.py
```diff
@@ -39,22 +40,22 @@
...
+            print("{}{}{}".format(prefix, LS_SEP_1 if last else LS_SEP_2, n))
             deps = p.getAllDepSteps() if showAll else p.getDirectDepSteps()
-            showTree([d.getPackage() for d in deps], showAll,
+            showTree({ d.getPackage().getName():d.getPackage() for d in deps }, showAll,
                      prefix + (LS_SEP_3 if last else LS_SEP_4))
             i += 1
 
     def showPrefixed(packages, recurse, showAll, stack, level=0):
-        for p in packages:
+        for n,p in packages.items():
```
jkloetzke: `sorted()`

pym/bob/cmds/misc.py
```diff
@@ -77,21 +78,14 @@ def showPrefixed(packages, recurse, showAll, stack, level=0):
     recipes.parse()
 
     showAll = args.all
-    roots = sorted(recipes.generatePackages(lambda s,m: "unused", sandboxEnabled=args.sandbox).values(), key=lambda a:a.getName())
+    roots = collections.OrderedDict(sorted(recipes.generatePackages(lambda s,m: "unused", sandboxEnabled=args.sandbox).items()))
```
jkloetzke: Due to the `sorted()` above this is unneeded. Just `recipes.generatePackages(...`...

pym/bob/cmds/misc.py
```diff
@@ -101,7 +95,7 @@ def showPrefixed(packages, recurse, showAll, stack, level=0):
         print("/".join(steps))
         showTree(roots, showAll)
     else:
-        for p in roots: print(p.getName())
+        for n,p in roots.items(): print(n)
```
jkloetzke: `sorted()`

pym/bob/cmds/misc.py
```diff
@@ -77,21 +78,14 @@ def showPrefixed(packages, recurse, showAll, stack, level=0):
...
-                print("{}: not found in '{}'".format(s, "/".join(stack)),
-                      file=sys.stderr)
-                sys.exit(1)
-            if showAll:
-                roots = [ d.getPackage() for d in roots[0].getAllDepSteps() ]
-            else:
-                roots = [ d.getPackage() for d in roots[0].getDirectDepSteps() ]
-            stack.append(s)
+        package = walkPackagePath(roots, args.package)
+        steps = package.getStack()
```
jkloetzke: This breaks the display of aliases. IMHO it should be:

```python
stack = steps = [ s for s in args.package.split("/") if s != "" ]
```

pym/bob/input.py
```diff
@@ -2731,6 +2736,12 @@ def generatePackages(self, nameFormatter, envOverrides={}, sandboxEnabled=False)
                 except ParseError as e:
                     e.pushFrame(root.getPackageName())
                     raise e
+            for i in self.__aliases:
+                try:
+                    p = walkPackagePath(result, self.__aliases[i])
```
jkloetzke: I think it's better to expand aliases only once. Otherwise an alias could again reference an alias. But the order of the for loop is undefined.

```python
tmp = result.copy()
for i in self.__aliases:
  ..
  p = walkPackagePath(tmp, ...
```

pym/bob/input.py
```diff
@@ -2731,6 +2736,12 @@ def generatePackages(self, nameFormatter, envOverrides={}, sandboxEnabled=False)
...
+                    result[i] = p
+                except BuildError as e:
+                    print("Bad alias: " + str(e))
```
jkloetzke: Should not be on stdout: `file=sys.stderr`. Maybe `colorize(..., "33")`?

rhubert: colorize looks a bit strange,  only the 'Bad alias:' in in color 33. The build error is still in red. Maybe BobError should be extended to get the plain slogan?

jkloetzke:
That's it. Just squash the last commit into the first and get rid of `import collections`. Then it's ready to merge...

rhubert:
Done :)

jkloetzke:
Amazing. Alias support with only 13 additional lines. Including documentation. :smile:

Thanks. :+1:

PR #76, opened 2016-11-19
cvs scm: implement 'bob status'

stefanreuther:
Simple, brute-force implementation using `cvs update` as the basis. It will therefore contact the server and tell you where the server's branch differs from yours (instead of just your local modifications), but everything else probably involves manually parsing CVS's metadata files.

I'm not sure how this interacts with the `status()` call in `_cookCheckoutStep`. Documentation about the expected behaviour would have helped :-)

### DIFF
--- a/pym/bob/cmds/build.py
+++ b/pym/bob/cmds/build.py
@@ -1058,7 +1058,7 @@ def doStatus(argv, bobRoot):
                 stats = {}
                 for scm in checkoutStep.getScmList():
                     stats.update({ dir : scm for dir in scm.getDirectories().keys() })
-                for (scmDir, scmDigest) in oldCheckoutState.copy().items():
+                for (scmDir, scmDigest) in sorted(oldCheckoutState.copy().items(), key=lambda a:'' if a[0] is None else a[0]):
                     if scmDir is None: continue
                     if scmDigest != checkoutState.get(scmDir): continue
                     stats[scmDir].status(checkoutStep.getWorkspacePath(), scmDir, verbose)
--- a/pym/bob/scm/cvs.py
+++ b/pym/bob/scm/cvs.py
@@ -15,8 +15,11 @@
 # along with this program.  If not, see <http://www.gnu.org/licenses/>.
 
 from ..utils import hashString
+from ..tty import colorize
 import re
 import schema
+import os
+import subprocess
 
 class CvsScm:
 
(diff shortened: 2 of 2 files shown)


pym/bob/input.py
```diff
@@ -911,8 +911,91 @@ def hasJenkinsPlugin(self):
         return False
 
     def status(self, workspacePath, dir, verbose = 0):
-        print("CVS SCM status not implemented!")
-        return 'error'
+        workDir = os.path.join(workspacePath, dir)
+        cvsDir = os.path.join(workDir, 'CVS')
+        if not os.path.exists(workDir) or not os.path.exists(cvsDir):
+            return 'error'
```
jkloetzke: Should be

```python
if not os.path.exists(workDir): return "emty"
if not os.path.exists(cvsDir): return "error"
```

@rhubert ?

pym/bob/input.py
```diff
@@ -911,8 +911,91 @@ def hasJenkinsPlugin(self):
...
+                else:
+                    commandLine += ['-r', self.__rev]
+                output = subprocess.check_output(commandLine,
+                                                 cwd=workDir,
+                                                 universal_newlines=True,
+                                                 env=environment,
+                                                 stderr=subprocess.DEVNULL,
+                                                 stdin=subprocess.DEVNULL)
+            except subprocess.CalledProcessError as e:
+                raise BuildError("cvs error: '{}' '{}'".format(" ".join(cmdLine), e.output))
```
jkloetzke: Just print the error and return `'error'`. The status command should not throw exceptions. Otherwise `--clean-checkout` wouldn't work and the user has the inconvenience of having to do something about it to get the status command to move on...

pym/bob/input.py
```diff
@@ -911,8 +911,91 @@ def hasJenkinsPlugin(self):
...
+            if (verbose >= 2) and (longStatus != ""):
+                print(longStatus)
+            return 'unclean'
+
+    @staticmethod
+    def _loadFile(name):
+        try:
+            with open(name) as f:
+                return f.read().rstrip('\r\n')
+        except Exception:
```
jkloetzke: Better just catch the `OSError` that you're expecting. Otherwise you would also catch programming errors like a misspelled variable, i.e. `NameError` exceptions.

pym/bob/input.py
rhubert: I think so too. If the workDir isn't existing return empty. If you return error the `_cookCheckoutStep` will try to move a not existing directory to attic. If there is no CVS dir you can maybe test if the workDir is empty and only if it's not return error. For `error` and `unclean` the directory is moved to attic if build with the `--clean-checkout` argument.

rhubert:
> @rhubert Would you take care of adding some lines, e.g. to GitScm about the expected behavior and return values?

I'll put this on my todo list...

pym/bob/input.py
stefanreuther: I copied this logic from SvnScm.callSubversion. So maybe it's wrong, but it's consistent :)

pym/bob/input.py
stefanreuther: Likewise, I copied this logic from SvnScm which returns "error" if it cannot find the directory.

    def status(self, workspacePath, dir, verbose=0):
        scmdir = os.path.join(workspacePath, dir)
        if not os.path.exists(os.path.join(os.getcwd(), scmdir)):
            return 'error'

Looking forward to an interface description of the `status` method :-)))

rhubert: Better copy from GitScm :-/ should be fixed for svn...

pym/bob/input.py
jkloetzke: Apparently svn is broken in this respect. Even though `callGit()` raises `BuildError` too, this is catched in `GitScm.status()` and converted to an `'error'` return.

stefanreuther:
Fixed review issues and ported the code to the new place.

PR #78, opened 2016-11-22
Qt/EclipseGenerators: clean and buildCfg argument

rhubert:
With --clean the generator removes the destination directory before generating
the project files. This may be usefull in case of a non working project because of
changes in the creator.user file.

The --buildCfg argument can be used to add additional buildConfigs. For example
--buildCfg "Bob dev DEBUG"::"-b -DDEBUG=1".

### DIFF
--- a/pym/bob/generators/EclipseCdtGenerator.py
+++ b/pym/bob/generators/EclipseCdtGenerator.py
@@ -24,7 +24,7 @@ import stat
 import uuid
 from os.path import expanduser
 from os.path import join
-from bob.utils import summonMagic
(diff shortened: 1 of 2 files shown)


jkloetzke:
Hmm, --clean is already used for another purpose in "bob build/dev". That may even be worth passing it through bob project. Maybe "--overwrite" is a better name?

rhubert:
yep

jkloetzke:
Ok. Did it get some light testing and is it ready for merge?

rhubert:
Tested both on a larger project today. Its ready to merge.

PR #79, opened 2016-11-22
BOB_VERSION: add git infos

rhubert:
Append the output of git describe to BOB_VERSION.

### DIFF
--- a/pym/bob/__init__.py
+++ b/pym/bob/__init__.py
@@ -14,5 +14,19 @@
 # You should have received a copy of the GNU General Public License
 # along with this program.  If not, see <http://www.gnu.org/licenses/>.
 
+def getVersion():
+    import os
+    import subprocess
+    version=""
+    try:
+       root = os.path.dirname(os.path.realpath(__file__))
+       _cwd = os.path.join(root, '..', '..')
+       version = subprocess.check_output("git describe --tags --dirty --always".split(" "), cwd=_cwd,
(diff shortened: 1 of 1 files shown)


jkloetzke:
I've amended your commit and added proper support for installed copies and downloaded tarballs: https://github.com/jkloetzke/bob/commit/5550adfef1adf3c971ae5b2ab133aed7b6c80f7f Are you ok with that? I would still keep you as the author.

rhubert:
Looks good to me. :+1: I don't mind if you remove me as author. What counts is the feature. ( and nobody can blame me for that ;) )

Btw. is `bobMinimumVersion` working for additional commits? Something like 
```yaml
bobMinimumVersion: "0.8-4"
```

jkloetzke:
The version comparison function will ignore anything after the first dash (-). That is not according to semver but that is the output of git describe neither. It should be enough for our purposes, though. I've considered the correct implementation that would give something like "0.9-dev+v0.8.0-5-g5550adf" but that looks just plain ugly.

jkloetzke:
To be more explicit: you cannot compare to a pre-relase version at all. For `bobMinimumVersion` you can only give one to three numbers separated by dots. Anything else will fail schema validation. That could be extended in the future, though.

rhubert:
Understand. Shall I close this and you merge this from your branch?

jkloetzke:
Yes, please. I've just pushed my rework.

PR #80, opened 2016-11-23
Added exclude option to qtcreator project file generation

elmodor:
Added the option to exclude projects for the qtcreator project.
This is a bit different than for eclipse:
No sym link and no include files will be created. Else qtcreator will parse all the files which we want to avoid. 

~Mark

### DIFF
--- a/pym/bob/generators/QtCreatorGenerator.py
+++ b/pym/bob/generators/QtCreatorGenerator.py
@@ -32,11 +32,19 @@ from bob.tty import colorize
 from pipes import quote
 
 # scan package recursivelely with its dependencies and build a list of checkout dirs
-def getCheckOutDirs(package, dirs):
+def getCheckOutDirs(package, excludes, dirs):
     if package.getCheckoutStep().isValid():
         dirs.append([package.getName(), package.getCheckoutStep().getWorkspacePath()])
+
     for d in package.getDirectDepSteps():
-        getCheckOutDirs(d.getPackage(), dirs)
+        excluded = False
+        for e in excludes:
+            if (e.match(d.getPackage().getName())):
+                excluded = True
+                break
+
+        if not excluded:
+            getCheckOutDirs(d.getPackage(), excludes, dirs)
 
 def generateFile(entries, fileName):
     try:
@@ -148,6 +156,8 @@ def qtProjectGenerator(package, argv, extra):
         help="Additional include directories. (added recursive starting from this directory)")
     parser.add_argument('-f', '--filter', metavar="Filter",
         help="File filter. A regex for matching additional files.")
+    parser.add_argument('--exclude', default=[], action='append', dest="excludes",
+            help="Package filter. A regex for excluding packages in QTCreator.")
     parser.add_argument('--kit',
         help="Kit to use for this project")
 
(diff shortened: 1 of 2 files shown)


pym/bob/generators/QtCreatorGenerator.py
```diff
@@ -169,6 +177,7 @@ def generateQtProject(package, destination, updateOnly, projectName, includeDirs
     # now go through all checkout dirs to find all header / include files
     for name,path in OrderedDict(sorted(dirs, key=lambda t: t[1])).items():
         newPath = os.path.join(symlinkDir, name)
+        if name in excludePackages: continue
```
rhubert: move this one line up - `newPath` is not needed for excluded Packages.

pym/bob/generators/QtCreatorGenerator.py
```diff
@@ -141,6 +141,14 @@ def generateQtProject(package, destination, updateOnly, projectName, includeDirs
...
         os.makedirs(destination)
 
+    # setup list of exclude packages
+    excludePackages = []
+    for e in excludes:
+        exp = re.compile(e)
+        for name,path in OrderedDict(sorted(dirs, key=lambda t: t[1])).items():
+            if exp.match(name):
+                excludePackages.append(name)
+
```
rhubert: I would move this to `getCheckoutDirs`. You can simply skip a excluded directory and not add it to the dict.

rhubert:
Please have a look at https://github.com/BobBuildTool/bob/pull/78 after this is merged your PR will not be able to merge.

elmodor:
Yes, this will need to be changed to work with #78

rhubert:
Just a mind game: Normally have one, maybe two ore three but not more than five packages you're active developing. But there are a lot of dependencies and all of them are picked up without the exclude option which brings all IDE's into trouble. For excluded this would result in a long list of packages. So what about reversing this and add also a --use-only argument to add only the packages you really need?

elmodor:
I see a use case for both options. Sometimes you need the dependencies of all the other internal modules while you only want to exclude the external ones. 
On the other hand if you don't case about dependencies the --use-only option seems to be a better one.

rhubert:
Challenge accepted? ;)

elmodor:
I moved the excludes to getCheckOutDirs. I will have to test this tomorrow.

elmodor:
Tested and it works (with and without excludes).

rhubert:
What about the `--use-only`? I think the passing a negative regex like `^((?!abc|def).)*$` to --exclude is not relay comfortable. At least because of the histexpand feature of bash you need to turn of first.

elmodor:
I guess the `--use-only` could still be added if needed with a later PR? This PR is mainly to have an `--exclude` feature for qtcreator similar to the already implemented one for eclipse.

rhubert:
You can add --use-only with 9 additional lines. Would make this PR more complete... 
```python
   group.add_argument('--use-only', default=[], action='append',  
          help="Use only these packages in QTCreator.")
   group.add_argument('--exclude', default=[], action='append', dest="excludes",  
          help="Packages will be excluded in QTCreator.")

[...]

excludes = []
if args.excludes:
   for e in args.excludes:
     excludes.append(re.compile(e))
if args.use-only:
   for e in args.use-only:
     excludes.append(re.compile(r"^((?!"+e+").)*$"))

[...]

 if args.use-only:
            for e in args.use-only
                projectCmd += "--use-only " + quote(e)
```

This is untested of course :)

jkloetzke:
:+1: Looks good to me. I would also suggest to make ``--use-only`` a separate merge request.

Some minors:

* Could you amend the manpage (doc/manpages/bob-project.rst) by adjusting the synopsis of the QtCreator paragraph and documenting the new option? That should be a separate commit, though.
* It would be nice if the commit messages follow the style of the repository, e.g.:

>    QtCreator: add --exclude option
>
>    Add option to exclude whole packages by matching the given regular
>    expression with the package name. No symlinks and no include files will
>    be created for excluded packages.

elmodor:
Done as you suggested @jkloetzke

PR #81, opened 2016-11-23
Added some more build targets

elmodor:
Added a Debug and Clean build target. 
I think it would be easier just to choose a different build target if you need a debug buil than to create a new project file. 

How do you even develop without a proper debug build? I think this should always be an option as a build target for developing.

~Mark

### DIFF
--- a/pym/bob/generators/QtCreatorGenerator.py
+++ b/pym/bob/generators/QtCreatorGenerator.py
@@ -110,8 +110,9 @@ def addBuildSteps(outFile, buildMeFile, buildConfigs):
     addBuildConfig(outFile, 2, "Bob dev (no checkout)", "-b", buildMeFile)
     addBuildConfig(outFile, 3, "Bob dev (no deps)", "-n", buildMeFile)
     addBuildConfig(outFile, 4, "Bob dev (no checkout, no deps)", "-nb", buildMeFile)
+    addBuildConfig(outFile, 5, "Bob dev (no checkout, clean)", "-b --clean", buildMeFile)
 
-    count = 5
+    count = 6
     for name,flags in buildConfigs:
         addBuildConfig(outFile, count, name, flags, buildMeFile)
         count += 1
--- a/pym/bob/generators/EclipseCdtGenerator.py
+++ b/pym/bob/generators/EclipseCdtGenerator.py
@@ -243,6 +243,7 @@ def eclipseCdtGenerator(package, argv, extra):
         addCConfig(cProjectFile, excludePackages, includeDirs, "Bob dev (no checkout)",id + "." + getId(), "-b", buildMeFile)
         addCConfig(cProjectFile, excludePackages, includeDirs, "Bob dev (no deps)",id + "." + getId(), "-n", buildMeFile)
         addCConfig(cProjectFile, excludePackages, includeDirs, "Bob dev (no checkout, no deps)",id + "." + getId(), "-bn", buildMeFile)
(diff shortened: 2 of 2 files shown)


pym/bob/generators/QtCreatorGenerator.py
```diff
@@ -109,8 +109,10 @@ def addBuildSteps(outFile, buildMeFile):
     addBuildConfig(outFile, 2, "Bob dev (no checkout)", "-b", buildMeFile)
     addBuildConfig(outFile, 3, "Bob dev (no deps)", "-n", buildMeFile)
     addBuildConfig(outFile, 4, "Bob dev (no checkout, no deps)", "-nb", buildMeFile)
+    addBuildConfig(outFile, 5, "Bob dev (no checkout, Debug)", "-b -DCMAKE_BUILD_TYPE='DEBUG'", buildMeFile)
```
rhubert: this makes no sense for non cmake projects. There is already another PR for this. https://github.com/BobBuildTool/bob/pull/78 

Please add the clean build config to eclipse also.

elmodor: I was not aware that this is used for non cmake projects.

In this case I guess this PR can be closed?

rhubert: No.. I think the clean build may be useful :+1:

elmodor:
Only the clean build target is added now for QTCreator and Eclipse.

rhubert:
Hmm.. The commits you want to merge are looking strange. 
Do you want to force build for clean builds?

elmodor:
When you do a clean build you want to execute all build steps. We had the case that when doing a clean build with no changes no build was made but instead the data in dev/dist was used. Forcing to build all steps prevents this. However this might have been a bug and then the -f option should be removed from clean builds, yes.

elmodor:
-f is removed

rhubert:
That's true. `--clean` clears the build directory. If you haven't changed the sources buildStep is not executed. So maybe you want the `-f`? I'm unsure...

elmodor:
Exactly. However what is the use of doing a bob dev `--clean`  if you don't build it afterwards? the `-f` will ensure that the project will be build afterwards with a clean build dir. 

That's why I think `-f` should be added.

elmodor:
However maybe this is the wrong solution. Maybe `--clean` should also do a `-f` by default? So a "clean build" is ensured.

rhubert:
This clean is a clean before build and not a 'remove all build results' like make clean. If you want to have the 'make clean' functionality I think you should add `-f`

elmodor:
How about adding the `-f` in the build target and name it "Bob dev (no checkout, make clean)" ? 
I understand how you might want to do a `--clean` while operating in the terminal but if you develop in QTCreator/Eclipse I don't think you want a clean, only a make clean?

rhubert:
What you want to do is `bob clean --remove-build --remove-dist`. I think we should improve the clean command first...

jkloetzke:
There is no such thing as "make clean" in Bob (yet). The "--clean" option is the opposite of "--incremental". It makes sure that packages are built in a clean workspace. If there was no change to the sources or the dependencies either then Bob will not rebuild the package. Generally speaking, if you have to use "-f" for a regular build then something is wrong with the recipes.

Using "-f" and "--clean" together will rebuild the whole project from scratch. It will even re-install the toolchains. That should hopefully never be needed.

elmodor:
So what does this mean for this PR? Is this on hold for now until a proper bob clean is implemented? Or could it be used as it is?

jkloetzke:
I think there is some value in adding the "--clean" option to the profiles. @rhubert do you agree? I'm ok with merging it...

rhubert:
Ok for me.

PR #82, opened 2016-11-24
Added build only (no checkout) as option for bob project

elmodor:
Added the option for a build only mode when creating projects:
bob project -b
does the same as bob dev -b (no checkout).
This should solve the issue that -n has but will allow you to generate a project if -n fails but you don't want to do a checkout.

~Mark

### DIFF
--- a/pym/bob/cmds/build.py
+++ b/pym/bob/cmds/build.py
@@ -1018,6 +1018,8 @@ def doProject(argv, bobRoot):
         help="Resume build where it was previously interrupted")
     parser.add_argument('-n', dest="execute_prebuild", default=True, action='store_false',
         help="Do not build (bob dev) before generate project Files. RunTargets may not work")
+    parser.add_argument('-b', dest="execute_buildonly", default=False, action='store_true',
+        help="Do build only (bob dev -b) before generate project Files. No checkout")
     args = parser.parse_args(argv)
 
     defines = {}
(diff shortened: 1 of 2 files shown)


pym/bob/cmds/build.py
```diff
@@ -1085,6 +1087,7 @@ def doProject(argv, bobRoot):
     if args.execute_prebuild:
         devArgs = extra.copy()
         if args.resume: devArgs.append('--resume')
+        if args.execute_buildonly: devArgs.append('-b')
```
rhubert: :+1: Can you add another fix to this and squash both together?

Line 1079:
```python
- extra.append('e')
+ extra.append('-e')
```

elmodor: Done

rhubert: Thanks :) ... Just saw this while reviewing your PR. Looks like nobody used this before...

pym/bob/cmds/build.py
```diff
@@ -1074,7 +1076,7 @@ def doProject(argv, bobRoot):
         extra.append('-c')
         extra.append(c)
     for e in args.white_list:
-        extra.append('e')
+        extra.append('-e')
```
jkloetzke: Would you mind splitting out this fix into a separate commit? Fixes and features should not be mixed.

elmodor:
Done as you suggested @jkloetzke

PR #85, opened 2016-11-28
Qtcreator useonly

elmodor:
Added the QtCreator `--use-only` feature as discussed in PR #80 .
Tested and it worked to include only projects.

~ Mark

### DIFF
--- a/pym/bob/generators/QtCreatorGenerator.py
+++ b/pym/bob/generators/QtCreatorGenerator.py
@@ -158,6 +158,8 @@ def qtProjectGenerator(package, argv, extra):
         help="File filter. A regex for matching additional files.")
     parser.add_argument('--exclude', default=[], action='append', dest="excludes",
             help="Package filter. A regex for excluding packages in QTCreator.")
+    parser.add_argument('--include', default=[], action='append', dest="include",
+            help="Include package filter. A regex for including only the specified packages in QTCreator.")
     parser.add_argument('--kit',
         help="Kit to use for this project")
 
@@ -174,6 +176,10 @@ def qtProjectGenerator(package, argv, extra):
     if args.excludes:
         for e in args.excludes:
             excludes.append(re.compile(e))
+
+    if args.include:
+        for e in args.include:
+            excludes.append(re.compile(r"^((?!"+e+").)*$"))
 
     getCheckOutDirs(package, excludes, dirs)
     if not projectName:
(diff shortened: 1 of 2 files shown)


pym/bob/generators/QtCreatorGenerator.py
```diff
@@ -158,6 +158,8 @@ def qtProjectGenerator(package, argv, extra):
         help="File filter. A regex for matching additional files.")
     parser.add_argument('--exclude', default=[], action='append', dest="excludes",
             help="Package filter. A regex for excluding packages in QTCreator.")
+    parser.add_argument('--use-only', default=[], action='append', dest="useonly",
```
rhubert: I would add both arguments to a mutual exclusive group. Doesn't make any sense to me to use both at the same time. And i think it's undefined who wins this game..

jkloetzke:
Hmm, as the implementation already shows it... One could also amend the manpage of "--exclude" to document that negative regex expressions are also possible.

I have no hard opinion here. But adding a new option that is already there in principle feels sub optimal. Thoughts? @rhubert

pym/bob/generators/QtCreatorGenerator.py
jkloetzke: Well you could do ``--use-only 'foobar-.*' --exclude 'foobar-baz'`` for example. It's all about the order...

elmodor:
It would be easier to use and you would not need to type a very long regex for every "use only".

rhubert:
Try to give a negative lookahead on your cmdline. e.g. for the binary repo tutorial:
```bash
$ bob project qt-creator --exclude ^((?!lib).)*$
bash: !lib: event not found
```
the `!` is interpreted as history substitution. You'll need to turn of this (`set +H`) which is not really comfortable. I see the main advantage of this option to avoid this and to increase usability.

jkloetzke:
> ```
> $ bob project qt-creator --exclude ^((?!lib).)*$
> bash: !lib: event not found
> ```

Well, it's a regex. You also have globbing characters in them (?*). If you set ``shopt -s nullglob`` it will make your regex disappear altogether. So you should always single quote them.

I do see the advantage of a handy dedicated option. But as Ralf pointed out it's not immediately clear how the "--use-only" option interacts with the "--exclude" option. @elmodor could you add some explanation in the manpage..?

rhubert:
single quote is a cool trick :+1:

pym/bob/generators/QtCreatorGenerator.py
rhubert: I'm still not so happy with the name of the argument, even if I don't have a counterproposition. What do you think about `--include-filter` (and `--exclude-filter` instead of `--exclude`)?

jkloetzke:
@elmodor ping? Do you want to update the switch to "--include" as suggested?

elmodor:
@jkloetzke Sorry that it took so long.

I updated the manpage and change the argument to "--include"

jkloetzke:
Thanks. :+1:  Just before the next release. :wink:

PR #86, opened 2016-12-01
build: do not pipe output of buildScript [RFC] [Do not merge :)]

rhubert:
Many tools like gcc, ls, .. detect whether they are running on a tty or not.
If running on a tty the output is different ( colored ). To get these colors
back in bob use a pty to capure the output.

I'm using a regex to remove ansi escape codes from the output before writing the logfile. The time python need to process there regex is hopefully much less compared to the cpu time of the called script. :sunglasses: 

Also I haven't fully understood the old output processing. 
```bash
      run_script | tee -a ../log.txt
   }} 3>&1 1>&2- 2>&3- | tee -a ../log.txt
}} 3>&1 1>&2- 2>&3-
``` 
may have a different effect than my new implementation on this...  :confused:

### DIFF
--- a/pym/bob/cmds/build.py
+++ b/pym/bob/cmds/build.py
@@ -26,9 +26,13 @@ from pipes import quote
 import argparse
 import datetime
 import os
+import pty
+import re
+import select
 import shutil
 import stat
 import subprocess
+import sys
 
 # Output verbosity:
 #    <= -2: package name
@@ -135,27 +139,7 @@ case "${{1:-run}}" in
         ;;
     __run)
         cd "${{0%/*}}/workspace"
-        case "$_verbose" in
-            0)
-                run_script >> ../log.txt 2>&1
-                ;;
-            1)
-                set -o pipefail
-                {{
-                    {{
-                        run_script | tee -a ../log.txt
-                    }} 3>&1 1>&2- 2>&3- | tee -a ../log.txt
-                }} 1>&2- 2>/dev/null
-                ;;
-            *)
-                set -o pipefail
-                {{
-                    {{
-                        run_script | tee -a ../log.txt
-                    }} 3>&1 1>&2- 2>&3- | tee -a ../log.txt
-                }} 3>&1 1>&2- 2>&3-
-                ;;
-        esac
+        run_script
         ;;
     shell)
         if [[ $_keep_env = 1 ]] ; then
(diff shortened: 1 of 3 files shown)


pym/bob/cmds/build.py
```diff
@@ -458,15 +442,43 @@ def _runShell(self, step, scriptName):
         elif self.__verbose >= 2:
             cmdLine.append('-vv')
 
-        proc = subprocess.Popen(cmdLine, cwd=step.getWorkspacePath(), env=runEnv)
+        oup = pty.openpty()
+        erp = pty.openpty()
+
+        logFile = open(os.path.join(step.getWorkspacePath(), '..', 'log.txt'), 'a')
+        proc = subprocess.Popen(cmdLine, cwd=step.getWorkspacePath(), env=runEnv,
+                universal_newlines=True, stdout=oup[1], stderr=erp[1])
```
jkloetzke: `universal_newlines` should have no influence as raw file descriptors are used.

pym/bob/cmds/build.py
```diff
@@ -458,15 +442,43 @@ def _runShell(self, step, scriptName):
...
+
+        ansi_escape = re.compile(r'\x1B\[([0-9]{1,2}(;[0-9]{1,2})?)?[m|K]')
+
         try:
-            if proc.wait() != 0:
+            while 1:
+                timeout = .1
```
jkloetzke: Instead of a timeout a SIGCHLD handler should be used to get an immediate feedback and prevent polling.

pym/bob/cmds/build.py
```diff
@@ -458,15 +442,43 @@ def _runShell(self, step, scriptName):
...
+                ready, _, _ = select.select([oup[0], erp[0]], [], [], timeout)
+                if ready:
+                    for r in ready:
+                        data = os.read(r, 1024)
+                        if ((r == erp[0]) and (self.__verbose >= 0)) or ((r == oup[0]) and (self.__verbose >= 1)):
+                            sys.stdout.buffer.write(data)
```
jkloetzke: stderr should go to stderr...

pym/bob/cmds/build.py
```diff
@@ -458,15 +442,43 @@ def _runShell(self, step, scriptName):
...
+                        logFile.write(ansi_escape.sub('', data.decode(sys.stdout.encoding)).replace('\r\n', '\n'))
```
jkloetzke: This can blow up for a variety of reasons:

* the `\r\n` spans a 1024 block
* a 1024 byte block break happens in the middle of an utf8 sequence
* the client does not play nice and prints some binary data or corrupt encodings

I have have no good idea yet how to handle these, though.

pym/bob/cmds/build.py
```diff
@@ -458,15 +442,43 @@ def _runShell(self, step, scriptName):
...
+                else: # select timeout
+                    if proc.poll() is not None:
+                        break # proc exited
+            if proc.returncode != 0:
+                sys.stdout.flush()
                 raise BuildError("Build script {} returned with {}"
                                     .format(absRunFile, proc.returncode),
                                  help="You may resume at this point with '--resume' after fixing the error.")
         except KeyboardInterrupt:
+            sys.stdout.flush()
```
jkloetzke: What does happen if you leave that out? Looks to me like this might be a more general fix and should go to scripts.py where the exception is printed.

pym/bob/cmds/build.py
```diff
@@ -135,27 +139,7 @@ class LocalBuilder:
...
-            *)
-                set -o pipefail
-                {{
-                    {{
-                        run_script | tee -a ../log.txt
-                    }} 3>&1 1>&2- 2>&3- | tee -a ../log.txt
-                }} 3>&1 1>&2- 2>&3-
-                ;;
-        esac
+        run_script
```
jkloetzke: Maybe we could keep this code for the "manual" invocation case..?

pym/bob/cmds/build.py
rhubert: You're right. If I leave that out the exception is printed prior to the last output of the script. I'll move this to scripts.py.

pym/bob/cmds/build.py
```diff
@@ -450,23 +459,83 @@ def _runShell(self, step, scriptName):
...
+                        if r == pipe[0]:
+                            os.read(pipe[0], 1)
+                            done = True
+                        else:
+                            oldLen = 0
+                            data = bytearray(b'')
+                            # read all available data from fd
+                            blockSize = 1024
+                            while True:
+                                data = data + os.read(r, blockSize)
```
jkloetzke: Don't you have to set the non-block flag for the pty file descriptors too? I guess these calls block until the client terminates... :thinking:

rhubert: indeed - this could explain some strange behavior...  :open_mouth:

PR #87, opened 2016-12-04
misc-cmds: make use of defines and config file.

rhubert:
Make ls, query-recipe and query-scm able to use -D and -c arguments. Also added a small unit test testing all highlevel commands if they are aware of -D and -c argument.


jkloetzke:
:+1: But the test should be a black box test because that's what it basically is. They have been refactored lately. It should be quite easy to do the tests there...

PR #89, opened 2016-12-10
Doc & bash completion updates

rhubert:
some minor updates for the man pages.
bash completion available for bob help. also added --destination option for bob dev/build.

### DIFF
--- a/contrib/bash-completion
+++ b/contrib/bash-completion
@@ -34,6 +34,7 @@ __bob_complete_words()
 
 __bob_path_hash="invalid"
 __bob_path_content=( )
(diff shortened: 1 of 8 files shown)


contrib/bash-completion
```diff
@@ -34,6 +34,8 @@ __bob_complete_words()
 
 __bob_path_hash="invalid"
 __bob_path_content=( )
+BOB_COMMANDS="build dev clean help jenkins ls project status  query-scm query-recipe query-path"
```
jkloetzke: __bob_commands would be better

PR #90, opened 2016-12-11
Update configuration.rst

MrTomJerry:
just added some more info because I came across this point. :)

### DIFF
--- a/doc/manual/configuration.rst
+++ b/doc/manual/configuration.rst
@@ -774,6 +774,8 @@ packages that are built from a library: a ``-target`` packet that has the
 shared libraries needed during runtime and a ``-dev`` packet that has the
 header files and other needed files to link with this library.
 
+Bob does not support multiple or nested multiPackages in a single recipe, yet.
+
 .. _configuration-recipes-privateenv:
 
 privateEnvironment


doc/manual/configuration.rst
```diff
@@ -774,6 +774,8 @@ packages that are built from a library: a ``-target`` packet that has the
 shared libraries needed during runtime and a ``-dev`` packet that has the
 header files and other needed files to link with this library.
 
+Bob does not support multiple or nested multiPackages in a single recipe, yet.
```
rhubert: What are multiple multiPackages?! Can you give me a example of what exactly you want to do/achieve?

jkloetzke: "multiple" "multiPackage"-Entries are are not possible with yaml. That's the same as having multiple "packageScript"s. I think that can be left out. What is indeed not possible is a nested `multiPackage`:

```yaml
multiPackage:
    foo:
        multiPackage:
            bar:
                packageScript: ...
```

This is not implemented and will be rejected during schema validation. With some effort it could be implemented, though.

MrTomJerry: Good point, I was not sure about the technical background. I conducted some experiments and thought it would be a good idea to add a note in the doc.

jkloetzke:
Well, this has been implemented in the meantime. Thanks anyway.

MrTomJerry:
I heared about the implementation of it. Thanks to @rhubert  :)

PR #104, opened 2017-01-20
Fix handling of interrupted/failed steps

### LINKED ISSUE #102: Incremental build failed
Steps to reproduce:
1. build a package -> success
2. add error in sources, rebuild package -> BUILD step fails
3. revert changes to sources, rebuild -> BUILD step is *not* executed -> FAIL

The problem is that Bob does not invalidate the result and input of the build step prior to building it.

jkloetzke:
Bob did not internally prepare for the failed execution of a step. Instead the previously determined input and result of the step was kept and no information was left that the work space is in an undefined state.

This PR fixes the problem and adds a test case that previously failed exactly due to the reason above.

Fixes #102

### DIFF
--- a/pym/bob/cmds/build.py
+++ b/pym/bob/cmds/build.py
@@ -576,6 +576,14 @@ esac
                     BobState().setDirectoryState(prettySrcPath,
                         { d:s for (d,s) in checkoutState.items() if d is not None })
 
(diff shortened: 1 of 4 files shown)


rhubert:
:+1:

PR #105, opened 2017-01-23
Efficient package calculation

### LINKED ISSUE #98: Parsing of big projects is slow
Parsing of big projects with many variants is slow and memory consumption is getting a concern too. Instead of traversing the whole dependency tree the descent should be stopped if a package has already been calculated.

jkloetzke:
Restructure the internal package data structures and speed up initial calculation.

@kolewu @ThomasFeher Could you please provide some feedback if this works for you or breaks something on your side?

At the same time this cleans up the direct/indirect dependencies semantic. It is now possible to build packages that are only there due to "provideDeps" of downstream packages.

Fixes #98


kolewu:
Will do some tests today or tomorrow.

kolewu:
I have made some tests with this branch and found no problems. The initial recipe parsing for `bob ls` is longer than in the old version, but all following calls to `bob ls` are faster.

For our recipes (82 root recipes that recursively expand to 3223 with `bob ls -ra | wc -l`) the numbers for `bob ls` are:

Without cache:
```
Old version:          New version:
--------------------------------------
real	0m1,838s      real    0m3,851s
user	0m1,708s      user    0m3,700s
sys	0m0,036s      sys     0m0,060s
```

With cache:
```
Old version:          New version:
--------------------------------------
real	0m1,122s      real    0m0,237s
user	0m1,092s      user    0m0,216s
sys	0m0,024s      sys     0m0,012s    
```

So from our side and from the outside (I haven't reviewed the changes) it looks fine.

jkloetzke:
Thanks for testing. This is looking promising. :smiley: 

Just as a side note. We have just 28 root recipes but they expand to staggering 663853 packages. Without the cache the new version is already miles ahead:

```
Old version:          New version:
--------------------------------------
real    1m44.313s     real    0m15.800s
user    1m43.796s     user    0m15.612s
sys     0m0.420s      sys     0m0.140s
```

with cache the difference is also huge:

```
Old version:          New version:
--------------------------------------
real    0m7.653s      real    0m0.886s
user    0m7.364s      user    0m0.832s
sys     0m0.288s      sys     0m0.048s
```

PR #108, opened 2017-01-30
get rid of CACHE_VERSION ... 

rhubert:
... by hashing bob sources

### DIFF
--- a/pym/bob/input.py
+++ b/pym/bob/input.py
@@ -14,7 +14,7 @@
 # You should have received a copy of the GNU General Public License
 # along with this program.  If not, see <http://www.gnu.org/licenses/>.
 
-from . import BOB_VERSION
+from . import BOB_VERSION, BOB_INPUT_HASH
 from .errors import ParseError, BuildError
 from .scm import CvsScm, GitScm, SvnScm, UrlScm
 from .state import BobState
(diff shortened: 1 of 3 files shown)


pym/bob/utils.py
```diff
@@ -228,11 +228,15 @@ def close(self):
         def check(self, prefix, name, st, process):
             return process(os.path.join(prefix, name))
 
-    def __init__(self, basePath=None):
+    def __init__(self, basePath=None, ignoredDirs=None):
         if basePath:
             self.__index = DirHasher.FileIndex(basePath)
         else:
             self.__index = DirHasher.NullIndex()
+        self.__ignoredDirs = []
```
jkloetzke: You can unify `DirHasher.IGNORE_DIRS` and `ignoredDirs`. Then you only have to check for `self.__ignoreDirs` later.

```python
if ignoreDirs:
  self.__ignoreDirs = DirHasher.IGNORE_DIRS | frozenset(os.fsencode(i) for i in ignoreDirs)
else:
  self.__ignoreDirs = DirHasher.IGNORE_DIRS
```

pym/bob/input.py
```diff
@@ -15,6 +15,7 @@
 # along with this program.  If not, see <http://www.gnu.org/licenses/>.
 
 from . import BOB_VERSION
+from . import BOB_INPUT_HASH
```
jkloetzke: Should be merged into the previous line: `from . import BOB_VERSION, BOB_INPUT_HASH`

PR #109, opened 2017-02-01
show status for active scmOverrides

### LINKED ISSUE #99: Add info about active scmOverrides
Currently there is no way to tell if an 'scmOverride' matched and what has been changed. Maybe extend "bob query-scm"?

rhubert:
Fixes #99


pym/bob/input.py
```diff
@@ -469,61 +469,17 @@ def onFinish(self, env, tools, properties, package):
...
-        return scm
-
 def Scm(spec, env, overrides):
-    # resolve with environment
-    spec = { k : ( env.substitute(v, "checkoutSCM::"+k) if isinstance(v, str) else v)
-        for (k, v) in spec.items() }
-
-    # apply overrides
-    for override in overrides:
-        spec = override.mangle(spec)
```
jkloetzke: The scm mangling should stay here and should not be moved into the scms. Currently it is possible to change the scm type with a scmOverride. This could be used to rewrite a git checkout to an URL download for example.

pym/bob/cmds/build.py
```diff
@@ -887,6 +909,13 @@ def commonBuildDevelop(parser, argv, bobRoot, develop):
     elif len(results) > 1:
         print("Build results are in:\n  ", "\n   ".join(results))
 
+    print("checkoutSteps: " + str(builder.getStatistic()['checkoutSteps']))
+    print("buildSteps:    " + str(builder.getStatistic()['buildSteps']))
+    print("packageSteps:  " + str(builder.getStatistic()['packageSteps']))
+    print("scms: " + str(builder.getStatistic()['scms']))
+    print(" -> overridden: " + str(builder.getStatistic()['scmOverrides']))
+    print("   -> packages: " + str(builder.getStatistic()['scmOverriddenPackages']))
```
jkloetzke: Hmm, this list could be rather long. Maybe there should be only a summary here (eg. 6 overrides active) and which packages were affected could be printed if the particular package is built ( CHECKOUT foo/bar/baz (2 scm overrides) ). What do you think?

rhubert: Well - I wasn't happy with my solution but had no better idea... :wink: 

I like your suggestions - they show some (maybe useful) informations while not over-rolling the user.

jkloetzke:
Looks promising so far. :+1:  Let's give it some testing...

pym/bob/cmds/build.py
```diff
@@ -266,6 +267,12 @@ def __init__(self, recipes, verbose, force, skipDeps, buildOnly, preserveEnv,
         self.__bobRoot = bobRoot
         self.__cleanBuild = cleanBuild
         self.__cleanCheckout = False
+        self.__statistic = {
```
jkloetzke: That should be a class or at lease a namedtuple instead of a dict.

pym/bob/cmds/build.py
```diff
@@ -514,10 +524,21 @@ def cook(self, steps, parentPackage, checkoutOnly, depth=0):
...
+                if not o in self.__statistic['activeOverrides']:
+                    self.__statistic['activeOverrides'].append(o)
+        overrides = len(overrides)
+
         checkoutDigest = checkoutStep.getVariantId()
         if self._wasAlreadyRun(checkoutStep):
             prettySrcPath = checkoutStep.getWorkspacePath()
-            self._info("   CHECKOUT  skipped (reuse {})".format(prettySrcPath))
+            self._info("   CHECKOUT  skipped (reuse {})".format(prettySrcPath)
+                    + ((" (" + str(overrides) + " scm " + ("overrides" if overrides > 1 else "override") +")") if overrides else ""))
```
jkloetzke: The override string should be calculated once at the beginning and just used here and below...

pym/bob/scm/git.py
```diff
@@ -278,14 +281,17 @@ def status(self, workspacePath, dir, verbose=0):
...
             if status == "":
                 if verbose >= 3:
-                    print(colorize("   STATUS      {0}".format(scmdir), "32"))
+                    print(colorize("   STATUS {0}     {1}".format(overriddenStatus, scmdir), "32"))
             elif status != "u":
                 ret = 'unclean'
 
             if (status != "") and (verbose != 0):
-                print(colorize("   STATUS {0: <4} {1}".format(status, scmdir), "33"))
+                print(colorize("   STATUS {0}{1: <4} {2}".format(overriddenStatus, status, scmdir), "33"))
```
jkloetzke: Minor: that will change the indentation if an override is active.

But more importantly I'm thinking about the expected behavior of here. Unless I use `-vv` it will depend on the status of the working tree whether I see that an override is active or not. I could imagine that this may surprise users..?

rhubert: Well in a previous version there was a `STATUS O` output if verbose >= 1. But a overridden SCM is still in clear state if it's only overridden and not modified - that’s why I think bob status shouldn't show a different behavior as for non overridden clean SCMs. Also if you have a override matching to all of you SCM's (maybe because you're using a local mirror) you'll get a lot of STATUS 0 prints.

pym/bob/cmds/build.py
```diff
@@ -833,6 +869,8 @@ def commonBuildDevelop(parser, argv, bobRoot, develop):
         else:
             parser.error("Malformed define: "+define)
 
+    startTime = time.time()
```
rhubert: is this the right place? It excludes bob's runtime from duration print... :thinking:

jkloetzke: Actually it does include all relevant time here. The parsing (`recipes.parse()`) and package calculation (`recipes.generatePackages(...)`) are done below that line...

pym/bob/cmds/build.py
```diff
@@ -41,6 +42,24 @@ def hashWorkspace(step):
...
+    def addOverrides(self, overrides):
+        for o in overrides:
+            if not o in self.__activeOverrides:
+                self.__activeOverrides.append(o)
+
+    def getActiveOverrides(self):
+        return self.__activeOverrides
+
+    def __getitem__(self, item):
+        return getattr(self, item)
```
rhubert: To access the members of the `LocalBuilderStatistic` class in `commonBuildDevelop`. Without I get:
```
  File "/home/biera/projects/bob/bob/pym/bob/cmds/build.py", line 967, in commonBuildDevelop
    + str(stats.packagesDownloaded) + " downloaded.")
TypeError: 'LocalBuilderStatistic' object is not subscriptable
```
Is there a better way to do this?

jkloetzke: The problem is `str(stats['checkouts'])`. Just change it to `str(stats.checkouts)` and it should IMHO work...

pym/bob/scm/scm.py
```diff
@@ -0,0 +1,85 @@
...
+            overrides = self.getActiveOverrides()
+            if len(overrides):
+                status = "O"
+                longStatus = ""
+                for o in overrides:
+                    overrideText = str(o).rstrip().replace('\n', '\n       ')
+                    longStatus += "    > Overridden by:\n       {}\n".format(overrideText)
+
+                print(colorize("   STATUS O    {0}".format(os.path.join(workspacePath, dir)), "32"))
+                print(colorize(longStatus, "32"))
```
rhubert: Populating status was also my first idea. But all scm's have some code like:
```python
if status != ""
   return 'unclean'
```
I'll rework this as suggested - once the baby stops crying every evening... :worried:

pym/bob/cmds/build.py
rhubert: Thanks... :+1:

PR #116, opened 2017-03-13
git: change fetch command for git compatibility <1.8.2

kkendzia:
When using git with a version below 1.8.2.1 there is a error after fetching. E.g. this happens when you use CentOS or RedHat.

### DIFF
--- a/pym/bob/scm/git.py
+++ b/pym/bob/scm/git.py
@@ -93,7 +93,7 @@ if [[ -z $(git remote) ]] ; then
 fi
 # checkout only if HEAD is invalid
 if ! git rev-parse --verify -q HEAD >/dev/null ; then
-    git fetch -t origin
+    git fetch -t origin '+refs/heads/*:refs/remotes/origin/*'
     git checkout -q {REF}
 fi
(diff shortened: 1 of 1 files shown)


pym/bob/scm/git.py
```diff
@@ -93,7 +93,7 @@ def asScript(self):
 fi
 # checkout only if HEAD is invalid
 if ! git rev-parse --verify -q HEAD >/dev/null ; then
-    git fetch -t origin
+    git fetch -t origin +refs/heads/*:refs/remotes/origin/*
```
jkloetzke: This should actually be quoted with single quotes. Otherwise the shell will try to expand it as a pattern.

kkendzia: True, haven't seen it either. Will fix it tomorrow.

jkloetzke: Ping :wink:

kkendzia: I modified the commit (squashed it for better history). Happy Merging :)

jkloetzke:
Thanks :+1:

PR #122, opened 2017-03-21
Meta vars

rhubert:
This adds meta variables to recipes. They should be used to hold a common set of informations like versions, license, author. Maybe also a short description can be added as a meta var to 'search' for packages using `bob query-meta -f DESCRIPTION | grep "whatever"` ...

Not sure if it's necessary to add a new keyword - `privateEnvironment` variables basically provide the same functionality. But I think it makes sense to distinguish between a `privateEnvironment` and a `metaVar`.

Next steps:
I think bob should be aware of some variables (VERSION, LICENSE, LICENSE_FILE) and add them to the audit trail. This will make it possible to generate a list of used licenses in a project like the manifest file generated by bitbake.

Feedback welcome.

### DIFF
--- a/doc/manpages/bob-query-meta.rst
+++ b/doc/manpages/bob-query-meta.rst
@@ -1,0 +1,36 @@
+
+.. _manpage-query-meta:
+
+bob-query-meta
+==============
+
+.. only:: not man
+
+   Name
+   ----
+
+   bob-query-meta - Query metaEnvironment variables
+
+Synopsis
+--------
+
+::
+
+    bob query-meta [-h] [-D DEFINES] [-c CONFIGFILE] [-r] PACKAGE [PACKAGE ...]
+
+Description
+-----------
+
+This command lists variables from the metaEnvironment section of the recipe.
+
+Options
+-------
+
+``-c CONFIGFILE``
+    Use config File
+
+``-D DEFINES``
+    Override default environment variable
+
+``-r``
+    Also list metaEnvironment variables for all dependencies.
--- a/doc/manual/configuration.rst
+++ b/doc/manual/configuration.rst
@@ -746,6 +746,18 @@ of the respective attributes of the recipe. For example the scripts of the
 inherited class of all steps are inserted in front of the scripts of the
 current recipe. 
 
+metaEnvironment
+~~~~~~~~~~~~~~~
+
+Type: Dictionary (String -> String)
+
+metaEnvironment variables behave like :ref:`configuration-recipes-privateenv` variables.
+They overrule other environment variables and can be used in all steps, but substitution is not
+available. In addition all metaEnvironment variables are added to the audit no matter they are
+used in a step or not.
+This predestines metaEnvironment variables to add the license type or version of a package.
+
+The :ref:`manpage-query-meta` command can be used to retrieve metaEnvironment variables.
 
 multiPackage
 ~~~~~~~~~~~~
--- a/pym/bob/audit.py
+++ b/pym/bob/audit.py
@@ -84,7 +84,7 @@ class Artifact:
             'date'     : str
         },
         "env" : str,
-        "metaEnv" : { schema.Optional(str) : str },
+        schema.Optional('metaEnv') : str,
         "scms" : [ dict ],
         schema.Optional("recipes") : dict,
         "dependencies" : {
@@ -148,7 +148,7 @@ class Artifact:
         self.__defines = data["meta"]
         self.__build = data["build"]
         self.__env = data["env"]
-        self.__metaEnv = data["metaEnv"]
+        self.__metaEnv = data.get("metaEnv", {})
 
         self.__scms = []
         scms = data["scms"]
@@ -186,10 +186,12 @@ class Artifact:
             "meta" : self.__defines,
             "build" : self.__build,
             "env" : self.__env,
-            "metaEnv" : self.__metaEnv,
             "scms" : [ s.dump() for s in self.__scms ],
             "dependencies" : dependencies
         }
+
+        if self.__metaEnv:
+            ret[ "metaEnv"] = self.__metaEnv
 
         if self.__recipes is not None:
             ret["recipes"] = self.__recipes.dump()
--- a/pym/bob/audit.py
+++ b/pym/bob/audit.py
@@ -84,6 +84,7 @@ class Artifact:
             'date'     : str
         },
         "env" : str,
+        "metaEnv" : { schema.Optional(str) : str },
         "scms" : [ dict ],
         schema.Optional("recipes") : dict,
         "dependencies" : {
@@ -125,6 +126,7 @@ class Artifact:
             'date'     : str(datetime.utcnow()),
         }
         self.__env = ""
+        self.__metaEnv = {}
         self.__scms = []
         self.__deps = []
         self.__tools = {}
@@ -146,6 +148,7 @@ class Artifact:
         self.__defines = data["meta"]
         self.__build = data["build"]
         self.__env = data["env"]
+        self.__metaEnv = data["metaEnv"]
 
         self.__scms = []
         scms = data["scms"]
@@ -183,6 +186,7 @@ class Artifact:
             "meta" : self.__defines,
             "build" : self.__build,
             "env" : self.__env,
+            "metaEnv" : self.__metaEnv,
             "scms" : [ s.dump() for s in self.__scms ],
             "dependencies" : dependencies
         }
@@ -217,6 +221,9 @@ class Artifact:
     def addTool(self, name, tool):
         self.__tools[name] = tool
         self.__id = None
+
+    def addMetaEnv(self, var, value):
+        self.__metaEnv[var] = value
 
     def setSandbox(self, sandbox):
         self.__sandbox = sandbox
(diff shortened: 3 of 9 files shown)


pym/bob/input.py
```diff
@@ -1950,6 +1959,12 @@ def prepare(self, pathFormatter, inputEnv, sandboxEnabled, inputStates, inputSan
             varPrivate[key] = env.substitute(value, "privateEnvironment::"+key)
         env.update(varPrivate)
 
+        # apply meta variables
+        for (key, value) in self.__varMeta.items():
+            if key in env:
+                raise ParseError("Duplikate variable in metaVars and environment: " + key)
```
jkloetzke: Do you think this is necessary? Always giving precedence to meta variables looks viable too...

doc/manual/configuration.rst
```diff
@@ -746,6 +746,14 @@ of the respective attributes of the recipe. For example the scripts of the
 inherited class of all steps are inserted in front of the scripts of the
 current recipe. 
 
+metaVars
```
jkloetzke: How about calling it `metaEnvironment`? This would be similar to `environment` and `privateEnvironment`. After all, they're just some private environment variables that can be queried.

doc/manpages/bob-query-meta.rst
```diff
@@ -0,0 +1,38 @@
...
+-------
+
+``-c CONFIGFILE``
+    Use config File
+
+``-D DEFINES``
+    Override default environment variable
+
+``-f FILTER``
+    Filter for metaVars matching a given regex.
```
jkloetzke: Do we need a regex here? A shell pattern should be sufficient and is easier to write.

Also, the implementation is not like a filter (that throws away matching variables) but more like "grep". I would keep the '-f' switch for the future to implement custom formats like the other query commands....

doc/manpages/bob-query-meta.rst
```diff
@@ -0,0 +1,38 @@
...
+   ----
+
+   bob-query-meta - Query meta information
+
+Synopsis
+--------
+
+::
+
+    bob query-meta [-h] [-D DEFINES] [-c CONFIGFILE] [-r] [-f FILTER]
```
jkloetzke: `-r` is not explained below...

pym/bob/cmds/build.py
```diff
@@ -333,6 +333,8 @@ def _generateAudit(self, step, depth, resultHash):
         audit = Audit.create(step.getVariantId(), self._getBuildId(step, depth), resultHash)
         audit.addDefine("bob", BOB_VERSION)
         audit.addDefine("recipe", step.getPackage().getRecipe().getName())
+        for var,val in step.getPackage().getMetaVars().items():
+            audit.addDefine(var, val)
```
jkloetzke: Hmm, this may collide with the other "defines". Either we rename existing variables to something that cannot collide (e.g. `.bob`, `.recipe`) or put meta variables in a new namespace (e.g. metaEnvironment). Thoughts?

pym/bob/cmds/misc.py
```diff
@@ -131,6 +132,57 @@ def __init__(self, default, *args, **kwargs):
...
+
+    recipes = RecipeSet()
+    recipes.setConfigFiles(args.configFile)
+    recipes.parse()
+    rootPackages = recipes.generatePackages(lambda s,m: "unused", defines)
+    package = walkPackagePath(rootPackages, args.package)
+
+    def showPackage(package, recurse, done=set()):
+        # show recipes only once for each checkout variant
+        key = (package.getRecipe().getName())
```
jkloetzke: Key should be `(package.getName(), package.getPackageStep().getVariantId())`. Currently it does not match the comment above.

pym/bob/cmds/misc.py
```diff
@@ -131,6 +132,57 @@ def __init__(self, default, *args, **kwargs):
...
+        if key not in done:
+            output = { var:val for (var, val) in package.getMetaVars().items() if not args.filter or any(f.match(var) for f in _filter)}
+            if output:
+                print(package.getRecipe().getName() + " " + str(output))
```
jkloetzke: Shuldn't that be `package.getName()`? I would favor something like this:
```
package LICENSE=GPL
package MAINTAINER=John Doe
...
```
where every pair is on a separate line. That would be parse-able by other scripts.

doc/manpages/bob-query-meta.rst
rhubert: No - I'll change the output as suggested by your last comment to one line per variable and remove the `-f`

doc/manual/configuration.rst
rhubert: :+1:

pym/bob/input.py
rhubert: Sounds good. Now I'm wondering if this is worth a warning... :thinking:

pym/bob/cmds/build.py
rhubert: I prefer a separate namespace. What about renaming the existing `meta` namespace to `info` or `base`?

pym/bob/input.py
jkloetzke: The only warning I could think of is when you specify the same variable in `privateEnvironment` and `metaEnvironment`. The latter will always win, so it makes no sense. I don't think a warning should be triggered if the variable was inherited from somewhere else.

pym/bob/cmds/build.py
jkloetzke: A separate namespace is certainly better. There is already a "build" namespace that just holds the uname output plus the build date and time. May we'll move the existing keys there to free the "meta" namespace?

rhubert: hmm...I don't think the existing "build" and "meta" namespaces should be merged. But in the audit there is also "env" --> "metaEnv" as a new namespace ?

pym/bob/cmds/jenkins.py
```diff
@@ -403,6 +405,7 @@ def dumpStepAuditGen(self, step):
         # wrap lines
         ret = []
         line = "bob-audit-engine"
+        print(cmd)
```
jkloetzke: Debug left over...

doc/manual/configuration.rst
```diff
@@ -746,6 +746,17 @@ of the respective attributes of the recipe. For example the scripts of the
...
 
+metaEnvironment
+~~~~~~~~~~~~~~~
+
+Type: Dictionary (String -> String)
+
+metaEnvironment variables behave like :ref:`configuration-recipes-privateenv` variables.
+They overrule other environment variables and can be used in all steps. In addition all
+metaEnvironment variables are added to the audit no matter they are used in a step or not.
+This predestines metaEnvironment variables to add the license type or version of a package
```
jkloetzke: Missing period at end of sentence. Maybe we should mention that variable substitution is not available...

pym/bob/cmds/misc.py
```diff
@@ -20,6 +20,7 @@
 from ..errors import BuildError
 import argparse
 import codecs
+import re
```
jkloetzke: Not needed anymore...

pym/bob/cmds/misc.py
```diff
@@ -131,6 +132,50 @@ def __init__(self, default, *args, **kwargs):
...
+        else:
+            parser.error("Malformed define: "+define)
+
+    recipes = RecipeSet()
+    recipes.setConfigFiles(args.configFile)
+    recipes.parse()
+    rootPackages = recipes.generatePackages(lambda s,m: "unused", defines)
+    package = walkPackagePath(rootPackages, args.package)
+
+    def showPackage(package, recurse, done=set()):
```
jkloetzke: The `done` default parameter is not a good idea. Even though it's not a problem here this idiom should be avoided. The default parameter exists only once. As the function modifies `done` this will actually change the default parameter too: http://docs.python-guide.org/en/latest/writing/gotchas/#mutable-default-arguments

pym/bob/cmds/jenkins.py
```diff
@@ -375,6 +375,8 @@ def dumpStepAuditGen(self, step):
             "-D", "jenkins-node", '"$NODE_NAME"',
             "-D", "jenkins-build-url", '"$BUILD_URL"'
         ]
+        for (var, val) in step.getPackage().getMetaEnv().items():
+            cmd.extend(["--metaEnv", var, val])
```
jkloetzke: `val` must be quoted. Maybe using a short option (`-E`?) could save some space...

pym/bob/scripts.py
```diff
@@ -228,6 +234,7 @@ def cmd():
         except ValueError:
             raise BuildError("Invalid digest argument")
         if args.env is not None: gen.setEnv(args.env)
+        if args.metaEnv is not None: gen.setMetaEnv(args.metaEnv)
```
jkloetzke: I couldn't find `setMetaEnv` in `Audit`class...

pym/bob/audit.py
```diff
@@ -84,6 +84,7 @@ class Artifact:
             'date'     : str
         },
         "env" : str,
+        "metaEnv" : { schema.Optional(str) : str },
```
jkloetzke: We have to make the metaEnv key optional. Otherwise existing audit trails in old workspaces will fail builds.

pym/bob/audit.py
```diff
@@ -146,6 +148,7 @@ def load(self, data):
         self.__defines = data["meta"]
         self.__build = data["build"]
         self.__env = data["env"]
+        self.__metaEnv = data["metaEnv"]
```
jkloetzke: Use `data.get("metaEnv", {})` because the key might not exist due to the change above...

jkloetzke:
Thanks a lot :+1:

PR #127, opened 2017-04-06
input: add lock to yamlcache

### LINKED ISSUE #118: Parallel query-scm fails with backtrace
```
  File "/home/user/BobBuildTool/pym/bob/scripts.py", line 114, in bob
  File "/home/user/BobBuildTool/pym/bob/scripts.py", line 60, in __queryscm
    doQuerySCM(*args, **kwargs)
  File "/home/user/BobBuildTool/pym/bob/cmds/misc.py", line 176, in doQuerySCM
    recipes.parse()
  File "/home/user/BobBuildTool/pym/bob/input.py", line 2477, in parse
    self.__cache.open()
  File "/home/user/BobBuildTool/pym/bob/input.py", line 2831, in open
    self.__shelve = shelve.open(".bob-cache.shelve")
  File "/usr/lib/python3.4/shelve.py", line 239, in open
    return DbfilenameShelf(filename, flag, protocol, writeback)
  File "/usr/lib/python3.4/shelve.py", line 223, in __init__
    Shelf.__init__(self, dbm.open(filename, flag), protocol, writeback)
  File "/usr/lib/python3.4/dbm/__init__.py", line 94, in open
    return mod.open(file, flag, mode)
_gdbm.error: [Errno 11] Resource temporarily unavailable
```

rhubert:
According to https://docs.python.org/3/library/shelve.html#restrictions The shelve module does not support concurrent read/write access to shelved objects. Added a lockfile to prevent duplicate opens.

Fixes #118

### DIFF
--- a/pym/bob/input.py
+++ b/pym/bob/input.py
@@ -17,7 +17,7 @@
 from . import BOB_VERSION, BOB_INPUT_HASH, DEBUG
 from .errors import BuildError, ParseError
 from .scm import CvsScm, GitScm, SvnScm, UrlScm, ScmOverride, auditFromDir
-from .state import BobState
+from .state import BobState, BobLock
 from .tty import colorize, WarnOnce
 from .utils import asHexStr, joinScripts, sliceString, compareVersion, binLstat
 from abc import ABCMeta, abstractmethod
@@ -2684,17 +2684,20 @@ class RecipeSet:
     def generateTree(self, envOverrides={}, sandboxEnabled=False):
         (env, cacheKey) = self.__getEnvWithCacheKey(envOverrides, sandboxEnabled)
         cacheName = ".bob-tree.dbm"
+        BobLock().getNamedLock(cacheName + ".lock")
 
         # try to load persisted tree
         roots = TreeStorage.load(cacheName, cacheKey)
-        if roots is not None:
-            return roots
 
-        # generate and convert
-        roots = self.__generatePackages(lambda p, m: "unused", env, cacheKey, sandboxEnabled)
+        if roots is None:
+            # generate and convert
+            roots = self.__generatePackages(lambda p, m: "unused", env, cacheKey, sandboxEnabled)
 
-        # save tree cache
-        return TreeStorage.create(cacheName, cacheKey, roots)
+            # save tree cache
+            roots = TreeStorage.create(cacheName, cacheKey, roots)
+
+        BobLock().releaseNamedLock(cacheName + ".lock")
+        return roots
 
 
 class TreeStorage:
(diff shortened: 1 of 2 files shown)


jkloetzke:
I think the real solution must be a bit more sophisticated. Unfortunately I'm not really sure how it should look like. In general it should be possible to use Bob in parallel in a workspace unless the instances collide. So many instances can read the workspace (ls, query-*) but only one instance is allowed to change the state (dev, build).

Right now there are three mutable files in the workspace:
* `.bob-state.pickle` - already protected by a lock file. Whoever gets the lock owns the workspace for commands that use the state (dev/build, jenkins). Not optimal yet but close.
* `.bob-packages*` - not protected but updated atomically. Not an issue IMHO.
* `.bob-tree.dbm` - not protected and updated incrementally

The proper solution is a multiple-read-single-writer lock for the last file. However this might look like...

rhubert:
The issue is describing a crash while accessing the YamlCache `.bob-cache.shelve`. I don't see why a lock for `.bob-tree.dbm` fixes this? :thinking: 
For the YamlCache I think it's fine to use a blocking wait for exclusive lock as this is opened only a short period of time?

jkloetzke:
You're absolutely right. The Yaml cache can be locked always for the short time. The same must be done for `.bob-tree.dbm` too. The current solution is lacking two things, though. A proper fall back on windows and error handling if bob is/was interrupted.

In the long run the dbm files could opened for reading under a shared lock. If the database needs updating an exclusive lock is taken and the dbm file is opened for writing. But this can be done as a separate optimization.

rhubert:
I added the handling for interrupted bob. On windows (msys2) fcntl seems to work in the same was as on linux systems...

pym/bob/input.py
```diff
@@ -2684,17 +2684,20 @@ def generatePackages(self, nameFormatter, envOverrides={}, sandboxEnabled=False)
...
+        if roots is None:
+            # generate and convert
+            roots = self.__generatePackages(lambda p, m: "unused", env, cacheKey, sandboxEnabled)
 
-        # save tree cache
-        return TreeStorage.create(cacheName, cacheKey, roots)
+            # save tree cache
+            roots = TreeStorage.create(cacheName, cacheKey, roots)
+
+        BobLock().releaseNamedLock(cacheName + ".lock")
```
jkloetzke: Hmm, at this stage we still have opened the `TreeStorage` database as read only. If the user changes the recipes and calls Bob in a separate shell then there will still be a read/write conflict. It's not very likely but I think the real solution must use shared/exclusive locks. In this case the lock must be stored in the `TreeStorage` object to persist until the database is closed.

pym/bob/state.py
```diff
@@ -255,6 +256,49 @@ def setBuildState(self, digest2Dir):
...
+            except OSError as e:
+                from .tty import colorize
+                from sys import stderr
+                print(colorize("Warning: cannot unlock workspace: "+str(e), "33"),
+                    file=stderr)
+            finally:
+                del self.__lockfiles[name]
+
+    def finalize(self):
+        for name, fd in self.__lockfiles.items():
```
jkloetzke: This will fail. While iterating the object must not change. A solution could be to store the keys in a list and iterate that:

```python
for name in list(self.__lockfiles.keys()):
    ...
```

PR #129, opened 2017-04-10
faster Jenkins job calculation by shorten description

kkendzia:
As discussed in a meeting last week I implemented a parameter to make "jenkins push" and "jenkins export" faster. By inserting a return condition for already visited packages, calculating Jobs just takes some seconds.

### DIFF
--- a/pym/bob/cmds/jenkins.py
+++ b/pym/bob/cmds/jenkins.py
@@ -35,6 +35,7 @@ import re
 import ssl
 import sys
 import textwrap
+import time
 import urllib.parse
 import xml.etree.ElementTree
 
@@ -982,7 +983,15 @@ class JobNameCalculator:
         return self.__regexJobName.sub('_', self.getJobDisplayName(step)).lower()
 
 
-def _genJenkinsJobs(step, jobs, nameCalculator, archiveBackend, seenPackages):
+def _genJenkinsJobs(step, jobs, nameCalculator, archiveBackend, seenPackages, allVariantIds,
+                    shortdescription=False):
+
+    if step.isPackageStep() and shortdescription:
+        if step.getVariantId() in allVariantIds:
+            return
+        else:
+            allVariantIds.add(step.getVariantId())
+
     name = nameCalculator.getJobInternalName(step)
     if name in jobs:
         jj = jobs[name]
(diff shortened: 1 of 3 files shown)


pym/bob/cmds/jenkins.py
```diff
@@ -38,6 +38,7 @@
 import urllib.parse
 import xml.etree.ElementTree
 
+
```
jkloetzke: There are some stray white space changes. They should be removed.

pym/bob/cmds/jenkins.py
```diff
@@ -982,7 +983,15 @@ def getJobInternalName(self, step):
         return self.__regexJobName.sub('_', self.getJobDisplayName(step)).lower()
 
 
-def _genJenkinsJobs(step, jobs, nameCalculator, archiveBackend, seenPackages):
+def _genJenkinsJobs(step, jobs, nameCalculator, archiveBackend, seenPackages, allVariantIds = set(),
+                    shortdescription=False):
```
jkloetzke: No need for default parameters IMHO. This function is internal to the module and all call sites can be adjusted. BTW, the `allVariantIds` default parameter is dangerous: http://docs.python-guide.org/en/latest/writing/gotchas/#mutable-default-arguments

pym/bob/cmds/jenkins.py
```diff
@@ -1797,6 +1808,8 @@ def doJenkinsSetOptions(recipes, argv):
                         help="Undefine environment variable override")
     parser.add_argument("--credentials", help="Credentials UUID for SCM checkouts")
     parser.add_argument('--authtoken', help='AuthToken for remote triggering jobs')
+    parser.add_argument('--shortdescription', action='store_true', default=False,
+                        help='Don\'t calculate all paths for description')
```
jkloetzke: I think for `set-options` we need the inverse `--longdescription` too. This is like the "keep", "download" and "sandbox" options below. Only one switch can be given at the same time. Without any option the "description" switch is kept unchanged.

test/test_jenkins_set_options.py
```diff
@@ -144,3 +144,9 @@ def testDelRoot(self):
         self.executeBobJenkinsCmd("push -q myTestJenkins")
         send = self.jenkinsMock.getServerData()
         assert(send[0][0] == '/job/test/doDelete')
+
+    def testShortDescription(self):
```
rhubert: Minor: Maybe you want to extend the testcase (add a more complex recipe) and parse the uploaded description with `--shortdescription` and `--longdescription`?

kkendzia:
I think it should be fine now. Feel free to review!

pym/bob/cmds/jenkins.py
```diff
@@ -1887,6 +1906,8 @@ def doJenkinsSetOptions(recipes, argv):
         config['credentials'] = args.credentials
     if args.authtoken is not None:
         config['authtoken'] = args.authtoken
+    if args.shortdescription is not False:
```
jkloetzke: The condition is slightly wrong. It should be `if args.shortdescription is not None` because the parameter can be `True` or `False` depending what the user gave or `None` if no option was given. In this case the parameter should stay unchanged.

kkendzia:
Fixed the False/None issue

PR #130, opened 2017-04-19
global config, disable logfile

rhubert:


doc/manual/configuration.rst
```diff
@@ -1015,13 +1015,18 @@ Plugins are loaded in the same order as listed here. For each name in this
 section there must be a .py-file in the ``plugins`` directory next to the
 recipes. For a detailed description of plugins see :ref:`extending-plugins`.
 
-User configuration (default.yaml)
----------------------------------
+User configuration (default.yaml / .bob-settings.yaml)
```
jkloetzke: I'd like to go even one step further an do it like git and be xdg-compatible: https://specifications.freedesktop.org/basedir-spec/basedir-spec-latest.html. Quoting git:

```
FILES
       If not set explicitly with --file, there are four files where git config will search for configuration options:

       $(prefix)/etc/gitconfig
           System-wide configuration file.

       $XDG_CONFIG_HOME/git/config
           Second user-specific configuration file. If $XDG_CONFIG_HOME is not set or empty, $HOME/.config/git/config will be used. Any
           single-valued variable set in this file will be overwritten by whatever is in ~/.gitconfig. It is a good idea not to create this file
           if you sometimes use older versions of Git, as support for this file was added fairly recently.

       ~/.gitconfig
           User-specific configuration file. Also called "global" configuration file.

       $GIT_DIR/config
           Repository specific configuration file.
```
For Bob this would be:
* `/etc/bobdefault.yaml`
* `$XDG_CONFIG_HOME/bob/default.yaml` resp. `~/.config/bob/default.yaml`
* `./default.yaml`

doc/manual/configuration.rst
```diff
@@ -1195,3 +1200,11 @@ Specifies alias names for packages::
 
    alias:
       myApp: "host/files/group/app42"
+
+build-args
```
jkloetzke: I'd rather find dedicated configuration options more elegant. Something like
```yaml
config:
  dev:
    mode: "build-only"
    logfiles: False
```

They would be taken as defaults that can be overridden by the various command line options.

doc/manpages/bob-build-dev.rst
```diff
@@ -61,6 +61,9 @@ Options
 ``-b, --build-only``
     Don't checkout, just build and package
 
+``-^b, --no-build-only``
```
rhubert: There are many options without a counter part... if all of them can be set in the user-config they'll all need one... even the new `--no-logfile` option will need a `--enable-logfile`...

IMO a option to ignore (`-i`) the `config` section of the user-config's would give a greater usability? :thinking:

contrib/bash-completion
```diff
@@ -79,7 +79,7 @@ __bob_cook()
    elif [[ "$prev" = "-c" ]] ; then
       COMPREPLY=( $(compgen -a -f -o dirnames "$cur" | cut -f 1 -d '.' ) )
    else
-      __bob_complete_path "--destination -e -f --force --clean-checkout -n --no-deps -b --build-only --clean --incremental --resume -q --quiet --upload --download --sandbox --no-sandbox -v --verbose"
+      __bob_complete_path "--destination -e -f --force --clean-checkout -n --no-deps -b -^b --build-only --clean --incremental --resume -q --quiet --upload --download --sandbox --no-sandbox -v --verbose --no-logfiles"
```
jkloetzke: `-^b` looks like a left over

pym/bob/input.py
```diff
@@ -2504,7 +2533,7 @@ def __parseUserConfig(self, fileName):
             self.__archive = cfg["archive"]
         self.__scmOverrides.extend([ ScmOverride(o) for o in cfg.get("scmOverrides", []) ])
         self.__aliases.update(cfg.get("alias", {}))
-
+        self.__userConfig.update(cfg.get("config", {}))
```
jkloetzke: I think this is a bit too coarse. This way a config file can override only all switches of `dev` or `build` but not individual ones.

pym/bob/input.py
```diff
@@ -2463,6 +2489,9 @@ def __parse(self):
         self.__createSchemas()
 
         # user config(s)
+        self.__parseUserConfig("/etc/bobdefault.yaml")
```
jkloetzke: I think we got a problem here with possible includes of other files. Currently they are included relative to the current working directory. That's broken for such global files.

The most sane solution would be to do it again like git where includes are done relative to the current file. Unfortunately this will break backwards compatibility. Another solution could be to invent a special syntax like `./` for relative includes. But this is easy to get wrong and comes unexpected to the uninformed user. Thoughts?

pym/bob/input.py
```diff
@@ -2224,6 +2225,23 @@ def validate(self, data):
 
 class RecipeSet:
 
+    BUILD_DEV_SCHEMA = schema.Schema(
+        {
+            schema.Optional('destination') : str,
+            schema.Optional('force') : bool,
+            schema.Optional('no-deps') : bool,
+            schema.Optional('build-only') : bool,
+            schema.Optional('checkout-only') : bool,
```
jkloetzke: `--build-only` and `--checkout-only` are mutually exclusive. Maybe `build-mode` with `normal`, `build-only` and `checkout-only` as possible values?

pym/bob/input.py
```diff
@@ -2224,6 +2225,23 @@ def validate(self, data):
...
+            schema.Optional('clean') : bool,
+            schema.Optional('quiet') : int,
+            schema.Optional('verbose') : int,
```
jkloetzke: `quiet` and `verbose` could be collapsed into `verbosity`. This is then the default that is incremented by `-v` and decremented by `-q`.

doc/manual/configuration.rst
```diff
@@ -1195,3 +1209,37 @@ Specifies alias names for packages::
...
+~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~
+
+Set default build arguments here. See :ref:`manpage-dev` or
+:ref:`manpage-build` for details.::
+
+
+    dev:
+        no-logfile: True
+        build-only: True
+    build:
```
jkloetzke: Isn't it

```
config:
   dev:
      ...
```
? I'm thinking if `command` would be more appropriate. `config` is quite generic.

pym/bob/input.py
rhubert: Do you think it is useful to have includes for system / user default's? If so I think we should break backwards compatibility cause relative to current working directory intuitive. Otherwise it may be easier to make them only available for the workspace default.yaml.

sirhcel: User defaults look quite useful to me. At least when I'm thinking about aliases. I'm frequently working on the same packages in different repositorys and so global aliases would be quite handy. What about a `~/.bobuser.yaml`? The concept of user defaults seems quite popular and understood to me.

rhubert: you can use `XDG_CONFIG_HOME/bob/default.yaml` or if XDG_CONFIG_HOME is not set `~/.config/bob/default.yaml` and add your alias.

sirhcel: Thank you very much @rhubert! I didn't know about that. :)

rhubert: That's what this PR is about.... :wink:

rhubert: I added another commit making includes relative and cleaned up the history...

doc/manual/configuration.rst
```diff
@@ -1195,3 +1209,47 @@ Specifies alias names for packages::
...
+
+The following table lists possible arguments and their type:
+
+=============== ===================================================================
+Key             Type
+=============== ===================================================================
+destination     String
+force           Boolean
+no-deps         Boolean
+build-mode      "build-only" or "checkout-only"
```
jkloetzke: What about the regular "normal" build mode?

pym/bob/cmds/build.py
```diff
@@ -961,6 +984,29 @@ def commonBuildDevelop(parser, argv, bobRoot, develop):
...
+            cfg = recipes.getUserConfig().get('dev', {})
+        else:
+            cfg = recipes.getUserConfig().get('build', {})
+        args.destination = cfg.get('destination', args.destination)
+        args.force = cfg.get('force', args.force)
+        args.no_deps = cfg.get('no-deps', args.no_deps)
+        buildmode = cfg.get('build-mode', None)
+        if buildmode == 'build-only':
+            args.build_only = True
+            args.checkout_only = False
```
jkloetzke: What about defining the arguments the following way:?

```python
    group.add_argument('-b', '--build-only', dest='build_mode', default="normal",
        action='store_const', const='build-only', 
        help="Don't checkout, just build and package")
    group.add_argument('-B', '--checkout-only', dest='build_mode',
        action='store_const', const='checkout-only', 
        help="Don't build, just check out sources")
    group.add_argument('--normal', dest='build_mode',
        action='store_const', const='normal', 
        help="Checkout, build and package")
```

Then it's easy to merge it with the default settings.

contrib/bash-completion
```diff
@@ -79,7 +79,7 @@ __bob_cook()
    elif [[ "$prev" = "-c" ]] ; then
       COMPREPLY=( $(compgen -a -f -o dirnames "$cur" | cut -f 1 -d '.' ) )
    else
-      __bob_complete_path "--destination -e -f --force --clean-checkout -n --no-deps -b --build-only --clean --incremental --resume -q --quiet --upload --download --sandbox --no-sandbox -v --verbose"
+      __bob_complete_path "--destination -e -f --force --clean-checkout -n --no-deps -b --build-only --clean --incremental --resume -q --quiet --upload --download --sandbox --no-sandbox -v --verbose --no-logfiles -i --ignore-userconfig"
```
jkloetzke: `-i --ignore-userconfig` left over

doc/manpages/bob-build.rst
```diff
@@ -12,11 +15,11 @@ Synopsis
 --------
 
 ::
-
     bob build [-h] [--destination DEST] [-f] [-n] [-b | -B]
-              [--clean | --incremental] [--resume] [-q] [-v] [-D DEFINES]
-              [-c CONFIGFILE] [-e NAME] [-E] [--upload] [--download MODE]
-              [--sandbox | --no-sandbox] [--clean-checkout]
+              [--clean | --incremental] [--resume] [-q] [-v]
+              [--no-logfiles] [-D DEFINES] [-c CONFIGFILE] [-i] [-e NAME]
```
jkloetzke: `-i` left over

doc/manpages/bob-dev.rst
```diff
@@ -14,11 +16,11 @@ Synopsis
 ::
 
     bob dev [-h] [--destination DEST] [-f] [-n] [-b | -B]
-            [--clean | --incremental] [--resume] [-q] [-v] [-D DEFINES]
-            [-c CONFIGFILE] [-e NAME] [-E] [--upload] [--download MODE]
-            [--sandbox | --no-sandbox] [--clean-checkout]
+            [--clean | --incremental] [--resume] [-q] [-v] [--no-logfiles]
+            [-D DEFINES] [-c CONFIGFILE] [-i] [-e NAME] [-E] [--upload]
```
jkloetzke: ditto

pym/bob/cmds/build.py
```diff
@@ -920,25 +944,29 @@ def commonBuildDevelop(parser, argv, bobRoot, develop):
         help="Decrease verbosity (may be specified multiple times)")
     parser.add_argument('-v', '--verbose', default=0, action='count',
         help="Increase verbosity (may be specified multiple times)")
+    parser.add_argument('--no-logfiles', default=None, action='store_true',
+        help="Disable logFile generation.")
     parser.add_argument('-D', default=[], action='append', dest="defines",
         help="Override default environment variable")
     parser.add_argument('-c', dest="configFile", default=[], action='append',
         help="Use config File")
+    parser.add_argument('-i', '--ignore-userconfig', default=False, action='store_true',
```
jkloetzke: ditto

rhubert: removed them...

PR #134, opened 2017-05-03
jenkins: use binary artifact server for sharing

rhubert:
Added a option to make use of the binary artifacts server for sharing
artifacts between jobs. This may lower the load on jenkins master.

### DIFF
--- a/pym/bob/cmds/jenkins.py
+++ b/pym/bob/cmds/jenkins.py
@@ -554,7 +554,8 @@ class JenkinsJob:
         prepareCmds.append("}")
         prepareCmds.append("")
         whiteList = []
-        whiteList.extend([ JenkinsJob._tgzName(d) for d in self.__deps.values()])
+        if options.get("artifacts.copy", "jenkins") == "jenkins":
+            whiteList.extend([ JenkinsJob._tgzName(d) for d in self.__deps.values()])
         whiteList.extend([ JenkinsJob._buildIdName(d) for d in self.__deps.values()])
         whiteList.extend([ d.getWorkspacePath() for d in self.__checkoutSteps.values() ])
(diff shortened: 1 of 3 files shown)


pym/bob/cmds/jenkins.py
```diff
@@ -628,6 +631,10 @@ def dumpXML(self, orig, nodes, windows, credentials, clean, options, date, autht
                 xml.etree.ElementTree.SubElement(
                     cp, "doNotFingerprintArtifacts").text = "true"
 
+            if archiveSharing:
+                prepareCmds.append("\n# download deps\n# =============")
+                prepareCmds.append(self.__archive.download(d, JenkinsJob._buildIdName(d), JenkinsJob._tgzName(d)))
```
jkloetzke: This is not indented correctly (not part of the loop above). It will download just the last dependency. Likewise, the download should not be done for shared artifacts if the condition does not match.

pym/bob/cmds/jenkins.py
```diff
@@ -1059,6 +1067,10 @@ def genJenkinsJobs(recipes, jenkins):
     archiveHandler = getArchiver(recipes)
     archiveHandler.wantUpload(config.get("upload", False))
     archiveHandler.wantDownload(config.get("download", False))
+    if config.get("archiveSharing", False):
+        if not archiveHandler.canUploadJenkins() or not archiveHandler.canDownloadJenkins():
+            print("No archive for up and download but archiveSharing enabled! Exiting..")
```
jkloetzke: `raise ParseError`?

jkloetzke:
This looks like it's already working. I would suggest to make this an extended option, though. In hindsight this should have been done for the `nodes`, `windows`,  `credentials`, `clean`, and `authtoken` options too. They all tweak the way the jobs are configured. The smoking gun is that we have to add another parameter again to `dumpXML()`...

PR #136, opened 2017-05-07
Allow `bob help` as an alternative to `bob --help`

sirhcel:
I attempted several times to use `bob help` as I'm used from Git & Co. So why not accepting this command verb without any further options as the other cool cats around do?

Yes, this predates the responsibility of the help command implemented in help.py. As this one only handles the help verb with an additional command there should be no trap for young players. A unified/hierarchical argument parser might streamline this.

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -1,5 +1,6 @@
 *~
 *.pyc
+.*.swp
 .coverage
 __pycache__/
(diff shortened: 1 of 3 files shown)


jkloetzke:
Thanks. :+1:

PR #143, opened 2017-06-29
Implement path queries

jkloetzke:
Most Bob commands are working on sets of packages. They can be specified by a query language that loosely resembles Unix paths for the common case and XPath for more advanced features.


jkloetzke:
@kolewu @ThomasFeher could you give this branch some testing on your side if it breaks anything? There is some extensive documentation in the new `bobpaths` manpage. Any input on the feature itself is very much welcome... :wink:

PR #150, opened 2017-08-21
build: fix force option

### LINKED ISSUE #147: --force option not working anymore.
When trying to force build a single package with

```shell
bob build path/to/package/ -f -b -n --clean
```

the building of the dependencies is correctly skipped (because of `-n`, though the messages state "due to --build-only" which might be OK here, too) but building of the target package is skipped, as well:

```shell
PACKAGE   skipped (deterministic output in work/…/dist/1/workspace)
```

I guess this is not the intended behaviour of the `--force` switch.

My bob version is

```shell
Bob version 0.12.0-71-gb74766d
```

rhubert:
Even in case of a downloaded package-result the force option should trigger
a rebuild.

Fixes #147

### DIFF
--- a/pym/bob/cmds/build.py
+++ b/pym/bob/cmds/build.py
@@ -819,8 +819,9 @@ esac
             wasDownloaded = False
             if ( (not checkoutOnly) and packageBuildId and (depth >= self.__downloadDepth) ):
                 # prune directory if we previously downloaded/built something different
-                if (oldInputBuildId is not None) and (oldInputBuildId != packageBuildId):
-                    print(colorize("   PRUNE     {} (build-id changed)".format(prettyPackagePath), "33"))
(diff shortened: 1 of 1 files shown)


pym/bob/cmds/build.py
```diff
@@ -848,7 +848,7 @@ def _cookPackageStep(self, packageStep, checkoutOnly, depth):
             # downloads are not possible anymore. Even if the package was
             # previously downloaded the oldInputHashes will be None to trigger
             # an actual build.
-            if not wasDownloaded:
+            if (not wasDownloaded) or self.__force:
```
jkloetzke: This will also cause re-builds of packages that just been downloaded, e.g. when building in an empty workspace with `-f`. I think the solution is to add `or self.__force` at line 822. This will wipe out the result and either download or build the package.

PR #152, opened 2017-08-28
graph: highlevel command for dependency graphs

rhubert:
bob graph can be used to make package dependencies visible. For now
'dot' graphs and 'd3' graphs are available, where d3 is the default.

D3 graphs are interactive (zoom, drag, clickable, with hover effects)
graphs using the d3.v4 javascript library.


rhubert:
Some time ago there has been some mails on the mailing list about the dependency graph. Maybe this PR is a better place to discuss this feature.

A dot graph did not scale for large projects. That's why I looked around and found the [d3.js](https://d3js.org/) library. With this you can create interactive graphs. I looked at many examples at [Mike Bostock's block](https://bl.ocks.org/mbostock) and did some try & error programming because of missing JS knowledge. Maybe anybody has some more experience with this javascript and can give some hints.

Here is a list of my open tasks:
- [ ] add header to d3 generated graph (bob version, recipes revision, ...)
- [ ] make d3 graph resize the svg on browser windows change event.
- [ ] add a highlight argument. Browser search is not sufficient on large projects to find one / more than one node.
- [ ] add the possibility to use a local version of d3.v4.min instead of loading it from d3js.org
- [ ] Documentation

pym/bob/cmds/graph.py
```diff
@@ -0,0 +1,379 @@
...
+    parser.add_argument('package', nargs='?', help="Sub-package that is the root of the graph")
+
+    parser.add_argument('-D', default=[], action='append', dest="defines",
+        help="Override default environment variable")
+    parser.add_argument('-c', dest="configFile", default=[], action='append',
+        help="Use config File")
+    parser.add_argument('-e', '--exclude', default=[], action='append', dest="excludes",
+        help="Do not show this Package. (And all it's deps)")
+    parser.add_argument('-n', '--max-depth', type=int, default=-1, help="Max depth.")
+    parser.add_argument('-t', '--type', default='d3', help="Graph type. d3 (default) or dot")
```
jkloetzke: You could use `choices=['d3', 'dot']` here to restrict the valid settings.

pym/bob/cmds/graph.py
```diff
@@ -0,0 +1,379 @@
...
+        elif len(d) == 2:
+            defines[d[0]] = d[1]
+        else:
+            parser.error("Malformed define: "+define)
+
+    recipes = RecipeSet()
+    recipes.setConfigFiles(args.configFile)
+    recipes.parse()
+    packages = recipes.generatePackages(lambda s,m: "unused", defines)
+    package = packages.walkPackagePath(args.package)
```
jkloetzke: It would be nice to support the query syntax too via `queryPackagePath`. The legacy `walkPackagePath` should be removed in the future...

pym/bob/cmds/graph.py
```diff
@@ -0,0 +1,379 @@
...
+    if (maxdepth != -1) and (level >= maxdepth):
+        return
+
+    for d in package.getDirectDepSteps():
+        if (d.getPackage()._getId(), package._getId()) not in links:
+            for e in excludes:
+                if (e.match(d.getPackage().getName())):
+                    break
+            else:
+                if (type == 'dot'):
```
jkloetzke: That's not so elegant. Maybe this whole function could be rewritten as a generator function by `yield`'ing the packages?

rhubert:
I think I'm done.

.coveragerc
```diff
@@ -0,0 +1,15 @@
+[run]
+omit =
+    # graph.py
+    pym/bob/cmds/graph.py
```
jkloetzke: I think we could add a simple black-box-test that generates a dot and d3 graph based on some simple recipes. That  doesn't need to be complete but only a smoke test that shows that the code at least still runs. Whether it generates something sensible cannot be tested, that's clear.

.coveragerc
```diff
@@ -0,0 +1,15 @@
...
+
+[report]
+# Regexes for lines to exclude from consideration
+exclude_lines =
+    # Have to re-enable the standard pragma
+    pragma: no cover
+
+    # Don't complain if tests don't hit defensive assertion code:
+    raise
+    except
```
jkloetzke: Hmm, that feels like cheating...

pym/bob/input.py
```diff
@@ -1930,6 +1930,13 @@ class RecipeSet:
             schema.Optional('clean_checkout') : bool,
         })
 
+    GRAPH_SCHEMA = schema.Schema(
+        {
+            schema.Optional('options') : schema.Schema({str : schema.Or(str, bool)}),
+            schema.Optional('type') : schema.Or("d3", "dot"),
+            schema.Optional('max_depth') : int,
```
jkloetzke: Maybe add some lines in the configuration.rst documentation?

pym/bob/cmds/graph.py
```diff
@@ -0,0 +1,555 @@
...
+    destination = os.path.join(os.getcwd(), 'graph') if not args.destination else args.destination
+    if not os.path.exists(destination):
+        os.makedirs(destination)
+
+    for p in args.packages:
+        # convert the package-query into a valid file name
+        filename = "".join([c for c in p if c.isalpha() or c.isdigit() or c==' ']).rstrip()
+
+        print(">>", colorize(p, "32;1"))
+        print(colorize("   PROJECT   {} ({})".format(p, args.type), "32"))
```
jkloetzke: Copy'n'paste: PROJECT

PR #153, opened 2017-09-10
Add git remotes

ThomasFeher:
**Problem this is trying to solve:**
Each time a new source directory is created, because of new tags or branches, or because of creating a new "workspace", i.e. cloning the recipe repository to a new location, the user has to re-add his personal remote repositories.

**Solution:**
Add a `scmOverride` with the users remotes that will be added each time the corresponding repositories are checked out. This would look something like:
```yaml
# file: default.yaml
scmOverride:
    match:
        url: "git@central/repo/url.git"
    remotes:
        my_remote: "git@my/remote/url.git"
        second_developer: "git@second/developers/url.git"
```
The key will be used as the remote name and the value as the remote URL.

This is a quick implementation to see if it works as expected. And it seems that it does.

So is this something you would consider to merge? If it is, I would go ahead and add documentation and care for corner-case, if there are any, etc. Or could this problem be solved differently?

### DIFF
--- a/pym/bob/scm/git.py
+++ b/pym/bob/scm/git.py
@@ -24,6 +24,7 @@ import os, os.path
 import re
 import schema
 import subprocess
+from pipes import quote
 
 class GitScm(Scm):
 
@@ -36,7 +37,9 @@ class GitScm(Scm):
         schema.Optional('tag') : str,
         schema.Optional('commit') : str,
         schema.Optional('rev') : str,
+        schema.Optional(schema.Regex('^remote-.*')) : str,
     })
+    REMOTE_PREFIX = "remote-"
 
     def __init__(self, spec, overrides=[]):
         super().__init__(overrides)
@@ -45,6 +48,7 @@ class GitScm(Scm):
         self.__branch = None
         self.__tag = None
         self.__commit = None
+        self.__remotes = {}
         if "rev" in spec:
             rev = spec["rev"]
             if rev.startswith("refs/heads/"):
@@ -66,9 +70,16 @@ class GitScm(Scm):
             # nothing secified at all -> master branch
             self.__branch = "master"
         self.__dir = spec.get("dir", ".")
+        # convert remotes into separate dictionary
+        for key, val in spec.items():
+            if key.startswith(GitScm.REMOTE_PREFIX):
+                stripped_key = key[len(GitScm.REMOTE_PREFIX):] # remove prefix
+                if stripped_key == "origin":
+                    raise ParseError("Invalid remote name: " + stripped_key)
+                self.__remotes.update({stripped_key : val})
 
     def getProperties(self):
-        return [{
+        properties = [{
             'recipe' : self.__recipe,
             'scm' : 'git',
             'url' : self.__url,
@@ -81,8 +92,43 @@ class GitScm(Scm):
                     ("refs/heads/" + self.__branch))
             )
         }]
+        for key, val in self.__remotes.items():
+            properties[0].update({GitScm.REMOTE_PREFIX+key : val})
+        return properties
 
     def asScript(self):
+        remotes_array = """
+# create an array of all remotes for this repository
+declare -A BOB_GIT_REMOTES=( [origin]={URL} )""".format(URL=quote(self.__url))
+        # add additional remotes to array
+        for name, url in self.__remotes.items():
+            remotes_array += """
+BOB_GIT_REMOTES[{NAME}]={URL}""".format(NAME=quote(name), URL=quote(url))
+        # create script to handle remotes
+        remotes_script = """
+(
+    {REMOTES_ARRAY}
+    # remove remotes from array that are already known to Git
+    while read -r REMOTE_NAME ; do
+        # check for empty variable in case no remote at all is specified
+        if [ -z "$REMOTE_NAME" ]; then
+            continue
+        fi
+        # check if existing remote is configured
+        if [ "${{BOB_GIT_REMOTES[$REMOTE_NAME]+_}}" ]; then
+            # check if URL has changed
+            if [ ! "${{BOB_GIT_REMOTES[$REMOTE_NAME]}}" == "$(git ls-remote --get-url $REMOTE_NAME)" ]; then
+                git remote set-url "$REMOTE_NAME" "${{BOB_GIT_REMOTES[$REMOTE_NAME]}}"
+            fi
+            # it is configured, therefore no need to keep in list
+            unset "BOB_GIT_REMOTES[$REMOTE_NAME]"
+        fi
+    done <<< "$(git remote)"
+    # add all remaining remotes in the array to the repository
+    for REMOTE_NAME in "${{!BOB_GIT_REMOTES[@]}}" ; do
+        git remote add "$REMOTE_NAME" "${{BOB_GIT_REMOTES[$REMOTE_NAME]}}"
+    done
+)""".format(REMOTES_ARRAY=remotes_array)
         if self.__tag or self.__commit:
             return """
 export GIT_SSL_NO_VERIFY=true
@@ -90,16 +136,16 @@ if [ ! -d {DIR}/.git ] ; then
     git init {DIR}
 fi
 cd {DIR}
-# see if we have a remote
-if [[ -z $(git remote) ]] ; then
-    git remote add origin {URL}
-fi
+{REMOTES}
 # checkout only if HEAD is invalid
 if ! git rev-parse --verify -q HEAD >/dev/null ; then
     git fetch -t origin '+refs/heads/*:refs/remotes/origin/*'
     git checkout -q {REF}
 fi
-""".format(URL=self.__url, REF=self.__commit if self.__commit else "tags/"+self.__tag, DIR=self.__dir)
+""".format(URL=self.__url,
+           REF=self.__commit if self.__commit else "tags/"+self.__tag,
+           DIR=self.__dir,
+           REMOTES=remotes_script)
         else:
             return """
 export GIT_SSL_NO_VERIFY=true
@@ -116,7 +162,11 @@ else
         exit 1
     fi
 fi
-""".format(URL=self.__url, BRANCH=self.__branch, DIR=self.__dir)
+{REMOTES}
+""".format(URL=self.__url,
+           BRANCH=self.__branch,
+           DIR=self.__dir,
+           REMOTES=remotes_script)
 
     def asDigestScript(self):
         """Return forward compatible stable string describing this git module.
--- a/pym/bob/scm/git.py
+++ b/pym/bob/scm/git.py
@@ -219,12 +219,12 @@ fi
     # Get GitSCM status. The purpose of this function is to return the status of the given directory
     #
     # return values:
-    #  - error: the scm is in a error state. Use this if git returned a error code.
-    #  - dirty: SCM is dirty. Could be: modified files, switched to another branch/tag/commit/repo, unpushed commits
-    #  - clean: same branch/tag/commit as specified in the recipe and no local changes.
-    #  - empty: directory is not existing
+    #  - error: The SCM is in a error state. Use this if git returned a error code.
+    #  - dirty: SCM is dirty. Could be: modified files, switched to another branch/tag/commit/repo, unpushed commits.
+    #  - clean: Same branch/tag/commit as specified in the recipe and no local changes.
+    #  - empty: Directory is not existing.
     #
-    # This function is called when build with --clean-checkou. 'error' and 'dirty' scm's are moved to attic,
+    # This function is called when build with --clean-checkout. 'error' and 'dirty' SCMs are moved to attic,
     # while empty and clean directories are not.
     def status(self, workspacePath, dir):
         scmdir = os.path.join(workspacePath, dir)
--- a/doc/manual/configuration.rst
+++ b/doc/manual/configuration.rst
@@ -536,6 +536,7 @@ git `Git`_ project               | ``url``: URL of remote repository
                                  | ``tag``: Checkout this tag (optional, overrides branch attribute)
                                  | ``commit``: SHA1 commit Id to check out (optional, overrides branch or tag attribute)
                                  | ``rev``: Canonical git-rev-parse revision specification (optional, see below)
+                                 | ``remote-*``: additional remote repositories (optional, see below)
 svn `Svn`_ repository            | ``url``: URL of SVN module
                                  | ``revision``: Optional revision number (optional)
 cvs CVS repository               | ``cvsroot``: repository location ("``:ext:...``", path name, etc.)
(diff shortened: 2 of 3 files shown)


jkloetzke:
I really like the idea. Maybe it would be more elegant to add it as first class property to the git scm? That way is would be nothing special for scmOverrides but you could already set it in the recipe too. Currently all properties of SCMs are assumed to be strings. That makes a bit cumbersome to handle multiple remotes. What about this..?

```yaml
checkoutSCM:
    scm: git
    url: "git@central/repo/url.git"
    remote-my_remote: "git@my/remote/url.git"
    remote-second_developer: "git@second/developers/url.git"
```

That way nothing needs to be changed in the `scmOverrides` area. Only the git SCM would have to be extended.

Corner cases that I can think of out of the blue:
* what should happen if the remotes are changed/added/deleted?
* what happens if the "origin" remote is set too (bail out? overwrite "url"? ignore?)
* does it contribute to the variant-id?

ThomasFeher:
That would mean that the remotes have to be set in the recipe, doesn't it? But that is not feasible, because each developer wants to add different remotes, his own fork in the first place. So it should just be a personal configuration that does not contribute to the recipe.

> what should happen if the remotes are changed/added/deleted?
> does it contribute to the variant-id?

Doesn't need to be considered if its a scmOverride.

> what happens if the "origin" remote is set too (bail out? overwrite "url"? ignore?)

Mh, I'd go for an error in this case.

jkloetzke:
> That would mean that the remotes have to be set in the recipe, doesn't it? But that is not feasible, because each developer wants to add different remotes, his own fork in the first place. So it should just be a personal configuration that does not contribute to the recipe.

No, you can set it in your configuration via `scmOverrides` too. The SCM overrides are a method to "patch" the properties in SCMs. You can add, delete and change properties of any SCM as you like. SCM overrides should not provide additional semantics, though. They are a general purpose mechanism to change SCMs without editing the recipes.

So my example was a bit unfortunate. A better one (leading to the same result) would have been:

```yaml
scmOverride:
    match:
        url: "git@central/repo/url.git"
    add:
        remote-my_remote: "git@my/remote/url.git"
        remote-second_developer: "git@second/developers/url.git"
```

> > what should happen if the remotes are changed/added/deleted?
> > does it contribute to the variant-id?
> 
> Doesn't need to be considered if its a scmOverride.

I'm not so sure. What should happen if the user changes it's scmOverrides and rebuilds? Should the checkout be run and the remotes are adapted? Or is it ok to skip the checkout step because it didn't change the actual source code?

> > what happens if the "origin" remote is set too (bail out? overwrite "url"? ignore?)
> 
> Mh, I'd go for an error in this case.

Yes, that may be best option.

ThomasFeher:
Ok, now I understand. Yes, I like that solution.

Having a nested dictionary for the remotes would be better instead of encoding that hierarchy in a string, but if that would mean too much refactoring I'm also fine with that.

> What should happen if the user changes it's scmOverrides and rebuilds? Should the checkout be run and the remotes are adapted? Or is it ok to skip the checkout step because it didn't change the actual source code?

In my opinion it would be convenient to have the remote added automatically, but not necessary. So we could omit that for now and add it later if there is a reason to.

ThomasFeher:
Ok, I had another look at the code. Inside the class GitScm there seems to be no reason to not use a nested dictionary. So I decided to give that solution another try, and the outcome was that adding `remotes` as first class member was only adding one line of code (see the last commit).

Replacing that dictionary with a variable number of strings grouped by a common prefix would involve writing some more code, though. The schema would be something like `schema.Optional(schema.Regex('^remote-.*')) : str`, that's still Ok. But as the class members can not change dynamically (or can they? I am no Python expert) one would need to create a member dictionary anyway and loop over all entries in `spec` applying the same regex search again. Generating the properties dictionary in `getProperties()` would also be more complex. So I wonder if it is really worth the hassle? Maybe you could elaborate a bit more on why only strings should be allowed. Thanks.

jkloetzke:
The reason is that SCM properties are assumed to be only strings. There are multiple places that make this assumption:

* The scmOverrides: what should happen if you use "replace" feature? How would you delete a certain remote.
* The query-scm command. It can display any property. How would we handle the "remote" nested property?
* How should the "matchScm" string function handle the nested property?

All these places have to be extended to handle the nested "remote" property properly. Sticking to plain string key-value pairs would keep them as is.

> The schema would be something like `schema.Optional(schema.Regex('^remote-.*')) : str`, that's still Ok.

Exactly.

> But as the class members can not change dynamically (or can they? I am no Python expert) one would need to create a member dictionary anyway and loop over all entries in spec applying the same regex search again. Generating the properties dictionary in getProperties() would also be more complex.

Sure, you would have to parse the "remote-" properties into a dict and re-create them in `getProperties`. But that is only a couple of lines I guess.

ThomasFeher:
Implemented changes as discussed. Not yet implemented:
* check for wrong remote "origin"
* documentation

I have no idea what goes wrong in the test for "status", though.

pym/bob/input.py
```diff
@@ -1958,7 +1958,8 @@ class RecipeSet:
                         'pattern' : str,
                         'replacement' : str
                     })
-                })
+                }),
+                schema.Optional(schema.Regex('^remote-.*')) : str,
```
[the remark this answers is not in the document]
ThomasFeher: deleted

pym/bob/scm/scm.py
```diff
@@ -24,6 +24,7 @@ def __init__(self, override):
         self.__del = override.get("del", [])
         self.__set = override.get("set", {})
         self.__replaceRaw = override.get("replace", {})
+        self.__remotes = override.get("remotes", {})
```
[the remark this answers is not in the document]
ThomasFeher: deleted

jkloetzke:
I've fixed the 'status' black box test on the master branch and now the test case output is:

```
Run black box tests...
   status ... FAIL (log follows...)
...
    15     CHECKOUT  dev/src/root/1/workspace (3 scm overrides)
    16  ../script: line 36: syntax error near unexpected token `fi'
    17  ../script: line 36: `    fi'
...
```

The culprit is the empty `else/fi` clause:

```bash
export GIT_SSL_NO_VERIFY=true
if [ -d git/.git ] ; then
    cd git
    if [[ $(git rev-parse --abbrev-ref HEAD) == "master" ]] ; then
        git pull --ff-only
    else
        echo "Warning: not updating git because branch was changed manually..." >&2
    fi
else
    if ! git clone -b master file:////tmp/tmp.WRstSsMKWS/ git ; then
        rm -rf git/.git git/*
        exit 1
    else
        
    fi
fi
```

ThomasFeher:
> The culprit is the empty else/fi clause:

Ok, I fixed that using a "noop". Probably not the most elegant way, but it keeps the bash script inside the python code readable.

ThomasFeher:
Added another test to have a positive delta for code coverage.

ThomasFeher:
Added check for remote "origin" and first draft for documentation.

Anything missing?

(I'll do an auto-squash before merging)

pym/bob/input.py
```diff
@@ -1958,7 +1958,7 @@ class RecipeSet:
                         'pattern' : str,
                         'replacement' : str
                     })
-                })
+                }),
```
jkloetzke: Stray left-over.

ThomasFeher:
> The documentation seems to be missing. 😮

Ups, forgot to push the commit. Should now be there.

> This looks already good. One thing that bothers me, though, is the fact that the remotes are only applied on the initial checkout. I guess some people may be surprised if their changes in the configuration are ignored unless the source folder is deleted. Would it be feasible to always add/update the remotes on every checkout?

I'll have a look at that.

ThomasFeher:
Updated documentation (missed adding a hunk in the previous commit).

Implemented updating remotes on each build. It's not so easy to keep that readable and have not too much script code duplication. I tried to find a compromise, though. If a branch is configured and no remotes then the `git remote` command is executed without any need, but I haven't found a good solution to that, maybe it's just too late now. :)

jkloetzke:
I've had a look and I think the usage of the `=~` operator is not so optimal. I'm not sure what remote names are allowed in git but it looks like some escape logic for special characters must be added. Maybe something like the following is more robust? (untested!)

```bash
declare -A BOB_GIT_REMOTES=( [origin]="url1" [my_remote]="url2" )
while IFS='\n' read i ; do
   unset ${BOB_GIT_REMOTES[$i]}
done < $(git remote)
for i in "${!BOB_GIT_REMOTES[@]}" ; do
   git remote add "$i" "${BOB_GIT_REMOTES[$i]}"
done
```

:arrow_right: Hold all remotes in an associative array, remove all entries that are already known to git and add the remaining ones.

One more thing: what should happen if the user changes the URL of an additional remote? Right now nothing will happen...

ThomasFeher:
I implemented proposed changes with some adjustments (and learned a lot about shell scripts).

> what should happen if the user changes the URL of an additional remote?

If the URL is changed, then this change should be applied to the repository. Also removing remotes from the list should remove them from the repository, shouldn't it? I'll add both features.

pym/bob/scm/git.py
```diff
@@ -81,25 +91,61 @@ def getProperties(self):
...
+        remotes_check_add_all = remotes_array + """
+# remove remotes from array that are already known to Git
+while read -r REMOTE_NAME ; do
+    # check for empty variable in case no remote at all is specified
+    if [ -z "$REMOTE_NAME" ]; then
+        continue
+    fi
+    # check if URL has changed
+    if [ ! "${BOB_GIT_REMOTES[$REMOTE_NAME]}" == "$(git remote get-url $REMOTE_NAME)" ]; then
+        git remote set-url "$REMOTE_NAME" "${BOB_GIT_REMOTES[$REMOTE_NAME]}"
```
jkloetzke: This should be moved below where it is clear that the remote is actually known to Bob. Otherwise the access to `"${BOB_GIT_REMOTES[$REMOTE_NAME]}"` will fail if the user added a remote manually.

pym/bob/scm/git.py
```diff
@@ -81,25 +91,61 @@ def getProperties(self):
...
+    fi
+    # check if existing remote is configured
+    if [ "${BOB_GIT_REMOTES[$REMOTE_NAME]+_}" ]; then
+        # it is configured, therefore no need to keep in list
+        unset "BOB_GIT_REMOTES[$REMOTE_NAME]"
+    else
+        # not configured but existing, therefore remove it
+        git remote remove "$REMOTE_NAME"
```
jkloetzke: I don't think this is a good idea. Previously it was perfectly ok for the user to add any remotes he likes and Bob would not touch it. Now we would remove them. As this might break some users work flows I would rather not touch them...

pym/bob/scm/git.py
```diff
@@ -81,25 +91,61 @@ def getProperties(self):
...
         }]
+        for key, val in self.__remotes.items():
+            properties[0].update({GitScm.REMOTE_PREFIX+key : val})
+        return properties
 
     def asScript(self):
+        remotes_array = """
+# create an array of all remotes for this repository
+declare -A BOB_GIT_REMOTES=( [origin]="{URL}" )""".format(URL=self.__url)
+        if self.__remotes:
```
jkloetzke: No need to check `self.__remotes` here.

pym/bob/scm/git.py
kolewu: Or you have to distinguish somehow between an unknown remote and one that has been configured before but is now removed from the configuration. The latter could be safely deleted.

jkloetzke: You're right. Maybe the `.git` directory could be abused to store the list of remotes that were configured by Bob. But I guess it will add quite some logic...

pym/bob/scm/git.py
ThomasFeher: Of course. Removed.

pym/bob/scm/git.py
```diff
@@ -81,25 +91,61 @@ def getProperties(self):
...
+            # add additional remotes to array
+            for name, url in self.__remotes.items():
+                remotes_array += """
+BOB_GIT_REMOTES[{NAME}]="{URL}" """.format(NAME=name, URL=url)
```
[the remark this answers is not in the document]
ThomasFeher: added

pym/bob/scm/git.py
ThomasFeher: Moved, hopefully I got it right, this time.

pym/bob/scm/git.py
ThomasFeher: Sounds complicated. I'd rather not implement that. Removed removing of remotes. :)

pym/bob/scm/git.py
```diff
@@ -81,25 +92,57 @@ def getProperties(self):
...
             )
         }]
+        for key, val in self.__remotes.items():
+            properties[0].update({GitScm.REMOTE_PREFIX+key : val})
+        return properties
 
     def asScript(self):
+        remotes_array = """
+# create an array of all remotes for this repository
+declare -A BOB_GIT_REMOTES=( [origin]="{URL}" )""".format(URL=quote(self.__url))
```
jkloetzke: The double-quotes in `[origin]="{URL}"` must be removed now. The `quote()` function will insert them as needed and this will collide if there are already some... (same below)

pym/bob/scm/git.py
jkloetzke: :smile: Ok for me. If there's a need it can be implemented later...

pym/bob/scm/git.py
ThomasFeher: done

ThomasFeher:
> One more thing that I completely overlooked: the remotes_check_add_all script snippet should run in a sub-shell, that is it should be put in parenthesis (( ... )). That prevents tainting the environment of the remaining checkout script by the used shell variables in this part.

Done

jkloetzke:
I've just tested it on Debian Jessie at work and it tripped on "git remote get-url" which is not supported yet in git 2.1.4. The alternative is "git ls-remote --get-url <remote>": https://stackoverflow.com/a/16880000 Could you change this?

With this changed it seems to work as intended. :+1: I think you can squash the commits after this last fix and I'll merge it if you're ok with it.

ThomasFeher:
> I've just tested it on Debian Jessie at work and it tripped on "git remote get-url" which is not supported yet in git 2.1.4. The alternative is "git ls-remote --get-url ": https://stackoverflow.com/a/16880000 Could you change this?

Done

> With this changed it seems to work as intended. 👍 I think you can squash the commits after this last fix and I'll merge it if you're ok with it.

Squashed into three separate commits, but I could also make it only one, if that is preferred.

jkloetzke:
I've took the freedom to remove an empty commit (seemed to be left "git rebase" even though it should have been garbage collected) and adjusted the commit messages to follow the rest of the repository. I hope you don't mind.

PR #155, opened 2017-09-26
Live-Build-Id support

jkloetzke:
Live-build-ids query the current state of a SCM server. These ids can be translated into the real build-id of a checkout step if this was cached on the artifact server. The final result is that, given a properly populated binary server, the local build does not need to checkout sources anymore before an artifact download can be started.


pym/bob/cmds/build.py
```diff
@@ -898,21 +918,110 @@ def _cookPackageStep(self, packageStep, checkoutOnly, depth):
...
+            BobState().setBuildId(key, bid)
+
+        return bid
+
+    def __getCheckoutStepBuildId(self, step, depth):
+        ret = None
+        v = (self.__verbose >= -1) and (self.__verbose < 1)
+
+        # try to use live build-ids for checkout steps
+        if not os.path.exists(step.getWorkspacePath()) and \
```
rhubert: I'm doing some tests with this and now I'm at a point where I miss my sources.
```
bob dev module_A
-> QUERY_SRC dep1
-> CHECKOUT module_A
```

I want to change something in dep1 -> no code available :(   Can you add a build argument to force the checkout?

jkloetzke: :rofl: I almost expected that but I've not had the chance so far to test it myself extensively. Either we add an explicit option (`--always-checkout`?) or we could resort to `--checkout-only` for the job. The drawback would be that would have to call Bob twice:

```
$ bob dev module_A
$ bob dev --checkout-only module_A/dep1
# hack...
$ bob dev module_A
```

Once the sources were fetched Bob will always use the local ones. Not sure if this is works well enough, though.

rhubert: I vote for the `--always-checkout`. Or / And a regex matching white-list for modules in my `user.yaml` where I always want to checkout the code.

jkloetzke: Added a `--always-checkout` option that takes a regex for packages that should be always checked out. Let's see if this works well in practice...

kkendzia:
Would like to see it merged ;)

jkloetzke:
Sure. I've pushed the last changes just yesterday. I think it's finished but especially with the latest changes I want to give it a day or two of some testing...

kkendzia:
Ah ok, haven't seen that you commited something. Yeah, give it some days!

jkloetzke:
No problem. You don't get any indication if commits are amended or if the branch is rebased. This is especially disturbing as a maintainer because I don't always get a trigger if a pull request is updated...

Issue #133, opened 2017-05-03
jenkins: Don't discard builds when package is added by --add-root


kkendzia:
When you add a non-root-recipe-package to jenkins with --add-root, the generated job discards old builds. I think they should persist - is it intentional or not? When the recipe has "root: True" it won't be generated with this feature.

Code line: https://github.com/BobBuildTool/bob/blob/master/pym/bob/cmds/jenkins.py#L473

jkloetzke:
You're right. I even think this was the behaviour some time ago. The `__isRoot` field should be set to `True` for all packages that were added by `--add-root` to keep their artifacts by default.

kkendzia:
I will submit a pull request then, probably next week.

PR #157, opened 2017-10-04
jenkins: 'discard builds' deactivated for 'roots'

kkendzia:
At the moment archived artifacts are deleted if the associated recipe isn't a root recipe. If you use 'jenkins add-root' this 'root' will be discarded.

- does not matter if jenkins root is root recipe

### DIFF
--- a/pym/bob/cmds/jenkins.py
+++ b/pym/bob/cmds/jenkins.py
@@ -137,7 +137,7 @@ class JenkinsJob:
         self.__displayName = displayName
         self.__nameCalculator = nameCalculator
         self.__recipe = recipe
-        self.__isRoot = recipe.isRoot()
+        self.__isRoot = False
         self.__archive = archiveBackend
         self.__checkoutSteps = {}
         self.__buildSteps = {}
@@ -155,6 +155,9 @@ class JenkinsJob:
 
     def isRoot(self):
         return self.__isRoot
+
+    def makeRoot(self):
+        self.__isRoot = True
 
     def getDescription(self, date):
         description = [
@@ -1045,11 +1048,12 @@ class JobNameCalculator:
 
 
 def _genJenkinsJobs(step, jobs, nameCalculator, archiveBackend, seenPackages, allVariantIds,
-                    shortdescription=False):
+                    shortdescription):
 
     if step.isPackageStep() and shortdescription:
         if step.getVariantId() in allVariantIds:
-            return
+            name = nameCalculator.getJobInternalName(step)
+            return jobs[name]
         else:
             allVariantIds.add(step.getVariantId())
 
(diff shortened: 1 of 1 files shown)


jkloetzke:
This fixes #133, right? It would be nice to include this (`Fixes #133.`) into the commit message as last line. This will automatically close the bug when being merged.

There is one corner case that is not handled yet. Suppose some simple recipes:

```
/
└── root
    └── foo
        └── bar
```

In case of

```
$ bob jenkins add -r /root/foo/bar -r /root/foo ...
```

we would create a job for "foo" first with the `__isRoot` flag being `False`. There is no way to set it to `True` when processing `/root/foo`. So I would propose to add a `makeRoot()` method to `JenkinsJob`, return `jj` from `_genJenkinsJobs` and simply call `makeRoot()` for all roots in `genJenkinsJobs`.

kkendzia:
Sorry, but I don't get why `/root/foo` has `__isRoot == false`

I've tested it again and both `/root/foo/bar`and `/root/foo` are root

jkloetzke:
Argh, bad example. Seems it was already too late yesterday. :wink: 

Better one:

```
$ bob ls -r
/
└── bar
    └── baz
        └── foo
```

With your branch and

```
bob jenkins set-options test --reset --add-root bar/baz/foo --add-root bar
```

the "foo"-Job will still discard old builds. The problem is that the Job generation starts with "bar" (due to name sorting) and this will already create a job for "foo". Then if the "bar/baz/foo" root is generated there is already a `JenkinsJobs` instance for "foo" with `__isRoot` being false...

kkendzia:
Ah, now I got it. Because of recursive calls.

jkloetzke:
> Ah, now I got it. Because of recursive calls.

Exactly. That's why it may be better to return the `JenkinsJob` from `_genJenkinsJobs` and mark them as "root" in `genJenkinsJobs`...

kkendzia:
- Added `Fixes #133.` to commit message
- implemented makeRoot function for JenkinsJob class

pym/bob/cmds/jenkins.py
```diff
@@ -1091,6 +1095,8 @@ def _genJenkinsJobs(step, jobs, nameCalculator, archiveBackend, seenPackages, al
                 _genJenkinsJobs(sandboxStep, jobs, nameCalculator, archiveBackend,
                                 seenPackages, allVariantIds, shortdescription)
 
+        return jj
```
jkloetzke: For consistency we should always return `jj`. Otherwise we're implicitly returning `None` for checkout- and build-steps. It doesn't matter at the moment but better not add a pit-fall.

jkloetzke:
Thanks. :+1:

PR #158, opened 2017-10-08
Add optimistic mode to drop the fsync.

stefanreuther:
bob will quite often flush to disk using `fsync`. This overhead becomes significant when using small projects and a rotating hard disk.  The protection it provides is comparatively minor, because if the system actually crashes, object files created by the compiler will be in an indeterminate state anyway; the compiler does not do `fsync`.  Thus, this adds an option `--optimistic` to disable the `fsync`.

-----

I spent about three seconds to invent the name, better ideas are appreciated. I would actually remove the `fsync` completely, but didn't want to leap ahead so far already.

### DIFF
--- a/pym/bob/scm/cvs.py
+++ b/pym/bob/scm/cvs.py
@@ -70,25 +70,26 @@ class CvsScm(Scm):
         #   cvs checkout: existing repository /home/stefan/cvsroot does not match /home/stefan/cvsroot/cxxtest
         #   cvs checkout: ignoring module cxxtest
         # Thus, we have to trick it with a symlink.
+        # Workaround 2: 'cvs co' does not have a '-P' option like 'cvs up' has.
+        # This option removes empty (=deleted) directories.
+        # We therefore use a 'cvs up' after the initial 'cvs co', to get the same behaviour for the initial and subsequent builds.
         if re.match('^:ext:', rootarg) and self.__dir == '.':
             return """
 # Checkout or update
-if [ -d CVS ]; then
-   {prefix}cvs -qz3 -d '{rootarg}' up -dP {revarg} .
-else
+if ! [ -d CVS ]; then
    ln -s . __tmp$$
    {prefix}cvs -qz3 -d '{rootarg}' co {revarg} -d __tmp$$ '{module}'
    rm __tmp$$
 fi
+{prefix}cvs -qz3 -d '{rootarg}' up -dP {revarg} .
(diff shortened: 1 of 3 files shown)


jkloetzke:
The `fsync` is not so much about the Bob state vs. the workspace state but having any state at all in case of a system crash. Without the `fsync` it is possible to end up with an unreadable state (https://stackoverflow.com/a/41362774). The worst thing that can happen is that you have to re-create the whole workspace...

Anyway, I think this shouldn't be a command line option but an entry in default.yaml et. al. You could set the option in `/etc/bobdefault.yaml` or `~/.config/bob/default.yaml` for all your projects. I guess that is what most users want.

As for the name I would call it `nosync`. But I'm not good at naming things either. :wink:

stefanreuther:
> The worst thing that can happen is that you have to re-create the whole workspace...

My point is that you have to do that anyway after a crash, because if the machine crashes mid-build, your object files will be in an indeterminate state, thanks to the compiler not doing `fsync`. Thus, the perfectly valid state that tells you "everything compiled" will point into a mess of compiler output.

To get real "transactional" builds if you need that, you'd probably have to up the syncing to something like "mark directory dirty, `fsync`, build, `sync`, mark directory clean+built, `fsync`". I wouldn't want this in a normal build, but I can imagine that it can be useful. I wrote this patch when porting a bunch of tiny utilities to bob (one or two *.c files plus one Makefile each), where disabling the `fsync`s halved the build times. That's probably not a typical project either.

(I'll look into doing that in the config files later.)

jkloetzke:
> > The worst thing that can happen is that you have to re-create the whole workspace...
>
> My point is that you have to do that anyway after a crash, because if the machine crashes mid-build, your object files will be in an indeterminate state, thanks to the compiler not doing `fsync`. Thus, the perfectly valid state that tells you "everything compiled" will point into a mess of compiler output.

It's not about the build- or dist-workspaces. They can be in any intermediate state and you can for sure not trust them. What I'm concerned about are the src-workspaces. All the information about checkouts done ages before may be lost. Especially if you have made changes to multiple repositories (the usual change/compile cycle) it can be a PITA if you have to start from scratch. Because if the state was lost Bob will try a fresh checkout and bail out because there are already files at the checkout location.

stefanreuther:
Just noticed that there is a package "eatmydata" to disable `fsync` using LD_PRELOAD. Using that reduces "run-tests.sh" time to one-third. So I guess there's no need to add a special case in bob for that.

PR #159, opened 2017-10-10
WIP: input: filter duplicate steps while calculation - not afterwards

kkendzia:
I noticed that you add all results into a list and filter it afterwards. Why not checking this before adding them? It would be a little performance improvement - I noticed that it will be done later, too, with provideDeps, but at first I wanted to hear your opinion.

### DIFF
--- a/pym/bob/input.py
+++ b/pym/bob/input.py
@@ -1783,7 +1783,7 @@ class Recipe(object):
         env.update(self.__metaEnv)
 
         # filter duplicate results, fail on different variants of same package
-        self.__filterDuplicateSteps(results)
+        results = self.__filterDuplicateSteps(results)
 
         # record used environment and tools
         env.touch(self.__packageVars | self.__packageVarsWeak)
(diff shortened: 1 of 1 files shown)


jkloetzke:
Adding the results to a list is not so heavy but the naive duplicate removal is expensive because it is O(n^2). I guess optimizing this can save some time on big projects. It has to be a list, though, because order is important. That's why you cannot simply use a dict and that's what is tripping the black box tests in the pull request.

Either we add a dedicated helper class that does the filtering upfront (based on `collections.OrderedDict` comes to my mind) or `__filterDuplicateSteps` is fixed by using a helper `dict`. I'm fine with any solution as long it's the faster variant. :wink:

jkloetzke:
I took the freedom to amend the commit to fix the CI failures. It definitely reduces the parsing time significantly. If you're ok with that I would merge it in this form...

kolewu:
Where did you see a significant reducing of the parsing time? At least with `bob ls` they are comparable for our recipes repo with 131 root recipes.
Initial run (without cache) and then subsequent run
```sh
rm .bob-*
time bob ls
time bob ls
```

```
real	0m7,493s
user	0m6,983s
sys	0m0,065s

real	0m0,161s
user	0m0,124s
sys	0m0,038s
```

jkloetzke:
The patch will only show an improvement if `provideDeps` is used pervasively with a deep dependency hierarchy. Our structure is quite deep and many dependencies are passed upwards. With 53 root recipes it reduces the parsing time from 34s to 27s. Comparing Bob 0.12.1 with 0.13.0-rc1 shows almost no difference here because the introduction of the query language increased the time in the "ls" command.

PR #164, opened 2017-10-23
doc: fix wording of "recipes" directory

### LINKED ISSUE #163: Contradiction in manual: relativeIncludes
The manual contains two contradictory statements on the old behavior of includes in file [policies.rst](https://github.com/BobBuildTool/bob/blob/ed3ea71f46e91bd1708e4f971a55932779b5ad01/doc/manual/policies.rst):

Compare this from the description (first paragraph)
> Versions of Bob before 0.13 included these files always relative to the root of the project configuration.

to the following in the description for **Old behaviour**

> Include further files from `default.yaml` and command line passed files relative to the current recipes directory.

So it's *root of the project configuration* vs. *current recipes directory* and IMHO the first one is correct or at least the second variant is ambiguous, because it will normally be read as *the directory of the current recipe* and not *the directory where the recipe folder is located*.

jkloetzke:
The word "recipes" was used interchangeably for the project root
directory (with config.yaml, default.yaml, recipes and classes
directory) and the "recipes" directory inside that itself. Solve this
ambiguity by using a consistent naming of the "project root directory"
which, among other things, contains the "recipes directory".

Fixes #163.

### DIFF
--- a/doc/manual/configuration.rst
+++ b/doc/manual/configuration.rst
@@ -2,7 +2,7 @@ Configuration
 =============
 
 When building packages Bob executes the instructions defined by the recipes.
-All recipes are located relative to the working directory in the ``recipes``
+All recipes are located relative to the project root directory in the ``recipes``
 subdirectory. Recipes are YAML files with a defined structure. The name of the
 recipe and the resulting package(s) is derived from the file name by removing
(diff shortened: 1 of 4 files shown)


kolewu:
👍 Looks good.

PR #165, opened 2017-10-25
WIP: build: make project command available for multiple package input

kkendzia:
For some plugins it would be useful to generate multiple packages at once with `bob project`.
- using query language
- add '--build' parameter for release path formatting
- you can use now multiple packages for projects instead of one (i'd rather call the plugin always with a list instead of seperating between single/multiple packages)

### DIFF
--- a/pym/bob/cmds/build.py
+++ b/pym/bob/cmds/build.py
@@ -1344,7 +1344,7 @@ def doDevelop(argv, bobRoot):
 def doProject(argv, bobRoot):
     parser = argparse.ArgumentParser(prog="bob project", description='Generate Project Files')
     parser.add_argument('projectGenerator', nargs='?', help="Generator to use.")
-    parser.add_argument('package', nargs='?', help="Sub-package that is the root of the project")
+    parser.add_argument('packages', nargs='+', help="(Sub-)packages to generate")
     parser.add_argument('args', nargs=argparse.REMAINDER,
                         help="Arguments for project generator")
 
@@ -1365,6 +1365,8 @@ def doProject(argv, bobRoot):
         help="Do not build (bob dev) before generate project Files. RunTargets may not work")
     parser.add_argument('-b', dest="execute_buildonly", default=False, action='store_true',
         help="Do build only (bob dev -b) before generate project Files. No checkout")
+    parser.add_argument('--build', default=False, action='store_true',
+        help="Do path calculations as if 'bob build', not as 'bob dev'")
     group = parser.add_mutually_exclusive_group()
     group.add_argument('--sandbox', action='store_true', default=False,
         help="Enable sandboxing")
@@ -1383,6 +1385,7 @@ def doProject(argv, bobRoot):
             parser.error("Malformed define: "+define)
 
     recipes = RecipeSet()
+    recipes.defineHook('releaseNameFormatter', LocalBuilder.releaseNameFormatter)
     recipes.defineHook('developNameFormatter', LocalBuilder.developNameFormatter)
     recipes.defineHook('developNamePersister', LocalBuilder.developNamePersister)
     recipes.setConfigFiles(args.configFile)
@@ -1391,12 +1394,18 @@ def doProject(argv, bobRoot):
     envWhiteList = recipes.envWhiteList()
     envWhiteList |= set(args.white_list)
 
-    nameFormatter = recipes.getHook('developNameFormatter')
-    developPersister = recipes.getHook('developNamePersister')
-    nameFormatter = developPersister(nameFormatter)
+    if args.build:
+        nameFormatter = recipes.getHook('releaseNameFormatter')
+        nameFormatter = LocalBuilder.releaseNamePersister(nameFormatter)
+    else:
+        nameFormatter = recipes.getHook('developNameFormatter')
+        developPersister = recipes.getHook('developNamePersister')
+        nameFormatter = developPersister(nameFormatter)
+
     nameFormatter = LocalBuilder.makeRunnable(nameFormatter)
     packages = recipes.generatePackages(nameFormatter, defines, sandboxEnabled=args.sandbox)
-    touch(packages)
+    if not args.build:
+        touch(packages)
 
     from ..generators.QtCreatorGenerator import qtProjectGenerator
     from ..generators.EclipseCdtGenerator import eclipseCdtGenerator
@@ -1408,7 +1417,7 @@ def doProject(argv, bobRoot):
             print(g)
         return 0
     else:
-        if not args.package or not args.projectGenerator:
+        if not args.packages or not args.projectGenerator:
             raise BobError("The following arguments are required: projectGenerator, package")
 
     try:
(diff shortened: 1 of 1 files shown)


kkendzia:
Ok I see there is a problem with additionals 'args' :/

jkloetzke:
I'm a bit lost how to integrate that in a backwards compatible way. It will at least break the plugin API but that should be feasible via the `apiVersion` tag in the plugin manifest.

What I'm more concerned about is the argument passing. Right now any extra arguments are passed to the plugin. Having an unbounded number of package names effectively breaks this. I don't like the `--args` proposal because it breaks the current calling convention that people are used to... :thinking:

pym/bob/cmds/build.py
```diff
@@ -1429,20 +1438,35 @@ def doProject(argv, bobRoot):
...
-        devArgs.append(args.package)
-        doDevelop(devArgs, bobRoot)
+        for p in backlog:
+            devArgs.append(p.getName())
+        if args.build:
+            doBuild(devArgs, bobRoot)
+        else:
+            doDevelop(devArgs, bobRoot)
+
+    if len(backlog) == 1:
```
jkloetzke: This is not a good idea. Depending on the number of given packages the plugin will receive an `Package` or a list of `Package`s. Either pass an object (old plugin API < 0.13) or a list (new plugin API >= 0.13).

kkendzia: Yeah, as mentioned I also think this is a bad idea. Will always provide a list then.

kkendzia:
In fact it would break the current convention. That's why I though you have an idea. Positional arguments would be a possibility as already committed but it feels bad to handle IMHO

jkloetzke:
Maybe @rhubert has an idea?

rhubert:
The only idea that came into my mind is to add a special package handling to avoid the `--args` switch. Something like: 
```
bob project <generator> "package1;package2;//*/package3" <generatorArgs>
```
Then you could check if your package string matches the `$"{.*;}*"` format and split it. Otherwise use the given string as package directly.

kkendzia:
Mh this doesn't look good either IMHO. But I think I would go for it if we just introduce query language and call the generators with a package list.

kunerd:
My suggestion is to use `--` like in other projects, e.g. `npm`, and `bash` in general.
```
bob project <generator> package1 package2 <...package> -- <generatorArgs>
```
https://unix.stackexchange.com/questions/11376/what-does-double-dash-mean-also-known-as-bare-double-dash

kkendzia:
That would be nice.

With just query language I think it's not suitable for more than 2 packages.

rhubert:
I also like this, but it has to be compatible to the existing solution. Otherwise all existing scripts would break. If you can support both styles, with multiple packages and `--` as separator and the old style with one package and without the `--` it would be fine.

jkloetzke:
I think this syntax could work. It would probably impact existing projects, though, because the parameters are passed recursively through the buildme file.

Also the plugin API must be kept backwards compatible. If an old plugin is called with multiple packages then we should just print an error message...

kkendzia:
I changed the request like @kunerd suggested it. It should have a check if the generator uses the right API, isn't it?

pym/bob/cmds/build.py
```diff
@@ -1344,7 +1344,7 @@ def doDevelop(argv, bobRoot):
 def doProject(argv, bobRoot):
     parser = argparse.ArgumentParser(prog="bob project", description='Generate Project Files')
     parser.add_argument('projectGenerator', nargs='?', help="Generator to use.")
-    parser.add_argument('package', nargs='?', help="Sub-package that is the root of the project")
+    #parser.add_argument('packages', nargs='+', help="(Sub-)packages to generate")
```
jkloetzke: Commented out code.

pym/bob/cmds/build.py
```diff
@@ -1365,13 +1365,23 @@ def doProject(argv, bobRoot):
         help="Do not build (bob dev) before generate project Files. RunTargets may not work")
     parser.add_argument('-b', dest="execute_buildonly", default=False, action='store_true',
         help="Do build only (bob dev -b) before generate project Files. No checkout")
+    parser.add_argument('--build', default=False, action='store_true',
+        help="Do path calculations as if 'bob build', not as 'bob dev'")
```
jkloetzke: Isn't that part of PR #170 already? It would anyway be an unrelated change for this PR.

pym/bob/cmds/build.py
```diff
@@ -1365,13 +1365,23 @@ def doProject(argv, bobRoot):
...
     group = parser.add_mutually_exclusive_group()
     group.add_argument('--sandbox', action='store_true', default=False,
         help="Enable sandboxing")
     group.add_argument('--no-sandbox', action='store_false', dest='sandbox',
         help="Disable sandboxing")
     args = parser.parse_args(argv)
 
+    if not '--' in args.args:
+        inputPackages = args.args
+        generatorArguments = []
```
jkloetzke: Isn't that completely breaking existing workspaces? The generated "buildme" files will not have the double-dash and all arguments will be interpreted as packages.

pym/bob/cmds/build.py
```diff
@@ -1383,6 +1393,7 @@ def doProject(argv, bobRoot):
             parser.error("Malformed define: "+define)
 
     recipes = RecipeSet()
+    recipes.defineHook('releaseNameFormatter', LocalBuilder.releaseNameFormatter)
```
jkloetzke: Unrelated

pym/bob/cmds/build.py
```diff
@@ -1391,12 +1402,18 @@ def doProject(argv, bobRoot):
...
-    developPersister = recipes.getHook('developNamePersister')
-    nameFormatter = developPersister(nameFormatter)
+    if args.build:
+        nameFormatter = recipes.getHook('releaseNameFormatter')
+        nameFormatter = LocalBuilder.releaseNamePersister(nameFormatter)
+    else:
+        nameFormatter = recipes.getHook('developNameFormatter')
+        developPersister = recipes.getHook('developNamePersister')
+        nameFormatter = developPersister(nameFormatter)
+
```
jkloetzke: Unrelated

pym/bob/cmds/build.py
```diff
@@ -1391,12 +1402,18 @@ def doProject(argv, bobRoot):
...
     nameFormatter = LocalBuilder.makeRunnable(nameFormatter)
     packages = recipes.generatePackages(nameFormatter, defines, sandboxEnabled=args.sandbox)
-    touch(packages)
+    if not args.build:
+        touch(packages)
```
jkloetzke: Unrelated. Actually missing in #170 .

pym/bob/cmds/build.py
```diff
@@ -1429,20 +1446,30 @@ def doProject(argv, bobRoot):
...
+            devArgs.append(p.getName())
+        if args.build:
+            doBuild(devArgs, bobRoot)
+        else:
+            doDevelop(devArgs, bobRoot)
+
+    print(">>", colorize("Generate project for", "32;1"))
+    for package in backlog:
+        print(colorize("   PROJECT   {} ({})".format(package.getName(), args.projectGenerator), "32"))
+    generator(backlog, generatorArguments, extra)
```
jkloetzke: Here must be a backwards compatible Plugin-API change. Old plugins will only get one package. If the user selected more than one package and the plugin is not updated we must show an error message.

jkloetzke:
Closing due to inactivity and remaining issues. Feel free to re-open if you want to pick it up again. Anyway, thanks for the effort.

PR #166, opened 2017-10-25
build: add forced-fallback download mode

kkendzia:
With this mode it will be tried to download the desired package artifact at first. If this fails all dependency artifacts must be available

### DIFF
--- a/pym/bob/cmds/build.py
+++ b/pym/bob/cmds/build.py
@@ -294,7 +294,7 @@ esac
         self.__currentPackage = None
         self.__archive = DummyArchive()
         self.__downloadDepth = 0xffff
-        self.__forcedDownload = False
+        self.__downloadDepthForce = 0xffff
         self.__bobRoot = bobRoot
         self.__cleanBuild = cleanBuild
         self.__cleanCheckout = False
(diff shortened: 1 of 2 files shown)


pym/bob/cmds/build.py
```diff
@@ -947,7 +953,10 @@ def _cookPackageStep(self, packageStep, checkoutOnly, depth):
                         workspaceChanged = True
                         wasDownloaded = True
                     elif self.__forcedDownload:
-                        raise BuildError("Downloading artifact failed")
+                        if self.__forcedDownloadFallback:
+                            self.__forcedDownloadFallback = False
```
jkloetzke: This won't work if multiple packages are passed on the command line because after the first one `__forcedDownloadFallback` may have been reset.

kkendzia:
I updated the PR

pym/bob/cmds/build.py
```diff
@@ -946,8 +950,11 @@ def _cookPackageStep(self, packageStep, checkoutOnly, depth):
                         packageHash = hashWorkspace(packageStep)
                         workspaceChanged = True
                         wasDownloaded = True
-                    elif self.__forcedDownload:
-                        raise BuildError("Downloading artifact failed")
+                    elif self.__forcedDownload > 0:
+                        if self.__downloadDepth == 0 and self.__forcedDownload == 2:
+                            self.__downloadDepth = 1
```
jkloetzke: No, you must not modify the settings. The LocalBuilder object is used multiple times if multiple packages are passed on the command line. My Idea was to define a `__downloadDepthForce` (which is >= `__downloadDepth`) and you just do:

```python
elif depth >= self.__downloadDepthForce:
    raise BuildError("Downloading artifact failed")
```

kkendzia:
Ok I think I finally understood what you've said ;) PR updated

PR #167, opened 2017-10-25
bob: don't add spaces after last PATH element

zub2:
bob project qt-creator appends two spaces after the contents of the PATH
variable in QtCreator project file. This breaks the last component of the
PATH variable. E.g. it turns "foo:bar:baz" into "foo:bar:baz  " and
whatever would be found in "baz" will not be found when the PATH contains
"baz  " instead.

Remove the extra spaces.


jkloetzke:
Thanks :+1:

PR #168, opened 2017-10-25
Shared toolchain

thwerner:
Installs packages (toolchains) to a defined folder (global, different from workspace), packages installed to this folder can be re-used by any workspace.

    recipes with "shared: True" and provideTools are affected

    after building successfully the dist-directory, env and audit.tgz will be copied into a selected folder (specified with "-i" for install, and --shared for destination)

    with "--shared" toolchains will be linked into the workspace instead of built again

    advantage:
        less initial build time for new workspaces
        can be used with jenkins

Attention: will not be working with non-relocatable toolchains

### DIFF
--- a/doc/manpages/bob-build-dev.rst
+++ b/doc/manpages/bob-build-dev.rst
@@ -88,6 +88,12 @@ Options
 ``-f, --force``
     Force execution of all build steps
 
+``-i, --installshared``
+    Install shared packages to a given location (has to be combined with --shared)
+
+``-s, --shared PATH``
+    Shared packages will be searched at or installed to provided path
+
 ``-n, --no-deps``
     Don't build dependencies
 
--- a/pym/bob/cmds/build.py
+++ b/pym/bob/cmds/build.py
@@ -32,6 +32,7 @@ import re
 import shutil
 import stat
 import subprocess
+import tempfile
 import time
 
 # Output verbosity:
@@ -897,10 +898,12 @@ esac
             # determine a build-id for all other artifacts.
             #
             # Create build-id for shared packages, but don't download them
+            canLinkShared = False
+            canDownload = False
             if packageStep.doesProvideTools() and (packageStep.getSandbox() is None):
-                if packageStep.isShared() and self.__sharedFolder:
-                    canDownload = False
+                if packageStep.isShared() and self.__sharedFolder and packageStep.isPackageRelocatable():
                     packageBuildId = self._getBuildId(packageStep, depth)
+                    canLinkShared = True
                 else:
                     packageBuildId = None
             else:
@@ -924,22 +927,24 @@ esac
                 oldInputHashes = oldInputBuildId
                 oldWasDownloadedOrShared = False
 
-            # If possible try to download the package. If we downloaded the
+            # If possible try to link shared or download the package. If we linked or downloaded the
             # package in the last run we have to make sure that the Build-Id is
             # still the same. The overall behaviour should look like this:
             #
             # new workspace -> try download
             # previously built
             #   still same build-id -> normal build
-            #   build-id changed -> prune and try download, fall back to build
+            #   build-id changed -> prune and try link/download, fall back to build
             # previously downloaded
             #   still same build-id -> done
-            #   build-id changed -> prune and try download, fall back to build
+            #   build-id changed -> prune and try link/download, fall back to build
+            # previously linked
+            #   still same build-id -> done
+            #   build-id changed -> prune and try link/download, fall back to build
             workspaceChanged = False
-            wasDownloaded = False
-            wasShared = False
-            if ( (not checkoutOnly) and packageBuildId and (depth >= self.__downloadDepth) ) and canDownload:
-                # prune directory if we previously downloaded/built something different
+            wasDownloadedOrShared = False
+            if ( (not checkoutOnly) and packageBuildId and (depth >= self.__downloadDepth) ) and (canDownload or canLinkShared):
+                # prune directory if we previously downloaded/built/linked something different
                 if ((oldInputBuildId is not None) and (oldInputBuildId != packageBuildId)) or self.__force:
                     print(colorize("   PRUNE     {} ({})".format(prettyPackagePath,
                             "build forced" if self.__force else "build-id changed"), "33"))
@@ -953,50 +958,52 @@ esac
                 # empty. If the directory holds a result and was downloaded it
                 # we're done.
                 if BobState().getResultHash(prettyPackagePath) is None:
-                    audit = os.path.join(prettyPackagePath, "..", "audit.json.gz")
-                    if self.__archive.downloadPackage(packageBuildId, audit, prettyPackagePath, self.__verbose):
-                        self.__statistic.packagesDownloaded += 1
+                    if packageStep.isShared() and self.__sharedFolder and packageStep.isPackageRelocatable():
+                        if self._checkSharedPackage(asHexStr(packageBuildId)):
+                            sharedpackagepath = os.path.join(self.__sharedFolder, asHexStr(packageBuildId))
+                            print(colorize("     Link from shared location {}".format(sharedpackagepath), "32"))
+                            dirlist = os.listdir(sharedpackagepath)
+                            for item in dirlist:
+                                itempath = os.path.join(sharedpackagepath, item)
+                                if item in ['audit.json.gz']:
+                                    os.symlink(itempath, os.path.join(prettyPackagePath, '..', item))
+                                else:
+                                    os.symlink(itempath, os.path.join(prettyPackagePath, item), os.path.isdir(itempath))
+                            wasDownloadedOrShared = True
+
+                    if not wasDownloadedOrShared and canDownload:
+                        audit = os.path.join(prettyPackagePath, "..", "audit.json.gz")
+                        if self.__archive.downloadPackage(packageBuildId, audit, prettyPackagePath, self.__verbose):
+                            self.__statistic.packagesDownloaded += 1
+                            BobState().setInputHashes(prettyPackagePath, packageBuildId)
+                        elif self.__forcedDownload:
+                            raise BuildError("Downloading artifact failed")
+
+                    if wasDownloadedOrShared:
                         BobState().setInputHashes(prettyPackagePath, packageBuildId)
                         packageHash = hashWorkspace(packageStep)
                         workspaceChanged = True
+<<<<<<< c628942e7ed63ce90d680b8ec3038e6aec5b8aa9
                         wasDownloaded = True
                     elif depth >= self.__downloadDepthForce:
                         raise BuildError("Downloading artifact failed")
                 elif oldWasDownloadedOrShared:
                     self._info("   PACKAGE   skipped (already downloaded in {})".format(prettyPackagePath))
                     wasDownloaded = True
+=======
+                        wasDownloadedOrShared = True
+>>>>>>> make installation of shared packages atomic, sharedFolder environment variable in user config files
 
-            if packageStep.isShared() and self.__sharedFolder and not wasDownloaded:
-                if BobState().getResultHash(prettyPackagePath) is None:
-                    sharedpackagepath = os.path.join(self.__sharedFolder, asHexStr(packageBuildId))
-                    # link into dist if shared package available
-                    if self._checkSharedPackage(packageStep, asHexStr(packageBuildId)):
-                        print(colorize("     Link from shared location {}".format(sharedpackagepath), "32"))
-                        dirlist = os.listdir(sharedpackagepath)
-                        cwd = os.getcwd()
-                        os.chdir(prettyPackagePath)
-                        for item in dirlist:
-                            if item == '.installed':
-                                continue
-                            itempath = os.path.join(sharedpackagepath, item)
-                            if item in ['env', 'audit.json.gz']:
-                                os.symlink(itempath, os.path.join(os.getcwd(), '..', item))
-                            else:
-                                os.symlink(itempath, item, os.path.isdir(itempath))
-                        os.chdir(cwd)
-                        wasShared = True
-                        workspaceChanged = True
-                        packageHash = hashWorkspace(packageStep)
                 elif oldWasDownloadedOrShared:
-                    self._info("   PACKAGE   skipped (deterministic output in {})".format(prettyPackagePath))
-                    wasShared = True
+                    self._info("   PACKAGE   skipped (already downloaded or linked in {})".format(prettyPackagePath))
+                    wasDownloadedOrShared = True
 
 
             # Run package step if we have not yet downloaded the package or if
             # downloads are not possible anymore. Even if the package was
             # previously downloaded the oldInputHashes will be None to trigger
             # an actual build.
-            if not wasDownloaded and not wasShared:
+            if not wasDownloadedOrShared:
                 # depth first
                 self._cook(packageStep.getAllDepSteps(), packageStep.getPackage(),
                            checkoutOnly, depth+1)
@@ -1017,17 +1024,24 @@ esac
                     packageHash = hashWorkspace(packageStep)
                     audit = self._generateAudit(packageStep, depth, packageHash)
                     # copy to shared location if possible
-                    if self.__installShared and packageStep.isShared():
-                        sharedpackagepath = os.path.join(self.__sharedFolder, asHexStr(packageBuildId))
+                    if self.__installShared and packageStep.isShared() and packageStep.isPackageRelocatable() and self.__sharedFolder:
+                        buildidhexstr =  asHexStr(packageBuildId)
+                        sharedpackagepath = os.path.join(self.__sharedFolder, buildidhexstr)
                         print(colorize("     Install in shared location {}".format(sharedpackagepath), "32"))
-                        if os.path.exists(sharedpackagepath):
-                            shutil.rmtree(sharedpackagepath)
+                        # create temp dir
+                        tmpdir = tempfile.mkdtemp()
+                        tmppackagepath = os.path.join(tmpdir, buildidhexstr)
                         # copy all files from workspace - keep symlinks
-                        shutil.copytree(prettyPackagePath, sharedpackagepath, symlinks=True)
-                        # copy env file - important for audit
-                        shutil.copy2(os.path.join(prettyPackagePath, '..', 'env'), sharedpackagepath)
-                        shutil.copy2(os.path.join(prettyPackagePath, '..', 'audit.json.gz'), sharedpackagepath)
-                        open(os.path.join(sharedpackagepath, '.installed'), 'a').close()
+                        shutil.copytree(prettyPackagePath, tmppackagepath, symlinks=True)
+                        # copy audit
+                        shutil.copy2(os.path.join(prettyPackagePath, '..', 'audit.json.gz'), tmppackagepath)
+                        if self._checkSharedPackage(buildidhexstr):
+                            shutil.rmtree(sharedpackagepath)
+                        # move - atomic operation
+                        shutil.move(tmppackagepath, sharedpackagepath)
+                        # cleanup tmpdir
+                        shutil.rmtree(tmpdir)
+
                     workspaceChanged = True
                     self.__statistic.packagesBuilt += 1
                     if packageBuildId and self.__archive.canUploadLocal():
@@ -1038,7 +1052,7 @@ esac
             # Rehash directory if content was changed
             if workspaceChanged:
                 BobState().setResultHash(prettyPackagePath, packageHash)
-                if wasDownloaded or wasShared:
+                if wasDownloadedOrShared:
                     BobState().setInputHashes(prettyPackagePath, packageBuildId)
                 else:
                     BobState().setInputHashes(prettyPackagePath, [packageBuildId] + packageInputHashes)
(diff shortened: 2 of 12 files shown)


pym/bob/cmds/build.py
```diff
@@ -976,15 +1012,29 @@ def _cookPackageStep(self, packageStep, checkoutOnly, depth):
                     self._runShell(packageStep, "package")
                     packageHash = hashWorkspace(packageStep)
                     audit = self._generateAudit(packageStep, depth, packageHash)
+                    # copy to shared location if possible
+                    if self.__installShared and packageStep.isShared():
+                        sharedpackagepath = os.path.join(self.__sharedFolder, asHexStr(packageBuildId))
+                        print(colorize("     Install in shared location {}".format(sharedpackagepath), "32"))
+                        if os.path.exists(sharedpackagepath):
+                            shutil.rmtree(sharedpackagepath)
```
jkloetzke: The installation is racy if two instances run in parallel. It should be done with a temporary directory and an atomic rename of the whole folder like in `jenkins.py`.

pym/bob/cmds/build.py
```diff
@@ -889,9 +891,16 @@ def _cookPackageStep(self, packageStep, checkoutOnly, depth):
...
             # determine a build-id for all other artifacts.
+            #
+            # Create build-id for shared packages, but don't download them
             if packageStep.doesProvideTools() and (packageStep.getSandbox() is None):
-                packageBuildId = None
+                if packageStep.isShared() and self.__sharedFolder:
+                    canDownload = False
+                    packageBuildId = self._getBuildId(packageStep, depth)
+                else:
+                    packageBuildId = None
```
jkloetzke: `canDownload = False`

pym/bob/cmds/build.py
```diff
@@ -948,15 +958,41 @@ def _cookPackageStep(self, packageStep, checkoutOnly, depth):
                         wasDownloaded = True
                     elif self.__forcedDownload:
                         raise BuildError("Downloading artifact failed")
-                elif oldWasDownloaded:
+                elif oldWasDownloadedOrShared:
                     self._info("   PACKAGE   skipped (already downloaded in {})".format(prettyPackagePath))
                     wasDownloaded = True
 
+            if packageStep.isShared() and self.__sharedFolder and not wasDownloaded:
```
jkloetzke: Isn't that already too late here? We would first try to download the artifact and then see if it's in the shared location...

pym/bob/cmds/build.py
```diff
@@ -948,15 +958,41 @@ def _cookPackageStep(self, packageStep, checkoutOnly, depth):
...
+                if BobState().getResultHash(prettyPackagePath) is None:
+                    sharedpackagepath = os.path.join(self.__sharedFolder, asHexStr(packageBuildId))
+                    # link into dist if shared package available
+                    if self._checkSharedPackage(packageStep, asHexStr(packageBuildId)):
+                        print(colorize("     Link from shared location {}".format(sharedpackagepath), "32"))
+                        dirlist = os.listdir(sharedpackagepath)
+                        cwd = os.getcwd()
+                        os.chdir(prettyPackagePath)
+                        for item in dirlist:
```
jkloetzke: No need to change the working directory (just use `os.path.join()`). Also there are only two things that must be linked: `workspace` and `audit.json.gz`.

pym/bob/cmds/build.py
```diff
@@ -1082,6 +1132,13 @@ def __getCheckoutStepBuildId(self, step, depth):
 
         return ret, predicted
 
+    def _checkSharedPackage(self, step, packageBuildId):
+         sharedPackagePath = os.path.join(self.__sharedFolder, packageBuildId)
+         if os.path.isdir(sharedPackagePath):
+             if os.path.exists(os.path.join(sharedPackagePath, '.installed')):
```
jkloetzke: The ".installed" canary is not needed if the folder is atomically "installed" (renamed) to the right place.

doc/manpages/bob-build-dev.rst
```diff
@@ -86,6 +86,12 @@ Options
 ``-f, --force``
     Force execution of all build steps
 
+``-i, --installshared``
+    Install shared packages to a given location (has to be combined with --shared)
```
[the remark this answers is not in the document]
kkendzia: I think it would be OK to install them by default, but I want to control if it should be installed or not. Would you go with a "dont install relocatable packages" parameter?

kkendzia:
I've realized most of your requested suggestions. Should I implement a parameter preventing any linking? For example if you want to have totally clean build and you want any package to be build.

doc/manpages/bob-build-dev.rst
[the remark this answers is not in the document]
jkloetzke: Wouldn't you just *not* define the shared location? Or set it to an empty string in your `user.yaml` if you defined it globally? I don't think this is something that should be a command line option because you don't typically change your mind between invocations.

Do I miss something?

pym/bob/cmds/build.py
```diff
@@ -939,24 +954,43 @@ def _cookPackageStep(self, packageStep, checkoutOnly, depth):
...
+                                    os.symlink(itempath, os.path.join(prettyPackagePath, '..', item))
+                                else:
+                                    os.symlink(itempath, os.path.join(prettyPackagePath, item), os.path.isdir(itempath))
+                            wasDownloadedOrShared = True
+
+                    if not wasDownloadedOrShared and canDownload:
+                        audit = os.path.join(prettyPackagePath, "..", "audit.json.gz")
+                        if self.__archive.downloadPackage(packageBuildId, audit, prettyPackagePath, self.__verbose):
+                            self.__statistic.packagesDownloaded += 1
+                            BobState().setInputHashes(prettyPackagePath, packageBuildId)
```
jkloetzke: Don't you have to set ` wasDownloadedOrShared = True` here?

pym/bob/cmds/build.py
```diff
@@ -939,24 +954,43 @@ def _cookPackageStep(self, packageStep, checkoutOnly, depth):
                 # empty. If the directory holds a result and was downloaded it
                 # we're done.
                 if BobState().getResultHash(prettyPackagePath) is None:
-                    audit = os.path.join(prettyPackagePath, "..", "audit.json.gz")
-                    if self.__archive.downloadPackage(packageBuildId, audit, prettyPackagePath, self.__verbose):
-                        self.__statistic.packagesDownloaded += 1
+                    if packageStep.isShared() and self.__sharedFolder and packageStep.isPackageRelocatable():
+                        if self._checkSharedPackage(asHexStr(packageBuildId)):
+                            sharedpackagepath = os.path.join(self.__sharedFolder, asHexStr(packageBuildId))
+                            print(colorize("     Link from shared location {}".format(sharedpackagepath), "32"))
```
jkloetzke: I would factor out the "link shared" boiler plate code into a dedicated function. The `_cookPackageStep` method is already barely readable...

pym/bob/cmds/build.py
```diff
@@ -939,24 +954,43 @@ def _cookPackageStep(self, packageStep, checkoutOnly, depth):
...
                         BobState().setInputHashes(prettyPackagePath, packageBuildId)
                         packageHash = hashWorkspace(packageStep)
                         workspaceChanged = True
-                        wasDownloaded = True
-                    elif self.__forcedDownload:
-                        raise BuildError("Downloading artifact failed")
-                elif oldWasDownloaded:
-                    self._info("   PACKAGE   skipped (already downloaded in {})".format(prettyPackagePath))
-                    wasDownloaded = True
+                        wasDownloadedOrShared = True
```
jkloetzke: If I read it correctly this is:

```python
if wasDownloadedOrShared:
    ...
    wasDownloadedOrShared = True
```

Doesn't look right...

pym/bob/cmds/build.py
```diff
@@ -976,15 +1010,36 @@ def _cookPackageStep(self, packageStep, checkoutOnly, depth):
                     self._runShell(packageStep, "package")
                     packageHash = hashWorkspace(packageStep)
                     audit = self._generateAudit(packageStep, depth, packageHash)
+                    # copy to shared location if possible
+                    if self.__installShared and packageStep.isShared() and packageStep.isPackageRelocatable() and self.__sharedFolder:
+                        buildidhexstr =  asHexStr(packageBuildId)
+                        sharedpackagepath = os.path.join(self.__sharedFolder, buildidhexstr)
+                        print(colorize("     Install in shared location {}".format(sharedpackagepath), "32"))
+                        # create temp dir
+                        tmpdir = tempfile.mkdtemp()
```
jkloetzke: You have to create the temporary directory in the shared folder. Otherwise the rename is not possible if /tmp is on another file system. At least the rename cannot be atomic anymore.

pym/bob/cmds/build.py
```diff
@@ -976,15 +1010,36 @@ def _cookPackageStep(self, packageStep, checkoutOnly, depth):
...
+                        tmppackagepath = os.path.join(tmpdir, buildidhexstr)
+                        # copy all files from workspace - keep symlinks
+                        shutil.copytree(prettyPackagePath, tmppackagepath, symlinks=True)
+                        # copy audit
+                        shutil.copy2(os.path.join(prettyPackagePath, '..', 'audit.json.gz'), tmppackagepath)
+                        if self._checkSharedPackage(buildidhexstr):
+                            shutil.rmtree(sharedpackagepath)
+                        # move - atomic operation
+                        shutil.move(tmppackagepath, sharedpackagepath)
```
jkloetzke: This is *not* atomic if you cross file systems and `shutil.move` will hide it from you. Directly use `os.rename` instead...

pym/bob/cmds/build.py
```diff
@@ -976,15 +1010,36 @@ def _cookPackageStep(self, packageStep, checkoutOnly, depth):
...
+                        # cleanup tmpdir
+                        shutil.rmtree(tmpdir)
+
                     workspaceChanged = True
                     self.__statistic.packagesBuilt += 1
                     if packageBuildId and self.__archive.canUploadLocal():
-                        self.__archive.uploadPackage(packageBuildId, audit, prettyPackagePath, self.__verbose)
+                        if not (packageStep.doesProvideTools() and (packageStep.getSandbox() is None)):
```
jkloetzke: `isPackageRelocatable()` instead of `doesProvideTools()`?

pym/bob/cmds/build.py
```diff
@@ -976,15 +1010,36 @@ def _cookPackageStep(self, packageStep, checkoutOnly, depth):
                     self._runShell(packageStep, "package")
                     packageHash = hashWorkspace(packageStep)
                     audit = self._generateAudit(packageStep, depth, packageHash)
+                    # copy to shared location if possible
+                    if self.__installShared and packageStep.isShared() and packageStep.isPackageRelocatable() and self.__sharedFolder:
+                        buildidhexstr =  asHexStr(packageBuildId)
```
jkloetzke: The "install shared" logic should be factored out like the "link shared" one...

pym/bob/input.py
```diff
@@ -428,9 +428,9 @@ def toStep(self, pathFormatter, upperPackage):
 class CoreStep(metaclass=ABCMeta):
     __slots__ = ( "package", "label", "tools", "digestEnv", "env", "args",
         "shared", "doesProvideTools", "providedEnv", "providedTools",
-        "providedDeps", "providedSandbox", "variantId", "sandbox", "sbxVarId" )
+        "providedDeps", "providedSandbox", "variantId", "sandbox", "sbxVarId", "packageRelocatable" )
```
jkloetzke: Maybe just name it "relocatable". We create millions of these objects and every byte counts. :wink:

pym/bob/cmds/build.py
```diff
@@ -889,9 +892,18 @@ def _cookPackageStep(self, packageStep, checkoutOnly, depth):
             # Can we theoretically download the result? Exclude packages that
             # provide host tools when not building in a sandbox. Try to
             # determine a build-id for all other artifacts.
+            #
+            # Create build-id for shared packages, but don't download them
+            canLinkShared = False
+            canDownload = False
             if packageStep.doesProvideTools() and (packageStep.getSandbox() is None):
```
jkloetzke: I think the logic here (and on more locations below) should be

```python
if packageStep.isPackageRelocatable() or (packageStep.getSandbox() is not None):
```

The rationale is the following:
* inside the sandbox all paths are stable -> we never relocate
* outside the sandbox the package must be relocatable to work correctly
* non-relocatable packages are never up- or downloaded -> should never be a problem

doc/manpages/bob-build-dev.rst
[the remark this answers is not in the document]
kkendzia: I meant to use them but not to install. Set the variable to an empty string would cause not to use it either. --shared isn't necessary but something like "--dontinstallsharedpackages" ;) would be nice

pym/bob/input.py
kunerd: I think pythons compiler and interpreter is smart enough to optimize this.

doc/manpages/bob-build-dev.rst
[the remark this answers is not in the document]
jkloetzke: I think the problem is that the shared packages are re-installed every time. This is not necessary and we should change the "install" part to only do it if the package is not yet available in the shared location. Then you only have the overhead of copying it once.

kkendzia:
I've pushed another set of changes. Thanks for reviewing :-)

pym/bob/cmds/build.py
```diff
@@ -980,20 +1008,62 @@ def _cookPackageStep(self, packageStep, checkoutOnly, depth):
                     self._runShell(packageStep, "package")
                     packageHash = hashWorkspace(packageStep)
                     audit = self._generateAudit(packageStep, depth, packageHash)
+                    # copy to shared location if possible
+                    self.__installSharedPackage(packageBuildId, packageStep, prettyPackagePath)
                     workspaceChanged = True
                     self.__statistic.packagesBuilt += 1
                     if packageBuildId and self.__archive.canUploadLocal():
-                        self.__archive.uploadPackage(packageBuildId, audit, prettyPackagePath, self.__verbose)
+                        if packageStep.isRelocatable():
```
jkloetzke: That should be replaced by `if canUpDownload and self.__archive.canUploadLocal():` instead of the nested `if`.

pym/bob/cmds/build.py
```diff
@@ -980,20 +1008,62 @@ def _cookPackageStep(self, packageStep, checkoutOnly, depth):
...
                 BobState().setResultHash(prettyPackagePath, packageHash)
-                if wasDownloaded:
+                if wasDownloadedOrShared:
                     BobState().setInputHashes(prettyPackagePath, packageBuildId)
                 else:
                     BobState().setInputHashes(prettyPackagePath, [packageBuildId] + packageInputHashes)
             self._setAlreadyRun(packageStep, False, checkoutOnly)
 
+    def __linkSharedPackage(self, packageBuildId, step, workspace):
+        if step.isShared() and step.isRelocatable() and self.__sharedFolder and self._checkSharedPackage(asHexStr(packageBuildId)):
```
jkloetzke: Don't replicate the conditions here. The only decision done here should be `self._checkSharedPackage(asHexStr(packageBuildId))`. The rest was already checked before calling `__linkSharedPackage`.

pym/bob/cmds/build.py
```diff
@@ -980,20 +1008,62 @@ def _cookPackageStep(self, packageStep, checkoutOnly, depth):
...
+            sharedpackagepath = os.path.join(self.__sharedFolder, asHexStr(packageBuildId))
+            print(colorize("     Link from shared location {}".format(sharedpackagepath), "32"))
```
jkloetzke: Please keep the style of the other message, e.g.

```python
"   LINK      {} ({})".format(workspace, sharedpackagepath)
```

pym/bob/cmds/build.py
```diff
@@ -980,20 +1008,62 @@ def _cookPackageStep(self, packageStep, checkoutOnly, depth):
...
+            # create temp dir
+            try:
+                tmpdir = tempfile.mkdtemp(dir=self.__sharedFolder)
+            except OSError:
+                raise OSError("Temporary directory couldn't be created.")
+            tmppackagepath = os.path.join(tmpdir, buildidhexstr)
+            # copy all files from workspace - keep symlinks
+            shutil.copytree(workspace, tmppackagepath, symlinks=True)
+            # copy audit
+            shutil.copy2(os.path.join(workspace, '..', 'audit.json.gz'), tmppackagepath)
```
jkloetzke: The audit trail is stored together with the results? What would happen if the package result contains a "audit.json.gz"? Better store the "workspace" and "audit.json.gz" separately.

pym/bob/cmds/build.py
```diff
@@ -980,20 +1008,62 @@ def _cookPackageStep(self, packageStep, checkoutOnly, depth):
...
+                else:
+                    os.symlink(itempath, os.path.join(workspace, item), os.path.isdir(itempath))
+            return True
+        return False
+
+    def __installSharedPackage(self, packageBuildId, step, workspace):
+        if step.isShared() and step.isRelocatable() and self.__sharedFolder and not self._checkSharedPackage(asHexStr(packageBuildId)):
+            buildidhexstr = asHexStr(packageBuildId)
+            sharedpackagepath = os.path.join(self.__sharedFolder, buildidhexstr)
+            print(colorize("     Install in shared location {}".format(sharedpackagepath), "32"))
```
jkloetzke: Message style too.

pym/bob/cmds/build.py
```diff
@@ -980,20 +1008,62 @@ def _cookPackageStep(self, packageStep, checkoutOnly, depth):
...
+                raise OSError("Temporary directory couldn't be created.")
+            tmppackagepath = os.path.join(tmpdir, buildidhexstr)
+            # copy all files from workspace - keep symlinks
+            shutil.copytree(workspace, tmppackagepath, symlinks=True)
+            # copy audit
+            shutil.copy2(os.path.join(workspace, '..', 'audit.json.gz'), tmppackagepath)
+            # second check
+            if not self._checkSharedPackage(buildidhexstr):
+                # move - atomic operation
+                shutil.move(tmppackagepath, sharedpackagepath)
```
jkloetzke: Better use `os.rename()` to catch errors where `shutil.move` would silently fall back to copy, thus breaking atomicy. Additionally you must handle an OSError because you may loose the race of the rename. The check above does not rule out that two instances of Bob do the rename at the same time. POSIX will make sure that only one of them wins.

pym/bob/cmds/build.py
```diff
@@ -980,20 +1008,62 @@ def _cookPackageStep(self, packageStep, checkoutOnly, depth):
...
+    def __installSharedPackage(self, packageBuildId, step, workspace):
+        if step.isShared() and step.isRelocatable() and self.__sharedFolder and not self._checkSharedPackage(asHexStr(packageBuildId)):
+            buildidhexstr = asHexStr(packageBuildId)
+            sharedpackagepath = os.path.join(self.__sharedFolder, buildidhexstr)
+            print(colorize("     Install in shared location {}".format(sharedpackagepath), "32"))
+            # create temp dir
+            try:
+                tmpdir = tempfile.mkdtemp(dir=self.__sharedFolder)
+            except OSError:
+                raise OSError("Temporary directory couldn't be created.")
```
jkloetzke: Raise a `BuildError`. Re-throwing `OSError` will lead to a backtrace.

pym/bob/cmds/build.py
```diff
@@ -980,20 +1008,62 @@ def _cookPackageStep(self, packageStep, checkoutOnly, depth):
...
+            # copy audit
+            shutil.copy2(os.path.join(workspace, '..', 'audit.json.gz'), tmppackagepath)
+            # second check
+            if not self._checkSharedPackage(buildidhexstr):
+                # move - atomic operation
+                shutil.move(tmppackagepath, sharedpackagepath)
+            else:
+                print(colorize("     {} already created by a different process.".format(sharedpackagepath),"33"))
+            # cleanup tmpdir
+            shutil.rmtree(tmpdir)
```
jkloetzke: Error handling. This basically applies to the whole function. Catch all possible OSErrors and make sure that the temporary directory is removed even in case of an error.

test/relocatable/plugins/checkPackages.py
```diff
@@ -0,0 +1,22 @@
+from bob.errors import ParseError
+
+# check if packages are relocatable or not
+
+def _checkPackages(step):
+    if step.isValid() and step.isPackageStep():
+        if not step.isRelocatable() and step.getPackage().getName() == 'some-tool-relocatable':
+            raise ParseError('Package some-tool-relocatable should be relocatable!')
+        if step.isRelocatable() and step.getPackage().getName() == 'some-tool':
```
jkloetzke: I think these kinds of tests should be unit-tests instead of black-box tests. Some black box tests still make sense to test the link/install logic.

pym/bob/cmds/build.py
```diff
@@ -980,20 +1008,62 @@ def _cookPackageStep(self, packageStep, checkoutOnly, depth):
...
+        if step.isShared() and step.isRelocatable() and self.__sharedFolder and self._checkSharedPackage(asHexStr(packageBuildId)):
+            sharedpackagepath = os.path.join(self.__sharedFolder, asHexStr(packageBuildId))
+            print(colorize("     Link from shared location {}".format(sharedpackagepath), "32"))
+            dirlist = os.listdir(sharedpackagepath)
+            for item in dirlist:
+                itempath = os.path.join(sharedpackagepath, item)
+                if item in ['audit.json.gz']:
+                    os.symlink(itempath, os.path.join(workspace, '..', item))
+                else:
+                    os.symlink(itempath, os.path.join(workspace, item), os.path.isdir(itempath))
```
jkloetzke: So the files in the root of the workspace will all be symlinks then? This is different from the original content!

jkloetzke:
Is this PR still considered? I've implemented the `relocatable` property already because of some other cleanup that I'm considering in the long run: https://github.com/BobBuildTool/bob/projects/3#card-7109023

kkendzia:
I think we can close it down.

jkloetzke:
Any change that this MR will be picked up again? I think the feature is still worthwhile to be implemented...

jkloetzke:
Superseded by #405

PR #170, opened 2017-11-13
project: add --release and --develop (default) mode to switch between name formatter

kkendzia:
For some plug-ins it would be useful if you can fetch the workspace for "bob build"-built projects.

### DIFF
--- a/doc/manpages/bob-project.rst
+++ b/doc/manpages/bob-project.rst
@@ -13,7 +13,7 @@ Synopsis
 
 ::
 
-    bob project [-h] [--list] [-D DEFINES] [-c CONFIGFILE] [-e NAME] [-E]
+    bob project [-h] [--develop | --release] [--list] [-D DEFINES] [-c CONFIGFILE] [-e NAME] [-E]
                 [--resume] [-n] [-b]
                 [projectGenerator] [package] ...
 
@@ -32,6 +32,9 @@ Options
 ``-D DEFINES``
     Override default environment variable
 
+``--develop``
+    Use developer mode. This is default.
+
 ``-e NAME``
     Preserve environment variable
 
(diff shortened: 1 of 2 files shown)


jkloetzke:
Have you tested this with a QtCreator and/or Eclipse project? I don't think this will work correctly because the plugins will still do a "bob dev" when acutally building the code...

kkendzia:
Yeah you are right - but implementing the possibility to let the plugin/generator know, if it's build with --release or --develop, would break the current API right? But adding a new parameter to plugins would be my only solution.

jkloetzke:
Yes, so Bob has to support the old API and extend it for this feature based on the `apiVersion` of the plugin. Using an old plugin with `--build` would have to throw an error because the plugin is not prepared for this. The buillt-in generator plugins have to be updated and must make sure that existing workspaces are still working (that shouldn't be a problem because `--develop` takes precedence).

kunerd:
Just out of curiosity, are breaking changes really an issue here? I'm asking because, according to semver, bob is still below first stable release `1.*` and people should not consider the API stable at this point.

jkloetzke:
While you're technically right I don't want to break things for people without very good reasons. For example we have a generator plugin for SmartGit in our recipes. If this change would be a hard break then you could not easily update Bob without adapting the plugin (and revert that if you go back). 

People don't look at version numbers. If you break stuff you always have to give good reasons or you will loose trust. In hindsight I'm not even convinced by adopting SemVer because the line between major/minor/patch updates are blurrier than I thought initially. Even a patch update may be considered a breaking change if your recipes relied on the faulty behavior.

In this particular PR I don't yet see a use case for regular users. Using release build mode with an IDE just makes no sense IMHO. That's why the develop mode was hard coded in the first place. Additionally I think maintaining compatibility is not that hard here. I guess it can be done with a couple of lines...

PR #172, opened 2017-11-16
Added coloring for Jenkins CI console output

mprzeplata:


jkloetzke:
I don't think this is the right approach. Having magic environment variable values feels wrong. I would rather much prefer a git-like solution. There should be a switch in default.yaml that could read like this:

```
ui:
   color: "always"/"auto"/"never"
```

Additionally a `--color=` global switch can be added to override the setting on a case-by-case basis:

```
bob --color=always build ...
```

pym/bob/input.py
```diff
@@ -2329,6 +2332,13 @@ def __parseUserConfig(self, fileName, relativeIncludes=None):
         if not self._ignoreCmdConfig:
             self.__commandConfig = updateDicRecursive(self.__commandConfig, cfg.get("command", {}))
         self.__buildHooks.update(cfg.get("hooks", {}))
+        # change color mode
+        ui = cfg.get('ui', {})
+        if ui:
+            mode = ui.get('color', None)
+            if mode:
+                # note: this is applied after the cmd line options are parsed
+                setColorMode(mode)
```
jkloetzke: You can use `updateDicRecursive()` to merge the dicts (see above). Also, the order is not as intended as the command line option should always overrule whatever is in the config files.

pym/bob/scripts.py
```diff
@@ -178,6 +181,9 @@ def cmd():
             from .input import RecipeSet
             RecipeSet.ignoreCommandCfg()
 
+        if args.color_mode:
+            setColorMode(args.color_mode)
```
jkloetzke: You could do it like above to set the command line preference on the `RecipeSet` class and do a single `setColorMode()` in `__parse` after the user configuration was read but before the recipes are parsed.

pym/bob/tty.py
```diff
@@ -15,6 +15,7 @@
 # along with this program.  If not, see <http://www.gnu.org/licenses/>.
 
 import sys
+import os
```
jkloetzke: Stray `import`.

mprzeplata:
Note: even with 'always' set on cmd line or config file, first 'INFO' line is printed without it (probably using default 'auto' set at 'import tty'), only lines printed few seconds later begin to be colored.

bob_color_issue

jkloetzke:
Looks good. Please squash it to two commits (code changes + doc updates) with commit messages in the style of the existing commits.

That the first policy info message is not colored correctly is expected. This is printed before the settings are parsed. I don't think this is something that needs to be addressed.

Issue #177, opened 2017-11-24
Newer bob versions are slow with many root recipes


kolewu:
We still used bob 0.8.0 because the newer versions are very slow. We have 134 root recipes and `bob ls` lets us wait 7-8 seconds everytime something has changed. Since the helpers like bash completion don't use a configuration, there often are changes that lead to rehashing and that takes time.

When disabling most of the root recipes, the hashing time drops down to under a second. Is there any other way to cut the hashing times down than reducing the root recipes? In my desperation I plan a script that deactivates all root recipes except for one or two ...

jkloetzke:
How much faster is 0.8 in your setup? Can you share the "time bob ls" for both versions? (You have to delete `.bob-packages*`starting with 0.9 to force a re-calculation of the packages. Deleting `.bob-cache*` will add the yaml parsing overhead but that should not have changed.)

I guess the first problem might be that the bash completion does not pass configuration files to "bob ls". Fixing this might prevent delays between bash completion and Bob invocations in your case.

The current behavior is more optimized towards projects with few root recipes but deep dependency hierarchies where the same packages are used often. I guess that is not the case for you. There is another cache for the data that "bob ls" is using. This only pays off though if Bob doesn't have to re-parse the recipes every time.

May I'll add a `--debug=prof` switch to run the command trough "cProfile" to get some performance numbers. But I do not expect to see something easily fixable.

jkloetzke:
I've added a profile switch. Maybe running the following reveals something:

```sh
bob ls
rm -f .bob-packages* .bob-tree*
bob --debug=prof ls
```

jkloetzke:
Please have a look at #178 . Maybe it improves your situation already a bit. The bash completion was really broken with respect to config files or manually passed variables (`-D FOO=bar`).

kolewu:
Profiling: Why do you only remove `.bob-packages*` and `.bob-tree*`? What about `.bob-cache.shelve`?

jkloetzke:
The `.bob-cache.shelve` holds the parsed yaml files (yaml -> python data structures). This overhead did not change between versions and shouldn't make a difference here. This would only pollute the profiling data.

kolewu:
I have worked on profiling the problem. We have another repo with 250 root recipes that needs whopping 86 seconds to build a simple recipe, that only echoes the build states.

[...]

PR #178, opened 2017-11-25
bash-completion: fix parameter passing

jkloetzke:
The bash completion should now pick up any "-C", "-c" and "-D" options
correctly. This will prevent any re-parsing after invoking a bash
completion. This should improve #177 by reducing the number of
re-parsings.

File names with white spaces are still not completed correctly, though.
This needs an additional round of fixing.

### DIFF
--- a/contrib/bash-completion
+++ b/contrib/bash-completion
@@ -32,13 +32,40 @@ __bob_complete_words()
    done
 }
 
+# Complete a directory while obeying "-C" parameters.
+__bob_complete_dir()
+{
+   local IFS=$'\n'
+   COMPREPLY=( $(for i in "${chroot[@]}" ; do eval cd "$i" || exit ; done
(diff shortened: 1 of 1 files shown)


contrib/bash-completion
```diff
@@ -32,56 +32,110 @@ __bob_complete_words()
...
+
 __bob_commands="build dev clean graph help jenkins ls project status  query-scm query-recipe query-path query-meta"
 
+# Complete a Bob path
+#
+# We effectively call "bob ls" to get the list of possible completions. What
+# makes this a bit complicated is that we have to use the same settings as the
+# original command, namely the sandbox, config files and defines.
+#
+# To make thins even more complicated we have to handle "-C" where we change
```
kolewu: *thins* :arrow_right:  *things* or *this*

PR #181, opened 2017-11-29
Convert shell startscripts to python

kolewu:
With this change it's now possible to use debug and other wrapping applications like the profiler:

```
python3 -m cProfile -o bob-ls.stats $BOB_PATH/bob ls
gprof2dot -f pstats bob-ls.stats | dot -Tpng -o bob-ls.png
```

Since we don't use the other start scripts, I have only testet `bob`.

While we are at it? Why are the other scripts not part of the normal bob runtime and implemented as subcommands?

### DIFF
--- a/bob
+++ b/bob
@@ -1,6 +1,12 @@
-#!/bin/sh
-D="$(dirname "$(readlink -f "${0}")")"
-export PYTHONPATH="$D/pym:${PYTHONPATH}"
-exec python3 -c "import sys
+#!/usr/bin/env python3
+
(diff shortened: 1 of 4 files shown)


jkloetzke:
Awesome. I didn't know this is possible. Also `gprof2dot` looks more than interesting.

Regarding the separate scripts: they are only used by Bob internally when running on Jenkins. These are helper tools that are not intended to be called by the user directly. Therefore they are not accessible by the main "bob" command.

PR #196, opened 2018-02-12
Checkout assert

rhubert:
+ checkoutAssert as discussed in #194

### DIFF
--- a/doc/manual/configuration.rst
+++ b/doc/manual/configuration.rst
@@ -468,6 +468,28 @@ If a step depends on a certain variable then the result of the following step
 is already indirectly dependent on this variable. Thus it can be set during the
 following step anyway.
 
+
+checkoutAssert
+~~~~~~~~~~~~~~
+
+Type: List of checkoutAssertions
+
+Using checkoutAsserts you can make a build fail if a file content has
+been changed. This is especially useful to detect modifications in
+License files.
+
+The following properties are known:
+
+================= ==================================================================
+Property           Description
+================= ==================================================================
+file              | The file in the workspace to check.
+digestSHA1        | Digest of the file / part. Either pre calculate it using
+                  | `sha1sum` command or take the output of the first (failing) run.
+start             | Optionally. Defaults to 1.
+end               | Optionally. Defaults to last line of file.
+================= ==================================================================
+
 checkoutDeterministic
 ~~~~~~~~~~~~~~~~~~~~~
 
--- a/pym/bob/input.py
+++ b/pym/bob/input.py
@@ -18,6 +18,7 @@ from glob import glob
 from pipes import quote
 from os.path import expanduser
 from string import Template
+from textwrap import dedent
 import copy
 import hashlib
 import fnmatch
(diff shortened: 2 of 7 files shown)


pym/bob/cmds/build.py
```diff
@@ -733,6 +733,7 @@ def _cookCheckoutStep(self, checkoutStep, depth):
             checkoutExecuted = False
             checkoutState = checkoutStep.getScmDirectories().copy()
             checkoutState[None] = checkoutDigest
+            checkoutState['checkoutAsserts'] = checkoutStep.getCheckoutAsserts().copy()
```
jkloetzke: That looks odd. There should be no need to treat checkout asserts as pseudo SCM.

pym/bob/input.py
```diff
@@ -6,7 +6,7 @@
 from . import BOB_VERSION, BOB_INPUT_HASH, DEBUG
 from .errors import ParseError
 from .pathspec import PackageSet
-from .scm import CvsScm, GitScm, SvnScm, UrlScm, ScmOverride, auditFromDir
+from .scm import CvsScm, GitScm, SvnScm, UrlScm, ScmOverride, auditFromDir, CheckoutAssert
```
jkloetzke: Please move `CheckoutAssert` into input.py. This is unrelated to SCMs.

pym/bob/input.py
```diff
@@ -947,6 +952,12 @@ def getScmDirectories(self):
             dirs.update(s.getDirectories())
         return dirs
 
+    def getCheckoutAsserts(self):
```
jkloetzke: That should not be needed. What **is** needed is that `getDigestScript` also takes the assertions into account. I guess you introduced `getCheckoutAssert` and treated it like an SCM to trigger the checkout step if the asserts are changed. But this should be really be done through `getDigestScript`. Otherwise release mode builds will not be affected.

pym/bob/input.py
```diff
@@ -1574,7 +1585,12 @@ def __init__(self, recipeSet, recipe, sourceFile, baseDir, packageName, baseName
             scm["__source"] = sourceName
             scm["recipe"] = "{}#{}".format(sourceFile, i)
             i += 1
-        self.__checkout = (checkoutScript, checkoutDigestScript, checkoutSCMs)
+        checkoutAsserts = recipe.get("checkoutAssert", [])
+        if isinstance(checkoutAsserts, dict):
+            checkoutAsserts = [checkoutAsserts]
+        elif not isinstance(checkoutAsserts, list):
+            raise ParseError("checkoutAssert must be a dict or a list")
```
jkloetzke: Shouldn't that be handled already by the schema check? BTW, the schema allows only a list anyway...

pym/bob/input.py
```diff
@@ -2290,6 +2310,7 @@ class RecipeSet:
                 schema.Optional('relativeIncludes') : bool,
                 schema.Optional('cleanEnvironment') : bool,
                 schema.Optional('tidyUrlScm') : bool,
+                schema.Optional('noLicenseMetaEnv') : bool,
```
jkloetzke: Unrelated...

pym/bob/scm/checkoutAssert.py
```diff
@@ -0,0 +1,50 @@
...
+    })
+
+    def __init__(self, spec):
+        self.__file = spec['file']
+        self.__digestSHA1 = spec['digestSHA1']
+        self.__start = spec.get('start', None)
+        self.__end = spec.get('end', None)
+
+    def asScript(self):
+        if self.__start is not None and self.__end is not None:
```
jkloetzke: What about only giving `start` *or* `end`? I think they should default to `1` resp. `last line`...

pym/bob/scm/checkoutAssert.py
```diff
@@ -0,0 +1,50 @@
...
+            if [[ "$COMPUTED_SHA1" != "{SHA1SUM}" ]]; then
+                echo "Error: CheckoutAssert {FILE} checksums did not match!" 1>&2
+                echo "Specified: {SHA1SUM}" 1>&2
+                echo "Now: $COMPUTED_SHA1" 1>&2
+                exit 1
+            fi
+            """).format(CALC=calc,
+                    FILE=self.__file, SHA1SUM=self.__digestSHA1)
+
+    def getAssert(self):
```
jkloetzke: That should be replaced by a `getDigestScript` method...

doc/manual/configuration.rst
```diff
@@ -468,6 +468,33 @@ If a step depends on a certain variable then the result of the following step
...
+================= ========================================================================================
+file              | The file in the workspace to check.
+digestSHA1        | Digest of the file / part. Either pre calculate it using `sha1sum` command or take the
+                  | output of the first (failing) run. 
+start             | Optionally. Defaults to 1.
+end               | Optionally. Defaults to last line of file.
+================= ========================================================================================
+
+To assert only a part of the file hasn't changed use ``start`` and ``end``.
+Bob uses `sed` to grep the file content between ``start`` and ``end`` making it possible to use regular
```
jkloetzke: I'm not sure that this is a good idea. It ties Bob's behavior to the version of sed that is installed on the host. We might even run into locale and escaping issues. I would only support line numbers until a use case comes up that needs advanced features.

rhubert: I used `sed` before to get the content between `start` and `end`. I don't know a pure bash solution for this. Do you think it's ok to use `head` and `tail`? Or add this partial file hashing to the `bob-hash-engine`?

jkloetzke: I think using sed is ok. I would just restrict start and end to numbers. Exposing the more advanced features of sed might prove difficult for the future.

jkloetzke:
Thanks. :+1:

PR #199, opened 2018-04-11
input: add ignoreRoots

rhubert:
To improve package parsing speed ignoreRoots can be specified in a
defaultYaml as a list of regular expression.

With a larger recipe set I measured for `bob ls`:
```
                  old      ignoreRoots (for all but the interesting root recipe)
clean:            1m4s     32,4s
recipe change:   56,3s     22,9s
```


jkloetzke:
Great feature. I think we can spare the alias part because it is possible to have an alias that maps to a non-existent package. And they are not really root recipes anyway.

On the other hand I would make the syntax a bit more generic. What about doing it like the `filter` keyword (http://bob-build-tool.readthedocs.io/en/latest/manual/configuration.html#filter) and allow a list of positive and negative glob patterns? Then we could call it `rootOverrides`:

```yaml
rootOverrides:
    - "foo-*"
    - "!foo-bar-*"
```

That would let all "foo-" root recipes but exclude the sub-set of "foo-bar-" recipes. It would also make the setting a "positive" setting instead of a "negative" one. You list all the root recipes that you're interested in. I guess this is what most people need and with the "!..." syntax we allow negative matches too. What do you think?

rhubert:
Ok - I changed this to `rootOverrides`. Even though I think this is more a `rootFilter` than a `rootOverride` ?

pym/bob/input.py
```diff
@@ -59,7 +59,7 @@ def __maybeGlob(pred):
 
 def maybeGlob(pattern):
     if isinstance(pattern, list):
-        return [ __maybeGlob(p) for p in pattern ]
+        return [ __maybeGlob(p) for p in pattern ] if len(pattern) != 0 else None
```
jkloetzke: That should be handled before calling `maybeGlob`. Otherwise the `filter` keyword will behave differently.

rhubert:
done

PR #203, opened 2018-05-25
Fix processing of defines

### LINKED ISSUE #200: Allow "=" in defines
It is currently not possible to have an equal sign in a define because of the way the argument is split. Is this intended behaviour?

My use-case is that I'm using a define to set parameters that will be passed to `make` during the build step. I'd like to use something like `bob ls -DMAKE_OPTIONS="-j4 VERBOSE=1" <package>`. It doesn't really make sense for `bob ls` but this currently fails with `error: Malformed define: MAKE_OPTIONS=-j4 VERBOSE=1`. `bob dev`, which is the actual use-case, even fails with an internal exception caused by using the function `processDefines`.

Both issues could be fixed by allowing equal  signs inside defines.
I'd propose to change the line that splits the argument from
```python
d = define.split("=")
```
to
```python
d = define.split("=", 1)
```
which allows at most one split and thus would put the complete string found after the first "=" into the define.

What do you think?

Side effect would be, that the `else` branch could be removed and the internal exception would not occur anymore (which occurs because there the `parser` object is used in this branch, but isn't provided as function argument to `processDefines`).

If you're fine with these changes, I could provide a PR.

ThomasFeher:
Defines containing "=" caused Bob to exit with an error, which is not intended.

Furthermore simplified `processDefines` function as suggested by @jkloetzke and used it consistently over the whole code base.

Fixes #200

### DIFF
--- a/pym/bob/cmds/build.py
+++ b/pym/bob/cmds/build.py
@@ -11,7 +11,7 @@ from ..input import RecipeSet
 from ..state import BobState
 from ..tty import colorize
 from ..utils import asHexStr, hashDirectory, hashFile, removePath, \
-    emptyDirectory, copyTree, isWindows
+    emptyDirectory, copyTree, isWindows, processDefines
 from datetime import datetime
 from glob import glob
 from pipes import quote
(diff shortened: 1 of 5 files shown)


jkloetzke:
Looks quite good. :+1:

One minor thing: could you just squash your commits and keep the commit message style like the rest? Maybe something like:

* utils: introduce common processDefines function
* bob: consistently use processDefines function everywhere

ThomasFeher:
done

jkloetzke:
Thank you. :+1:

PR #205, opened 2018-06-13
build: do not generate audit trail in dev mode

rhubert:
In dev mode a common used scenario is to rebuild after a small code
change. For this generating the audit trail may take a significant amount of
time or even more time compared to the real compile time.

For a large module with some dependencies a

 ```bob dev -b <module>```

takes 1:15min with audit trail and 0:30min without after modifying a
single line of code.

This change disable the audit trail for 'bob dev' unless upload is
enabled and adds some controll switches.

### DIFF
--- a/pym/bob/audit.py
+++ b/pym/bob/audit.py
@@ -3,16 +3,18 @@
 #
 # SPDX-License-Identifier: GPL-3.0-or-later
 
+from . import BOB_INPUT_HASH
 from .errors import BuildError, ParseError
 from .scm import GitAudit, SvnAudit, UrlAudit, auditFromData
 from .tty import colorize
-from .utils import asHexStr, hashFile
+from .utils import asHexStr, hashFile, binStat
 from datetime import datetime, timezone
 import gzip
 import hashlib
 import io
 import json
 import os
+import pickle
 import schema
 import struct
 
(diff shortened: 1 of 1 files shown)


rhubert:
bob-profile

looks like it is the schema validation. Any ideas how to make this faster?

jkloetzke:
We could generate a pickle'ed version of the data structures next to the audit.json.gz. If it's up-to-date and Bob hasn't changed then we can use it. Otherwise the audit.json.gz is read and the cache file is written. If we generate the audit.json.gz then both versions are written. This would prevent any validation if we generate the file locally and still validate it once if it was downloaded.

Somethin along these lines:

```python

class Audit:
    @classmethod
    def fromFile(cls, file):
        try:
            cacheName = file + ".pickle"
            cacheKey = binStat(file) + BOB_INPUT_HASH
            with open(cacheName, "rb") as f:
                persistedCacheKey = f.read(len(cacheKey))
                if cacheKey == persistedCacheKey:
                    return pickle.load(f)
        except (EOFError, OSError, pickle.UnpicklingError):
            pass

        audit = cls()
        try:
            with gzip.open(file, 'rb') as gzf:
                audit.load(gzf)
            with open(cacheName, "wb") as f:
                f.write(cacheKey)
                pickle.dump(audit, f, -1)
        except OSError as e:
            print(colorize("Error loading audit: " + str(e), "33"))
        return audit

```

Likewise the `save()` method should write both files. That should almost eliminate the json loading overhead...

rhubert:
Looks good. This reduces the overhead in the case from 1:15min to 37sec.

jkloetzke:
That's still a 7s overhead which is more than expected. I guess this can be further optimized by storing the audit trail parts in a sqlite database instead. But that is a bit more involved and may be worth a separete MR.

Should I merge it or is there still some testing needed?

rhubert:
Please merge..

PR #207, opened 2018-06-25
build: fix KeyError in loadBuildState

rhubert:
Fix KeyError occurring on a bob dev/build --resume without having a valid
.bob-state / bob was not run before.

### DIFF
--- a/pym/bob/cmds/build.py
+++ b/pym/bob/cmds/build.py
@@ -474,9 +474,9 @@ esac
 
     def loadBuildState(self):
         state = BobState().getBuildState()
-        self.__wasRun = dict(state['wasRun'])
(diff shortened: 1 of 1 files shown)


pym/bob/cmds/build.py
```diff
@@ -474,9 +474,12 @@ def saveBuildState(self):
 
     def loadBuildState(self):
         state = BobState().getBuildState()
-        self.__wasRun = dict(state['wasRun'])
```
jkloetzke: Better use `dict(state.get('wasRun', {}))`. That'll save the else clause...

rhubert: :+1: too easy..

PR #208, opened 2018-06-26
build: speedup status by stopping recursion

rhubert:
As for bob project stop the recursive search for checkoutSteps when
seeing a processed step again.


rhubert:
Well - you're right... This was my first solution but I thought I can omit the additional set. :-/

jkloetzke:
And you almost convinced me. :wink: I was just about to merge it...

Thanks.

PR #209, opened 2018-06-28
scmOverrides: allow use of variables

rhubert:
Override scms having a common used set of overrides controlled by variables.

This reduces the number of overrides needed when you entered branching hell for your project... :see_no_evil: :hear_no_evil:

I'm not sure if it's a good idea to do the substitution only with the environment from `default.yaml` / command line but I have the feeling that substitution when constructing the `CheckoutStep` using the `fullEnv` is too late. It gives the recipes the possibility to change to override again which might be a pitfall.  :confused:

### DIFF
--- a/pym/bob/scm/scm.py
+++ b/pym/bob/scm/scm.py
@@ -27,9 +27,10 @@ class ScmOverride:
         (self.__match, self.__del, self.__set, self.__replaceRaw) = s
         self.__init()
 
-    def __doesMatch(self, scm):
+    def __doesMatch(self, scm, env):
         for (key, value) in self.__match.items():
             if key not in scm: return False
+            value = env.substitute(value, "svmOverride::match")
             if not fnmatch.fnmatchcase(scm[key], value): return False
         return True
 
@@ -41,15 +42,23 @@ class ScmOverride:
         return ((self.__match, self.__del, self.__set, self.__replace) ==
             (other.__match, other.__del, other.__set, other.__replace))
 
-    def mangle(self, scm):
+    def __applyEnv(self, env):
+        rm = [ env.substitute(d, "svmOverride::del") for d in self.__del ]
+        set = { k : env.substitute(v, "svmOverride::set"+k) for (k,v) in self.__set.items() }
+        replace = { k : env.substitute(v, "svmOverride::replace"+k) for (k,v) in self.__replace.items() }
+        return rm, set, replace
+
+    def mangle(self, scm, env):
         ret = False
-        if self.__doesMatch(scm):
+        if self.__doesMatch(scm, env):
+            rm, set, replace = self.__applyEnv(env)
+
             ret = True
             scm = scm.copy()
-            for d in self.__del:
+            for d in rm:
                 if d in scm: del scm[d]
-            scm.update(self.__set)
-            for (key, (pat, repl)) in self.__replace.items():
+            scm.update(set)
+            for (key, (pat, repl)) in replace.items():
                 if key in scm:
                     scm[key] = re.sub(pat, repl, scm[key])
         return ret, scm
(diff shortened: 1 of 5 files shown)


jkloetzke:
I have thought about it for some time now and I think we should make the substitution late when creating the SCM object. That is in `ScmOverride.mangle()`. Otherwise, when doing the substitution just with the initial environment, the behavior will be inconsistent. Suppose you have the following recipe, where `BRANCH` is set by some other recipe (an not in the default environment):

```yaml
environment:
    FALLBACK_BRANCH: "master"

checkoutSCM:
    scm: git
    url: ...
    branch: "${BRANCH}"
```

What should be the result when applying the following override?

```yaml
scmOverrides:
    -
        match:
            branch: "${BRANCH}"
        set:
            branch: "${FALLBACK_BRANCH:-${BRANCH}}"
```

With the current approach you could neither match the branch (because the substitiution of the the attributes is done with the final environment) nor could you replace it with something from the final environment.

The drawback will be that the package calculation will slow down. But an inconsistent behavior would be worse IMO.

pym/bob/scm/scm.py
```diff
@@ -41,15 +43,18 @@ def __eq__(self, other):
         return ((self.__match, self.__del, self.__set, self.__replace) ==
             (other.__match, other.__del, other.__set, other.__replace))
 
-    def mangle(self, scm):
+    def mangle(self, scm, env=Env()):
+        override = copy.deepcopy(self)
```
jkloetzke: Quite frankly I don't like the approach. As this is in some of the most performance critical paths of Bob I would prefer a more "classic" approach. :wink: 

```python
if self.__doesMatch(scm, env):
    rm, set, replace = self.__applyEnv(env)
    for d in rm:
...
```

The `__doesMatch` method could substitute the value only if the key is present. That may save some cycles if a key is missing. The `__applyEnv` method should just return a tuple of the substituted fields instead of mutating an object copy.

pym/bob/scm/scm.py
```diff
@@ -60,6 +65,15 @@ def __str__(self):
                 + (("set: " + str(self.__set)+ "\n") if self.__set else "")
                 + (("replace: " + str(self.__replaceRaw)) if self.__replaceRaw else "")).rstrip()
 
+
+    def __substitute(self, env):
+        for attr in ['__match', '__del', '__set', '__replaceRaw']:
+            i = getattr(self, "_ScmOverride"+attr)
```
jkloetzke: :scream: Uhh, that works but is almost unreadable for people who are not Python pro's... :wink:

rhubert: :disappointed: it took me some time to google that :wink:

rhubert:
Performance impact:

```
$ rm .bob-tree.* .bob-packages* 
$ time bob ls
```

### without Overrides

#### bob-master
```
real	1m25,753s
user	1m24,936s
sys	0m0,708s
```

#### this branch
```
real	1m26,179s
user	1m25,204s
sys	0m0,860s
```

### with Overrides (87 overrides active)

#### bob-master
```
real	1m27,200s
user	1m26,232s
sys	0m0,840s
```

#### this branch
```
real	1m31,959s
user	1m31,000s
sys	0m0,828s
```

jkloetzke:
Looks good. The performance impact is there but IMHO in an acceptable range.

Is that something that needs to go into Bob 0.14? I'm a bit reluctant to merge new features during stabilization. OTOH it should not cause any harm...

rhubert:
Some teams would love to see it in 0.14

jkloetzke:
Ok, then we go for 0.14. Would you add some words in the respective chapter in the documentation and update the release notes? Then I'll merge it...

rhubert:
done

PR #210, opened 2018-06-28
scripts: give feedback when stopping execution

rhubert:
As stopping bob could take some time (write the state) give some
feedback that bob is stopping. Otherwise many users press ctrl+c
again leading to a incomplete written state.

### DIFF
--- a/pym/bob/scripts.py
+++ b/pym/bob/scripts.py
@@ -126,6 +126,7 @@ def catchErrors(fun, *args, **kwargs):
         print(e, file=sys.stderr)
         ret = 1
     except KeyboardInterrupt:
+        print(colorize("Stopping execution. (Don't press Ctrl+C again!)", "32;1"))
         ret = 2
     except ImportError as e:
         if e.name:


jkloetzke:
> As stopping bob could take some time (write the state) give some
> feedback that bob is stopping. Otherwise many users press ctrl+c
> again leading to a incomplete written state.

While this may be the perception I don't think this is really true. While running the state should always be written back on every change. The only things that are written on shutdown:

* the build-id cache
* deleting the lock-file

So my expectation is that pressing ctrl+c again just leads to a stale lock file. Other than that the state should be consistent. If it's not then we have a different problem.

Also, instead of just printing "don't press crtl+c again", why not catch it? Maybe (untested):

```python
import signal

def handler(signum, frame):
    print(colorize("Bob is already shutting down. Please wait...", "32;1"))

...

         print(e, file=sys.stderr)
         ret = 1
     except KeyboardInterrupt:
         signal.signal(signal.SIGINT, handler) # catch subsequent ctrl+c
         ret = 2
     except ImportError as e:
         if e.name:
```

What do you think? That should even work on Windows...

jkloetzke:
That looks unsuspicious. So I guess the original print also showed up only immediately before Bob exited, right? My feeling is that most of the waiting time is coming from the garbage collector. What I don't understand is why a 2nd ctrl+c typically causes Bob to exit immediately. There must be something going on.

So if you quickly press ctrl+c twice do you have a stale lock file left behind? If not then it was a clean exit. Otherwise there is room for improvement...

rhubert:
No - there is no stale lock file. I think this is not needed.

PR #211, opened 2018-07-03
url: tar: use consistent permissions

rhubert:
Add two flags to `tar` invocation that will provide consistent behavior
whether `bob` is executed as root or a normal user.

### DIFF
--- a/pym/bob/scm/url.py
+++ b/pym/bob/scm/url.py
@@ -40,7 +40,7 @@ class UrlScm(Scm):
     ]
 
     EXTRACTORS = {
-        "tar"  : ("tar xf", "--strip-components={}"),
+        "tar"  : ("tar -x --no-same-owner --no-same-permissions -f", "--strip-components={}"),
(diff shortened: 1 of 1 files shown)


pym/bob/scm/url.py
```diff
@@ -40,11 +40,11 @@ class UrlScm(Scm):
     ]
 
     EXTRACTORS = {
-        "tar"  : ("tar xf", "--strip-components={}"),
-        "gzip" : ("gunzip -kf", None),
-        "xz"   : ("unxz -kf", None),
-        "7z"   : ("7z x -y", None),
-        "zip"  : ("unzip -o", None),
+        "tar"  : ("tar xf", "--strip-components={}", "--no-same-owner --no-same-permissions"),
```
jkloetzke: I think another field is not needed. Shouldn't `tar -x --no-same-owner --no-same-permissions -f` as command suffice?

rhubert:
Yes, this works as well...

PR #212, opened 2018-07-05
input: fix missing / duplicated indirectPackages

rhubert:
With 38bc5df1 the behavior of the indirectPackages filter has been
changed. This restores the old behavior avoiding duplicate/missing
dependencies.

I added a small repocase for the 'missing' dependencies. Maybe it was
not correct before 38bc5df1 and this is not a bug? (Although it looks
strange to have a dependency twice in the audit trail..)

[test.tar.gz](https://github.com/BobBuildTool/bob/files/2167846/test.tar.gz)

The main difference is, that in 

```
            "package" : "a-foo",
            "step" : "build",
```

the dependency for `b` is missing:

Bob 0.14:
``` 
         "dependencies" : {
            "args" : [
               "4cf372d74153954ce130c0f5407baf00841a18dd", a-bar dist
               "8715f688fe40af6bc2d1bdd8651afac46c366cd9"  c dist
            ]
         },
```

Bob 0.13:
```  
         "dependencies" : {
            "args" : [
               "9025a0841f1100f142dd1e3a62de9c35d7ac259e", a-bar dist
               "094ec0234f473af316d230332ed2d4ff073f4e76", b dist
               "e825a6f131bee1adff00890205ed68b6bc1fc8e6"  c dist
            ]
         },
```

### DIFF
--- a/pym/bob/input.py
+++ b/pym/bob/input.py
@@ -1583,6 +1583,29 @@ class UniquePackageList:
     def result(self):
         return self.ret
 
+class DepTracker:
+
+    __slots__ = ('package', 'isNew', 'usedResult')
+
+    def __init__(self, package):
+        self.package = package
+        self.isNew = True
+        self.usedResult = False
+
+    def prime(self):
+        if self.isNew:
+            self.isNew = False
+            return True
+        else:
+            return False
+
+    def useResultOnce(self):
+        if self.usedResult:
+            return False
+        else:
+            self.usedResult = True
+            return True
+
 class Recipe(object):
     """Representation of a single recipe
 
@@ -1955,6 +1978,7 @@ class Recipe(object):
         subTreePackages = set()
         directPackages = []
         indirectPackages = []
+        provideDeps = UniquePackageList(self.__raiseIncompatibleProvided)
         results = []
         depEnv = env.derive()
         depTools = tools.derive()
@@ -1981,24 +2005,26 @@ class Recipe(object):
                 e.pushFrame(r.getPackageName())
                 raise e
 
-            p2 = thisDeps.setdefault(dep.recipe, p)
-            if p2 is p:
-                # new pacakage
+            # A dependency should be named only once. Hence we can
+            # optimistically create the DepTracker object. If the dependency is
+            # named more than one we make sure that it is the same variant.
+            depTrack = thisDeps.setdefault(dep.recipe, DepTracker(p))
+            if depTrack.prime():
                 directPackages.append(p)
-                # add provided dependencies at the end
-                if dep.useDeps: indirectPackages.extend(p._getProvidedDeps())
-                # result is only picked once
-                if dep.useBuildResult: results.append(p)
-            elif p.getVariantId() != p2.getVariantId():
+            elif p.getVariantId() != depTrack.package.getVariantId():
                 self.__raiseIncompatibleLocal(p)
             else:
                 warnDepends.show("{} -> {}".format(self.__packageName, dep.recipe))
 
-            # pick up various other results of package
+            # pick up various results of package
             for (n, s) in states.items():
                 if n in dep.use:
                     s.onUse(p.getPackage()._getStates()[n])
                     if dep.provideGlobal: depStates[n].onUse(p.getPackage()._getStates()[n])
+            if dep.useDeps:
+                indirectPackages.extend(p._getProvidedDeps())
+            if dep.useBuildResult and depTrack.useResultOnce():
+                results.append(p)
             if dep.useTools:
                 tools.update(p._getProvidedTools())
                 if dep.provideGlobal: depTools.update(p._getProvidedTools())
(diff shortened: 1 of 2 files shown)


rhubert:
Maybe there is still something missing. One package has a different build ID. I'll look into this on Monday...

jkloetzke:
Do you mean from the attached test case or from a real project? Also, if the variant-ID is the same, then something else is probably the root cause.

jkloetzke:
Ok, now I got it. What's going on here is the following: in the dependency list there is a package `b` where the result is not picked up. On the other hand the same package is provided by some other package (`a-bar`). Effectively all provided dependencies are picked up with a `use: [result, deps]` declaration if they are not already in the dependency list. The question is what should happen now if the provided package was already in the dependency list:

1. Stick to what is defined in the recipe. The recipe defined `use: []` for package `b`. Bob 0.14 now sticks to this declaration and drops the provided dependency of `b` by `a-bar`. Note that transitive dependencies are still picked up, though.
2. "Merge" the use declaration of the recipe and the provided dependencies. If the result was not picked up yet then Bob 0.13 would add the result to the build step arguments.

Having thought about the problem for some time now I think the old behavior is correct. I'll comment on the change itself separately...

pym/bob/input.py
```diff
@@ -2015,10 +2015,12 @@ def prepare(self, pathFormatter, inputEnv, sandboxEnabled, inputStates, inputSan
         # filter indirect packages and add to result list
         tmp = indirectPackages
         indirectPackages = []
+        cache = {}
```
jkloetzke: By tracking it separately a package could now be in the direct *and* indirect dependency list of a package. That should not happen. If the package is in the dependency list then it is "direct". Otherwise it is "indirect".

pym/bob/input.py
```diff
@@ -2015,10 +2015,12 @@ def prepare(self, pathFormatter, inputEnv, sandboxEnabled, inputStates, inputSan
...
         for p in tmp:
-            p2 = thisDeps.setdefault(p.getPackage().getName(), p)
+            p2 = cache.setdefault(p.getPackage().getName(), p)
             if p2 is p:
-                results.append(p)
+                if not p in results:
```
jkloetzke: This (implicitly) uses the variant-id of a step to see if if it is already in the result list. But identical packages with different names are (and were) allowed. Oh, and this is O(n^2) in the worst case.

pym/bob/input.py
```diff
@@ -1983,12 +1985,17 @@ def prepare(self, pathFormatter, inputEnv, sandboxEnabled, inputStates, inputSan
...
+                # new package
                 directPackages.append(p)
                 # add provided dependencies at the end
                 if dep.useDeps: indirectPackages.extend(p._getProvidedDeps())
                 # result is only picked once
-                if dep.useBuildResult: results.append(p)
+                if dep.useBuildResult:
+                    p2 = inResults.setdefault(dep.recipe, p)
+                    if p2 is p:
+                        results.append(p)
```
jkloetzke: This check is not needed. Each package (name) can only show up once here because of the `thisDeps.setdefault()` above.

pym/bob/input.py
```diff
@@ -2012,13 +2019,23 @@ def prepare(self, pathFormatter, inputEnv, sandboxEnabled, inputStates, inputSan
                     env.update(sandbox.getEnvironment())
                     if dep.provideGlobal: depEnv.update(sandbox.getEnvironment())
 
+        # if a dependency is a direct dependency with a empty use-section and also provided by some other package use the buildResult
+        for p in providedPackages:
```
jkloetzke: Now it's wrong. The `indirectPackages` list is the only correct one in this case. We pick up the results of provided packages only iff the package that provided them had `deps` in its use clause (which is the default). The whole hunk can be removed.

pym/bob/input.py
```diff
@@ -2012,13 +2019,23 @@ def prepare(self, pathFormatter, inputEnv, sandboxEnabled, inputStates, inputSan
...
+                    results.append(p)
+
         # filter indirect packages and add to result list
         tmp = indirectPackages
         indirectPackages = []
         for p in tmp:
             p2 = thisDeps.setdefault(p.getPackage().getName(), p)
             if p2 is p:
-                results.append(p)
+                p2 = inResults.setdefault(p.getPackage().getName(), p)
```
jkloetzke: This can only happen because of the above (wrong) hunk. The thing is that you may have to pick up the result if `p2 is not p`.

rhubert:
There are still packages with different dependencies. I'll diff the audit and come back...

jkloetzke:
Strange. I have the feeling that the problem is not understood completely yet.

There is are two things that might still be different:

1. Naming a dependency multiple times in a recipe with different use flags. That worked in 0.13 but will fail with the current code. Might be worth fixing.
2. A recipe that creates packages with the same variant-id but that provides different dependencies.

Both cases are patological IMHO. But let's see what the audit trail tells...

jkloetzke:
I think I have an idea what goes wrong. The code might introduce the same indirect dependency multiple times. I'll post an update later this afternoon when I have some time...

jkloetzke:
I think there is no way how we can retain full 0.13 compatibility. The same bug that I introduced with my last commit was already in 0.13 in some other form. There you could also end up having the same dependency multiple times in a package. I don't fully understand yet how this can happen but I have some recipes that show this behavior.

I also spotted another regression with provided dependencies. If a dependency is disabled due to a failing "if" condition but also provided by some other dependency then Bob 0.13 will not forward it but Bob 0.14 will. I'm not yet sure what should be the right behavior.

jkloetzke:
> Today I diffed the audit trails. (After writing a tool resolving the artifact IDs, sorting dependencies,...) 

Sounds like a first step to some usefull audit trail tooling... :sunglasses:

> The difference is indeed a duplicate dependeny in 0.14 .

I have a fix for that but then I found some duplicate dependencies in 0.13 as well. :disappointed: 

> I tried to create a small repocase but without success until end of work. The package has some failing if condition to the duplicate depends so it's likely the same 'issue' you found. Do you think it can help to have a repocase for this?

I don't think it makes sense to invest more time on your side. I will update this MR with what I think is the correct solution. Then you can test it on your side. I expect that there will be differences but they should be only duplicate dependencies in 0.13. I don't see a point carrying over the buggy behavior. So there will be packages that do not match between 0.13 and 0.14...

rhubert:
Sound's like a good plan. It should not cause any trouble if we can't use artifacts from 0.13 due to a bug fixed in 0.14. We just need to set `bobMinimumVersion` to 0.14

In case you need the tool for your investigations you can find the first version in the contrib/audit folder of https://github.com/rhubert/bob/tree/auditDiffer ...

rhubert:
:ok_hand: As predicted with this fix the only difference between 0.14 and 0.13 are duplicated dependencies in 0.13. :+1: 

Please change the author of the 1st commit before merging.

PR #213, opened 2018-07-17
misc: query-scm: stop recursion for known packages

rhubert:
As we did before for bob project and bob status we can  stop the recursive
search when seeing a processed package again.


jkloetzke:
This is getting annoying. :disappointed: All the recursive package traversing should probably be refactored in the future to have a single correct algorithm. AFAICT `query-meta` and `clean` should suffer from the same problem...

Thanks anyway.

PR #214, opened 2018-07-18
No link deps

rhubert:


jkloetzke:
I'm inclined to merge this *after* 0.14 is out. Otherwise the release will drag out even longer...

rhubert:
OK :relaxed: This is not needed for 0.14.

PR #215, opened 2018-09-10
Improve parsing speed and memory consumption

jkloetzke:
Get rid of as many intermediate objects as possible during package calculation. This gives quite a speed bonus and dramatically lowers the memory consumption.

@kolewu @ThomasFeher This breaks the plugin API by removing some (hopefully) unused parameters. Old plugins should still work as long as they do not use these parameters. They are filled with stub values if the plugin announces an old API version. It would be good to get your feedbach if this is a deal-breaker for you.


ThomasFeher:
:+1: We did not use that removed parameters. Running tests on our code base showed no problems.

Sorry for the delayed reply.

PR #217, opened 2018-10-06
Parallel builds

jkloetzke:
Add the '-j' option to build packages in parallel.

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -1,8 +1,7 @@
 language: python
 python:
-    - "3.3"
-    - "3.4"
     - "3.5"
+    - "3.6"
 
 install:
     - pip install PyYAML
--- a/Makefile
+++ b/Makefile
@@ -25,7 +25,7 @@ SPHINX := $(shell command -v sphinx-build 2>/dev/null)
 
 .PHONY: all install pym check doc
 
-all: bin/namespace-sandbox pym check doc
+all: bin/namespace-sandbox check pym doc
 
 bin/namespace-sandbox: $(patsubst %,$(DIR)/%,$(SOURCE) $(HEADERS))
 	@gcc -o $@ -std=c99 $^ -lm
@@ -58,6 +58,7 @@ ifdef SPHINX
 endif
 
 check:
+	@python3 -c 'import sys; sys.exit(0 if sys.hexversion >= 0x03050000 else 1)' || { echo "Pyton >= 3.5.0 is required!"; exit 1 ; }
 	@python3 -c 'import schema' || { echo "Module 'schema' missing. Please install: 'pip3 install --user schema'..." ; exit 1 ; }
 	@python3 -c 'import yaml' || { echo "Module 'yaml' missing. Please install: 'pip3 install --user PyYAML'..." ; exit 1 ; }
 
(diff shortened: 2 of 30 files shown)


Kalabaza:
I think the generators (for qt and eclipse) must also be modified to have at least an extra configuration with the -j flag.

Also, I tried adding the flag to one of the tests (rerun-failed-build since it has some code and it is actually being compiled), but after deleting the semi-colon from the printf of the foo.c file the compilation error is not seen on the IDE, maybe the new text interfieres with the parsing of the log:

```
[Start    1] NOTE: CHECKOUT  root - dev/src/root/1/workspace 
[End      1] NOTE: CHECKOUT  root - dev/src/root/1/workspace 
[Start    2] NOTE: BUILD     root - dev/build/root/1/workspace
[End      2] ERR : BUILD     root - dev/build/root/1/workspace
[...      2] NOTE: |root| /home/user/BobBuildTool/test/rerun-failed-build/dev/src/root/1/workspace/foo.c: In function 'main':
[...      2] NOTE: |root| /home/user/BobBuildTool/test/rerun-failed-build/dev/src/root/1/workspace/foo.c:6:2: error: expected ';' before 'return'
[...      2] NOTE: |root|   return 0;
[...      2] NOTE: |root|   ^~~~~~
[...      2] NOTE: |root| Step failed with return status 1; Command: gcc -o foo $1/foo.c
[...      2] NOTE: |root| Call stack (most recent call first)
[...      2] NOTE: |root|     #0: Recipe root, line 2, in main
```

jkloetzke:
Tested with QtCreator. LGTM...

Kalabaza:
Thanks, now the errors are parsed correctly again.

rhubert:
making bob with python 3.4 fails.

```
BobBuildTool$ make
Listing 'pym'...
Listing 'pym/bob'...
Compiling 'pym/bob/__init__.py'...
Compiling 'pym/bob/archive.py'...
***   File "pym/bob/archive.py", line 84
    async def uploadPackage(self, step, buildId, audit, content):
            ^
SyntaxError: invalid syntax
```

Checking for python >= 3.5 should also be done in the Makefile...

jkloetzke:
Added a check that hopefully should trigger early enough.

kkendzia:
Building with "bob dev -j4 <pkg>"
When the CheckoutStep (checkoutscript in this case) fails I run into an endless loop.
There is python script generating files which failed with a "FileNotFoundError" leading to a stuck buildId calculation.

jkloetzke:
Hopefully fixed with the last commit... :)

kkendzia:
Thanks! I deployed it on our test server and give a feedback in some days

kkendzia:
Thanks! I deployed it on our test server and give a feedback in some days

kkendzia:
Thanks! I deployed it on our test server and give a feedback in some days

kkendzia:
Now the buildId (downloading liveid) calculation stucks even without a failing step.

kkendzia:
Works fine for 1 week. Didn't have any problems and runs well

jkloetzke:
Thanks for testing and the numerous feedback. :+1:

Issue #187, opened 2018-01-12
'bob ls -a' should also show tool dependencies


jkloetzke:
Currently only direct and provided dependencies are shown by 'bob ls -a'. Forwarded tool dependencies are also built but are not visible in 'bob ls'.

PR #221, opened 2018-11-20
add option to build dependencies only

kkendzia:
I have a use case where it's only necessary to build dependencies without the package itself. I tried to added the option, but I am not very pleased with the _checkDepsOnly function. Maybe you have an idea.

### DIFF
--- a/pym/bob/cmds/build.py
+++ b/pym/bob/cmds/build.py
@@ -418,7 +418,7 @@ esac
 
         return fmt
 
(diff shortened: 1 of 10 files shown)


kkendzia:
Would be fine for me! Yeah I also tried <path>/* and encountered the same issue with tools.

jkloetzke:
I did have a look and solving the problem with the query language is not so easy as it seems. Nonetheless I think this is the correct approach given that #187 needs internally the same information. I would rather not solve this only for the local build with this PR.

Still, many thanks for submitting the PR in the first place.

PR #225, opened 2018-12-21
doc: improve examples to make nesting more visible

ThomasFeher:
From the headings it is not really visible that "build", "dev" and
"graph" belong to "command". Therefore the examples are changed to
contain "command", too. Also added one example for graph to make it
visible there, too.


jkloetzke:
Thanks. :+1:

PR #227, opened 2019-01-15
build: use start method "fork" on MSYS platforms

ThomasFeher:
instead of "forkserver", to start a new process.
Start method "forkserver" is only "available on Unix platforms, which
support passing file descriptors over Unix pipes" (see documentation of
"multiprocessing" module, section "Contexts and start methods"). This is
not the case for MSYS and leads to a "RuntimeError: received 0 items of
ancdata" in "/usr/lib/python3.7/multiprocessing/reduction.py", line 161,
in recvfds" for Python 3.7.

### DIFF
--- a/pym/bob/cmds/build/build.py
+++ b/pym/bob/cmds/build/build.py
@@ -125,7 +125,11 @@ def commonBuildDevelop(parser, argv, bobRoot, develop):
         loop = asyncio.get_event_loop()
         origSigInt = signal.getsignal(signal.SIGINT)
         signal.signal(signal.SIGINT, signal.SIG_IGN)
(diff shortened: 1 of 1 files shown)


jkloetzke:
Out of curiosity: does CTRL+C work, especially with `-j` >= 2. I must admit I did not think about the msys case when implementing the parallel build. :worried:

ThomasFeher:
Yes, CTRL-C works with more than one process. Jobs are canceled correctly and no remaining processes afterwards.

jkloetzke:
> Yes, CTRL-C works with more than one process. Jobs are canceled correctly and no remaining processes afterwards.

Glad to hear that it still works. :sweat_smile:  Thanks for the patch. :+1:

PR #230, opened 2019-02-07
Extend "bob status" to check all unpushed commits of a Git repository

gruberchr:
We propose to extend the behavior of `bob status` for Git repositories. We'd like to see not only unpushed commits on local branches but also from stash or the possibly detached HEAD for instance.

### DIFF
--- a/doc/manpages/bob-status.rst
+++ b/doc/manpages/bob-status.rst
@@ -118,10 +118,11 @@ Status codes can be interpreted as follows:
   :ref:`configuration-config-scmOverrides`. Pass ``--show-overrides`` to force
   the output of a ``STATUS`` line if the SCM is otherwise unmodified.
 - ``S`` = switched. The commit/tag/branch/URL is different from the recipe.
-- ``U`` = unpushed commmits. Git only. Some commits were made locally to the
-  configured branch but they have not yet been pushed to the remote.
-- ``u`` = unpushed branch(es). Git only. At least one branch exists with commits
-  that have not been pushed to a remote.
+- ``U`` = unpushed commits on configured branch. Git only. Some commits were made
+  locally to the configured branch but they have not yet been pushed to the remote.
+- ``u`` = unpushed commits not on configured branch. Git only. There are commits
+  on other local branches than the configured branch, on a possibly detached HEAD
+  or in the stash that have not been pushed to a remote.
 
 The command shows the status of the current checkout. If the recipe was changed and
 the next build would move a checkout to the attic then the current information still
--- a/pym/bob/scm/git.py
+++ b/pym/bob/scm/git.py
@@ -313,16 +313,17 @@ class GitScm(Scm):
                 status.add(ScmTaint.modified, joinLines("> modified:",
                     indent(output, '   ')))
 
-            # The following shows unpushed commits on all local branches.
+            # The following shows all unpushed commits reachable by any ref
+            # (local branches, stash, detached HEAD, etc).
             # Exclude HEAD if the configured branch is checked out to not
             # double-count them. Does not mark the SCM as dirty.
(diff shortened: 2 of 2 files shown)


jkloetzke:
Sounds reasonable. Could you please add a commit adapting the man-page (doc/manpages/bob-status.rst) accordingly?

gruberchr:
yes

gruberchr:
man-age adapted

gruberchr:
There are still two open issues, which are maybe relevant.

1. The status line now is `> unpushed commits` in both cases, first for unpushed commits on the main branch and second for all remaining unpushed commits. And they are shown separately, if both exist.
1. I encountered, that `bob status -v` generates output for unpushed commits, which are not on the main branch, if there are unpushed commits on the main branch as well, although this should only be shown with `-vv`. That means, if your repository is clean and you have unpushed commits on other branches nothing is shown with `-v`, but when you have unpushed commits on main branch as well, all unpushed commits are shown with `-v` the same way as with `-vv`. Is this desired behavior?

jkloetzke:
> There are still two open issues, which are maybe relevant.
> 
> 1. The status line now is `> unpushed commits` in both cases, first for unpushed commits on the main branch and second for all remaining unpushed commits. And they are shown separately, if both exist.

That's not really optimal. Maybe call it "unpushed branch commits" and "unpushed local commits"? I'm open for suggestions of better names...

> 
> 2. I encountered, that `bob status -v` generates output for unpushed commits, which are not on the main branch, if there are unpushed commits on the main branch as well, although this should only be shown with `-vv`. That means, if your repository is clean and you have unpushed commits on other branches nothing is shown with `-v`, but when you have unpushed commits on main branch as well, all unpushed commits are shown with `-v` the same way as with `-vv`. Is this desired behavior?

That is indeed surprising. It also partially contradicts the current man-page. What about adding the following patch?

```diff
diff --git a/pym/bob/cmds/build/status.py b/pym/bob/cmds/build/status.py
index 104b4e2..48e92c0 100644
--- a/pym/bob/cmds/build/status.py
+++ b/pym/bob/cmds/build/status.py
@@ -49,6 +49,9 @@ class PackagePrinter:
                 severity = NORMAL
             elif self.showClean:
                 severity = NORMAL
+            elif ScmTaint.unpushed_local in status.flags:
+                # show changes that are not considered as modifications with -v
+                severity = INFO
             else:
                 # otherwise show only with '-vv'
                 severity = DEBUG
```

That would consistenly show the detailed information about unpushed commits even is they are not considered a modification...

gruberchr:
I changed wording regarding status line

gruberchr:
> That would consistenly show the detailed information about unpushed commits even is they are not considered a modification.

I don't recommend this. It's too much information, if you have several local unpushed branches for instance. Actually `bob status -v` should really only show the commits, which make the repository dirty and not more.
I see, that this behavior could not be fixed easily. In this case I would start a separate issue regarding this.

jkloetzke:
> > That would consistenly show the detailed information about unpushed commits even is they are not considered a modification.
> 
> I don't recommend this. It's too much information, if you have several local unpushed branches for instance. Actually `bob status -v` should really only show the commits, which make the repository dirty and not more.

Sounds reasonable. I have never used `-v` very much.

> I see, that this behavior could not be fixed easily. In this case I would start a separate issue regarding this.

Yes, that's a different topic. It shoud be doable by extending the `ScmStatus` class to filter the description by severity. But let's keep that separated.

jkloetzke:
Thank you. :+1:

gruberchr:
Just for information, is there a planned date for next release?

jkloetzke:
I pushed the "fingerprinting" feature yesterday. We will test it here internally and I will make a release soon if it works as expected.

PR #232, opened 2019-02-16
Allow local files with the "url" SCM.

stefanreuther:
If the URL starts with '/', it is treated as a local file and is just copied
with 'cp' instead of downloaded using 'curl'.  This makes it possible to
easily build from tarballs in a sandbox that has no network access and no
'curl' command.

### DIFF
--- a/pym/bob/scm/url.py
+++ b/pym/bob/scm/url.py
@@ -88,7 +88,17 @@ class UrlScm(Scm):
     def asScript(self):
         options = "-sSgLf"
         if not self.__sslVerify: options += "k"
-        ret = """
+        ret = ""
+        if self.__url[0] == '/':
+            # Local files: copy only if newer (u), target never is a directory (T)
+            tpl = """
+{HEADER}
+mkdir -p {DIR}
+cd {DIR}
+cp -uT {URL} {FILE}
+"""
+        else:
+            tpl = """
 {HEADER}
 mkdir -p {DIR}
 cd {DIR}
@@ -103,7 +113,8 @@ else
         mv $F {FILE}
     )
 fi
-""".format(HEADER=super().asScript(), DIR=quote(self.__dir), URL=quote(self.__url),
+"""
+        ret = tpl.format(HEADER=super().asScript(), DIR=quote(self.__dir), URL=quote(self.__url),
            FILE=quote(self.__fn), OPTIONS=options)
 
         if self.__digestSha1:
--- a/pym/bob/input.py
+++ b/pym/bob/input.py
@@ -2517,6 +2517,7 @@ class RecipeSet:
                 schema.Optional('sandboxInvariant') : bool,
                 schema.Optional('uniqueDependency') : bool,
                 schema.Optional('mergeEnvironment') : bool,
+                schema.Optional('secureSSL') : bool,
             },
             error="Invalid policy specified! Maybe your Bob is too old?"
         )
--- a/test/url-scm-file/file.txt
+++ b/test/url-scm-file/file.txt
@@ -1,0 +1,1 @@
+This is a file.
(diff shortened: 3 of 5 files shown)


pym/bob/scm/url.py
```diff
@@ -88,7 +88,16 @@ def getProperties(self):
...
         options = "-sSgLf"
         if not self.__sslVerify: options += "k"
-        ret = """
+        ret = ""
+        if self.__url[0] == '/':
+            tpl = """
+{HEADER}
+mkdir -p {DIR}
+cd {DIR}
+cp {URL} {FILE}
```
jkloetzke: I guess we should add `-uT` to replicate the old behavior.

* `u` to update the file only if the source is newer or the destination does not exist.
* `T` to fail if the target name already exists as a directory.

stefanreuther: +1 for `-T` (if we depend on bash, dependingon GNU cp is ok?).

I wouldn't agree with `-u`: if I place a file to have it picked up by a recipe, I don't want it to ruin my day by silently failing a mtime check, even when swapping old and new files back and forth.

You're saying "old behaviour", what do you mean with that? I didn't see anything equivalent in the download branch, but must admit I haven't fully understood what it is doing, especially the `set`/`curl`/`mv` part. Could it be that the branches of the `if` are swapped? Would make more sense to me. In any case, the `mv` could probably also use a `-T`?

jkloetzke: > +1 for `-T` (if we depend on bash, dependingon GNU cp is ok?).

GNU coreutils are used at other places too. At least the Jenkins backend relies heavily on them.

> I wouldn't agree with `-u`: if I place a file to have it picked up by a recipe, I don't want it to ruin my day by silently failing a mtime check, even when swapping old and new files back and forth.

Hmm, I guess every build system is supposed to make such an "optimization".
 
> You're saying "old behaviour", what do you mean with that? I didn't see anything equivalent in the download branch, but must admit I haven't fully understood what it is doing, especially the `set`/`curl`/`mv` part. Could it be that the branches of the `if` are swapped? Would make more sense to me. In any case, the `mv` could probably also use a `-T`?

If the file exists in the workspace then `curl` is called with the `-z` option. This will request a file from the server that has been modified later than the given file. This is the equivilent to the `-u` cp switch. The rationale is that you only want to download a (possibly very large) file again if there is a newer version on the server. A `cp` may not take so long but if the file is very large it might still be significant.

The `mktemp`/`curl`/`mv` dance is there to prevent parial downloads in the workspace. Because of the earlier `if [ -e {FILE} ] ...` check the `mv` should never face a collision.

stefanreuther: > > I wouldn't agree with `-u`: if I place a file to have it picked up by a recipe, I don't want it to ruin my day by silently failing a mtime check, even when swapping old and new files back and forth.

> Hmm, I guess every build system is supposed to make such an "optimization".

Not if I told it to unconditionally copy a file before starting its work.

> If the file exists in the workspace then `curl` is called with the `-z` option. This will request a file from the server that has been modified later than the given file.

Ah, ok, missed that one.

I guess the proper way to solve this is to use a checksum in the recipe so bob sees a recipe change to force re-download?

> A `cp` may not take so long but if the file is very large it might still be significant.

Given that my recipes routinely have to copy entire installation trees between checkout, build and package this is a weak argument :)

> The `mktemp`/`curl`/`mv` dance is there to prevent parial downloads in the workspace. Because of the earlier `if [ -e {FILE} ] ...` check the `mv` should never face a collision.

I would have expected that in the "file does already exist" branch because in that case, a failed download will leave me with a partial file and will already have overwritten the old, good file (and, if I'm unlucky, fail to download the file next time due to the time condition).

PR #233, opened 2019-02-17
Streamline bob-status verbosity handling

### LINKED ISSUE #231: 'bob status -v' should only show detailed information about modified files
The definition of verbosity level `-v` is not precise regarding detailed information about changes, which do not modify a checkout workspace.

The man pages say:
> `-v` shows a detailed description of each listed flag. In particular the modified files of an SCM are listed.

Listing the modified files of an SCM would mean to show detailed information about flags `M` and `U`. Whereas showing detailed description of each listed flag would also include to show detailed information about flag `u`, if set.

The current implementation shows detailed information about all three flags `M`, `U` and `u`, when using verbosity level `-v` and when checkout workspace is modified, i.e. one or both of flags `M` and `U` are set. If only `u` is set nothing is displayed.

It would be more consistent, if `bob status -v` shows detailed information about modified files only (flags `M` and `U`). And with `bob status -vv` one can additionally get information about modifications related to flag `u`.

jkloetzke:
Handle the verbosity of detailed descriptions per flag. Each flag has a dedicated verbosity where modifying flags are shown by default.The detailed description of each flag is shown on the next severity level of that flag instead of all at once.

Fixes #231.

### DIFF
--- a/doc/manpages/bob-status.rst
+++ b/doc/manpages/bob-status.rst
@@ -78,8 +78,8 @@ Options
     Show checkouts that have active :ref:`configuration-config-scmOverrides`
     (``O``) even if the SCM is unchanged. Override information is always
     displayed if a checkout is shown but a ``STATUS`` line is normally only
-    emitted if the SCM was modified. See :ref:`manpage-bob-status-verbosity`
-    below.
+    emitted if the SCM was modified. Adding ``-v`` will additionally show the
+    detailed override status. See :ref:`manpage-bob-status-verbosity` below.
 
 ``-v, --verbose``
     Increase verbosity. May be specified multiple times.  See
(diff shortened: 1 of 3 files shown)


gruberchr:
I had a look on your changes and did some experiments with it. Looks good :+1:

The implementation of the verbosity levels is more clear now and `bob status` does, what I would expect. I only have a recommendation for changing the man page for better readability:
```patch
diff --git a/doc/manpages/bob-status.rst b/doc/manpages/bob-status.rst
index dfbe07c..8146a94 100644
--- a/doc/manpages/bob-status.rst
+++ b/doc/manpages/bob-status.rst
@@ -144,13 +144,12 @@ can be enabled. The following levels are available:
 - ``-v`` shows a detailed description of all important flags (``C``, ``E``,
   ``M``, ``S``, ``U`` and ``?``). In particular the modified files and dirty
   commits of a SCM are listed. This level also shows git checkouts that have
-  only unpushed commits that are not related to the checked out branch (``u``
-  flag).
-- ``-vv`` additionally shows the status of SCMs that are not modified, .e.g.
-  without flags or only ``A``, ``N`` and ``O``. If you want to display this
-  information without also enabling the detailed descriptions (see above) use
-  ``--show-clean`` or ``--show-overrides``. This level also shows the full list
-  of commits relatd to the ``u`` flag.
-- ``-vvv`` additionally shows skipped workspaces that do not exist. It also
-  enables the detailed descriptions of the ``A``, ``N`` and ``O`` flags.
-
+  only unpushed commits that are not related to the configured branch (``u``
+  flag), but without detailed description.
+- ``-vv`` shows the full list of commits related to the ``u`` flag. Additionally
+  the status of SCMs that are not modified is shown, .e.g. without flags or only
+  ``A``, ``N`` and ``O``. If you want to display this information without
+  also enabling the detailed descriptions (see above) use ``--show-clean`` or
+  ``--show-overrides``.
+- ``-vvv`` enables the detailed description of the ``A``, ``N`` and ``O``
+  flags. Additionally skipped workspaces that do not exist are shown.
```

jkloetzke:
@gruberchr I took the suggested manpage update as-is. Thank's for the feedback. :+1:

PR #234, opened 2019-03-05
Added git recursive feature for using git-submodules

mahaase:
Added the "git clone --recursive" feature to bob's git plugin. So it's possible to use clone git submodules.

### DIFF
--- a/pym/bob/scm/git.py
+++ b/pym/bob/scm/git.py
@@ -33,6 +33,7 @@ class GitScm(Scm):
         schema.Optional('rev') : str,
         schema.Optional(schema.Regex('^remote-.*')) : str,
         schema.Optional('sslVerify') : bool,
+        schema.Optional('recursive') : bool,
     })
     REMOTE_PREFIX = "remote-"
 
@@ -72,6 +73,10 @@ class GitScm(Scm):
                     raise ParseError("Invalid remote name: " + stripped_key)
                 self.__remotes.update({stripped_key : val})
         self.__sslVerify = spec.get('sslVerify', secureSSL)
+        self.__recursive = ""
+        self.__recursive = spec.get('recursive', self.__recursive)
+        if self.__recursive:
+            self.__recursive = "--recursive"
 
     def getProperties(self):
         properties = super().getProperties()
@@ -87,6 +92,7 @@ class GitScm(Scm):
                     ("refs/heads/" + self.__branch))
             ),
             'sslVerify' : self.__sslVerify,
+            'recursive' : self.__recursive,
         })
         for key, val in self.__remotes.items():
             properties.update({GitScm.REMOTE_PREFIX+key : val})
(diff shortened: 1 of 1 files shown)


jkloetzke:
I'm afraid the topic a a bit more involved than it looks at the first glance. :thinking: I never used submodules but a couple of questions/issues come to my mind:

* what should happen on incremental updates? How is "git pull" supposed to work if there are changes in submodules? What happens with user changes in submodules when the super-module updated the submodule reference (rebase, merge or fail?).
* the case when a specific tag or commit is checked out must be adapted too.
* the Jenkins backend should support the feature too (`asJenkins` method) or fail gracefully if "recursive" is enabled
* I guess the `status` method needs to be updates too to detect changes in submodules
* the audit trail generation certainly needs an update too
* the live build-id support may need to understand submodules too

Given all that a couple of unit- and black box-tests will probably be necessary to prevent regressions in the future. I guess it will need a couple of iterations to get the feature right. Do you think you will have the time to adress the things mentioned above? I know it will be quite a bit of work...

pym/bob/scm/git.py
```diff
@@ -72,6 +73,10 @@ def __init__(self, spec, overrides=[], secureSSL=None):
                     raise ParseError("Invalid remote name: " + stripped_key)
                 self.__remotes.update({stripped_key : val})
         self.__sslVerify = spec.get('sslVerify', secureSSL)
+        self.__recursive = ""
+        self.__recursive = spec.get('recursive', self.__recursive)
```
jkloetzke: Just do a `self.__recursive = spec.get('recursive', False)`. The conversion to the command line switch should be done in the `asScript` method.

mahaase:
> 
> 
> I'm afraid the topic a a bit more involved than it looks at the first glance. 🤔 I never used submodules but a couple of questions/issues come to my mind:
> 
>     * what should happen on incremental updates? How is "git pull" supposed to work if there are changes in submodules? What happens with user changes in submodules when the super-module updated the submodule reference (rebase, merge or fail?).
> 
>     * the case when a specific tag or commit is checked out must be adapted too.
> 
>     * the Jenkins backend should support the feature too (`asJenkins` method) or fail gracefully if "recursive" is enabled
> 
>     * I guess the `status` method needs to be updates too to detect changes in submodules
> 
>     * the audit trail generation certainly needs an update too
> 
>     * the live build-id support may need to understand submodules too
> 
> 
> Given all that a couple of unit- and black box-tests will probably be necessary to prevent regressions in the future. I guess it will need a couple of iterations to get the feature right. Do you think you will have the time to adress the things mentioned above? I know it will be quite a bit of work...

Wow. Thanks for replay. I guess i can't handle this. At least I will not use submodules. That was made for adapting an existing module. All your points say me: don't use submodules and split it in regular way.

jkloetzke:
I have to agree that not using plain git repositories is the better approach. But in principle Bob should be able to support submodules.

Regarding the feature it's not impossible but it will certainly take some time to properly implement the submodule support.  Feel free to close the PR if you don't want to put so much effort into it...

mahaase:
I guess i can't handle this and "using plain git repositories is the better approach".

PR #235, opened 2019-03-06
replaced colons by underlins for qt-creator plugin paths

mahaase:
Replaced colons by underlines for qt-creator plugin paths.
I guess that was just missing. For other paths colons already replaced by underlines.


pym/bob/generators/QtCreatorGenerator.py
```diff
@@ -239,6 +239,7 @@ def qtProjectGenerator(package, argv, extra):
 
     # now go through all checkout dirs to find all header / include files
     for name,path in OrderedDict(sorted(dirs, key=lambda t: t[1])).items():
+        name = name.replace(':', '_')
```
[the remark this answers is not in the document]
jkloetzke: @mahaase Any feedback on this? Have you tried the qt-creator generator on Windows? I don't think it is working without the proposed change...

pym/bob/generators/QtCreatorGenerator.py
```diff
@@ -79,12 +81,16 @@ def addRunSteps(outFile, runTargets):
 
 def addBuildConfig(outFile, num, name, buildArgs, buildMeFile):
     outFile.write('<valuemap type="QVariantMap" key="ProjectExplorer.Target.BuildConfiguration.' + str(num) + '">\n')
-    outFile.write('<value type="QString" key="ProjectExplorer.BuildConfiguration.BuildDirectory">' + os.getcwd() + '</value>\n')
+    outFile.write('<value type="QString" key="ProjectExplorer.BuildConfiguration.BuildDirectory">' + (os.getcwd() if not isWindows() else os.popen('pwd -W').read().replace('\n', '')) + '</value>\n')
```
mahaase: we can't user getcwd for msys2, because of the generated qt-creator files shall be loaded under native windows qt-creator. So the used pathes have to be windows ones.

pym/bob/generators/QtCreatorGenerator.py
```diff
@@ -79,12 +81,16 @@ def addRunSteps(outFile, runTargets):
...
     outFile.write('<valuemap type="QVariantMap" key="ProjectExplorer.BuildConfiguration.BuildStepList.0">\n')
     outFile.write(' <valuemap type="QVariantMap" key="ProjectExplorer.BuildStepList.Step.0">\n')
     outFile.write('  <value type="bool" key="ProjectExplorer.BuildStep.Enabled">true</value>\n')
-    outFile.write('  <value type="QString" key="ProjectExplorer.ProcessStep.Arguments">' + buildArgs + '</value>\n')
-    outFile.write('  <value type="QString" key="ProjectExplorer.ProcessStep.Command">' + buildMeFile + '</value>\n')
+    if isWindows():
+        outFile.write('  <value type="QString" key="ProjectExplorer.ProcessStep.Arguments">-msys2 -defterm -no-start -use-full-path -here -c "PATH=~/bob:/usr/bin /usr/bin/sh ' + buildMeFile + ' -v ' + buildArgs + '"</value>\n')
+        outFile.write('  <value type="QString" key="ProjectExplorer.ProcessStep.Command">' + os.path.normpath(os.path.join(os.getenv('WD').replace('\\', '/'), '..', '..', 'msys2_shell.cmd')) + '</value>\n')
```
mahaase: It is imported to start msys2 by "msys2_shell.cmd" to get WD environment-vaiable.
I got no better solution to get install path of msys2 to execute bob by qt-creator.

pym/bob/generators/QtCreatorGenerator.py
```diff
@@ -200,11 +206,11 @@ def qtProjectGenerator(package, argv, extra):
     getCheckOutDirs(package, excludes, dirs)
     if not projectName:
         # use package name for project name
-        projectName = package.getName()
+        projectName = package.getName().replace('::', '__')
```
mahaase: Replace "::" is important in some case, because of "C:/" shall not match and replaced to "C_/".

pym/bob/generators/QtCreatorGenerator.py
```diff
@@ -292,7 +307,10 @@ def qtProjectGenerator(package, argv, extra):
     name = None
     kits = []
     try:
-        profiles = xml.etree.ElementTree.parse(os.path.join(expanduser('~'), ".config/QtProject/qtcreator/profiles.xml")).getroot()
+        if isWindows():
+            profiles = xml.etree.ElementTree.parse(os.path.join(os.getenv('APPDATA'), "QtProject/qtcreator/profiles.xml")).getroot()
```
mahaase: This is important, that windows can resolve the build presets and kits (kit ids) under windows.

mahaase:
finally there is one issue remaining: inside the qt-creator the project-explorer shows all directories starting at root point (e.g. "C:"), because of the source files not inside the directory of the project files. I have no idea how to solve this. (in linux it is solved by symlinks.)

pym/bob/generators/QtCreatorGenerator.py
jkloetzke: I think we have to live with that. But I would suggest to move the "getcwd/pwd -W" dance into one location and pass the path from there to the functions.

Also I think you can replace the `.replace('\n', '')` with `.strip()` because I guess you just want to remove the trailing \n.

pym/bob/generators/QtCreatorGenerator.py
jkloetzke: Looking at `PATH`: does Bob has to be installed in `~/bob`? Could that be taken from `os.getenv('PATH')` instead?

jkloetzke:
> finally there is one issue remaining: inside the qt-creator the project-explorer shows all directories starting at root point (e.g. "C:"), because of the source files not inside the directory of the project files. I have no idea how to solve this. (in linux it is solved by symlinks.)

I don't have a solution either. Maybe that should be solved in a separate MR when a solution is found...

mahaase:
oO something failed completely on this push...

mahaase:
I close this one. Sorry - new MG is coming... :/

PR #236, opened 2019-03-07
Fix status of unpushed tags

jkloetzke:
Currently all tags that are not on a remote branch are flagged as unpushed commit. This is typically not true and almost always a false positive. Remove tags completely from the game because git does not know anyway which remote has what tags.

### DIFF
--- a/pym/bob/scm/git.py
+++ b/pym/bob/scm/git.py
@@ -283,6 +283,7 @@ class GitScm(Scm):
         status = ScmStatus()
         try:
             onCorrectBranch = False
+            onTag = False
             output = self.callGit(workspacePath, 'ls-remote' ,'--get-url')
             if output != self.__url:
                 status.add(ScmTaint.switched,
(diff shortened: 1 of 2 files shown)


jkloetzke:
@gruberchr Do you have any objections?

gruberchr:
That's right, if a tag is not on a branch, you can not find out, if the commits in the history of that tag are pushed to the remote repository or not. For this you have to check, if the tag has been pushed to remote and that is not possible.
But in my opinion it's not a good idea to simply assume, that all local tags have been pushed to remote and therefore exclude any commits from checking, which are not on a branch and can only be reached from a tag. This is dangerous, since users could not fully rely on `bob status` anymore to really find all unpushed commits in any case. Even if this is _almost always a false positive_, I do not recommend this.
Moreover from my point of view it's even not a good practice to use tags, which are not on a branch. Such tags are not fetched by default with `git fetch` and `git pull` for instance. Such tags have to be fetched always explicitly, whether with the option `--tags` or by explicitly specifying a corresponding refspec.

jkloetzke:
I think you have a point. Maybe excluding all tags is really not a good idea even though such tags are not uncommon in our code base. That's why I noticed the issue in the first place.

But maybe we could still get rid of the issue if Bob just does not clone all tags but only the ones that are on branches. Then such tags would not be there in the first place. Of course if the user requested Bob to check out a specic tag it must still be fetched. But this can then be ignored selectively...

jkloetzke:
Ok, so I adjusted two things:

1. Bob does not clone tags that are not on branches except the recipe requests such a specific tag
2. If such a tag is checked out it is ignored as "unpushed" because it must have come from a remote

This should prevent false positives but still show tags that are not on a remote (branch). Thoughts?

gruberchr:
Looks fine for me

jkloetzke:
Yeah, works here too as expected. Thanks for the feeback.

PR #240, opened 2019-05-04
WIP: archive: added JFrog Artifactory backend

rhubert:
The JFrog artifactory backend uses the dohq-artifcatory python lib and supports the
authentication methods (username+password, API-Key, global configuration file) provided
by this library.

### DIFF
--- a/pym/bob/input.py
+++ b/pym/bob/input.py
@@ -2555,6 +2555,7 @@ class ArchiveValidator:
         artifactoryArchive.update({
             'url' : str,
             schema.Optional('key') : str,
+            schema.Optional('properties') : schema.Schema({ schema.Regex(r'^[A-Za-z_][A-Za-z0-9_]*$') : str }),
             schema.Optional('username') : str,
         })
         self.__backends = {
--- a/pym/bob/input.py
+++ b/pym/bob/input.py
@@ -2526,7 +2526,8 @@ class PackageMatcher:
 
 class ArchiveValidator:
     def __init__(self):
-        self.__validTypes = schema.Schema({'backend': schema.Or('none', 'file', 'http', 'shell', 'azure')},
+        self.__validTypes = schema.Schema({'backend': schema.Or('none',
+            'file', 'http', 'shell', 'azure', 'artifactory')},
             ignore_extra_keys=True)
         baseArchive = {
             'backend' : str,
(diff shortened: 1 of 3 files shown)


rhubert:
This has been tested with artifactory community edition. (https://www.jfrog.com/confluence/display/RTF/Installing+with+Docker)

Other artifactory editions have support for properties. I think a useful feature would be to add metaEnv and/or a list of environment variables (form packageStep?) as property during upload so filtering using properties will be possible.

ATM the `Step` is not provided to `_uploadPackage` so there is no access the the StepEnv during upload. Is there any reason for this or was it just not needed to pass this to `_uploadPackage`?

pym/bob/input.py
```diff
@@ -2550,12 +2551,20 @@ def __init__(self):
             schema.Optional('key') : str,
             schema.Optional('sasToken"') : str,
         })
+        artifactoryArchive = baseArchive.copy()
+        artifactoryArchive.update({
+            'url' : str,
+            schema.Optional('key') : str,
+            schema.Optional('properties') : schema.Schema({ schema.Regex(r'^[A-Za-z_][A-Za-z0-9_]*$') : str }),
```
jkloetzke: Hmm, better leave the `properties` out. The use case is not really clear and it feels wrong to add such a feature just for Artifactory.

rhubert:
Closing this due to missing time (and feedback from the guys who requested this..). My Artifactory-pro evaluation period is also expired.

Thanks for taking time to review this.

jkloetzke:
Nonetheless, many thanks for the effort.

mahaase:
So there is no jfrog artifactoy support yet?

rhubert:
> So there is no jfrog artifactoy support yet?

No

jkloetzke:
> So there is no jfrog artifactoy support yet?

Not directly. You can still use the generic http upload if Artifactory is configured to accept anonymous uploads. But there will be no metadata associated with the artifacts.

mahaase:
okay, thanks, just good to know. so i will not advertise this feature :)

mahaase:
hey, i tested this pull-request. it is still working good already. 👍

there is an option to test and develop with jfrog artifactory for free: https://jfrog.com/start-free/
it is a jfrog cloud test artifactory.

but until now, i don't see why it could be a good idea to use this. it isn't an interface between conan and bob.
the meta-data stuff hasn't a good usability (e.g. for searching uploaded artifacts).

currently it is also hard to bring the `BOB_RECIPE_NAME` as property/meta data into the artifactory, because this variable isn't available in package step (`packageVars`) automatically.

rhubert:
Hi,
wow - I totally forgot about this when I was asked about artifactory support some days ago 🙄 ...

```quote
but until now, i don't see why it could be a good idea to use this. it isn't an interface between conan and bob.
```
No, and it will never be.
Do you have a feeling about how you want to use artifactory with bob?

I think the main advantage of using artifactory would be to add some metadata to the artifacts and tag them as nightly, release artifacts. If the same artifact and is used for different tagged builds the tag needs to be updated,... This way it's easy to clean the cache.

PR #241, opened 2019-05-06
url: Use wget as fallback in case curl isn't available

JoeKar:
This fallback has some limitations:
- wget doesn't support as much protocols as curl, but for the usual
HTTP(S) and FTP it does the job
- the time-stamp mechanism (-N) can't be used as with curl, because wget
doesn't support it together with own outputfile names (-O)

The motivation of this change was to lower the dependencies for simple
tasks.

Maybe there is a simple solution for the time-stamp approach.

### DIFF
--- a/pym/bob/scm/url.py
+++ b/pym/bob/scm/url.py
@@ -86,8 +86,11 @@ class UrlScm(Scm):
         return ret
 
     def asScript(self):
-        options = "-sSgLf"
-        if not self.__sslVerify: options += "k"
+        curlOptions="-sSgLf"
+        wgetOptions="-q --no-glob"
+        if not self.__sslVerify:
+            curlOptions+="k"
+            wgetOptions+=" --no-check-certificate"
         ret = ""
         if self.__url[0] == '/':
             # Local files: copy only if newer (u), target never is a directory (T)
@@ -98,24 +101,34 @@ cd {DIR}
 cp -uT {URL} {FILE}
 """
         else:
+            # do command logic inside the shell script to leave it to the backend
             tpl = """
 {HEADER}
 mkdir -p {DIR}
 cd {DIR}
 if [ -e {FILE} ] ; then
-    curl {OPTIONS} -o {FILE} -z {FILE} {URL}
+    if [ -x "$(command -v curl)" ] ; then
+        curl {CURL_OPTIONS} -o {FILE} -z {FILE} {URL}
+    fi
 else
     (
         F=$(mktemp)
         trap 'rm -f $F' EXIT
         set -e
-        curl {OPTIONS} -o $F {URL}
+        if [ -x "$(command -v curl)" ] ; then
+            curl {CURL_OPTIONS} -o $F {URL}
+        elif [ -x "$(command -v wget)" ] ; then
+            wget {WGET_OPTIONS} -O $F {URL}
+        else
+            >&2 echo "\nERROR: Don't know how to download. Please install 'curl'!\n"
+            exit 1
+        fi
         mv $F {FILE}
     )
 fi
 """
         ret = tpl.format(HEADER=super().asScript(), DIR=quote(self.__dir), URL=quote(self.__url),
-           FILE=quote(self.__fn), OPTIONS=options)
+           FILE=quote(self.__fn), CURL_OPTIONS=curlOptions, WGET_OPTIONS=wgetOptions)
 
         if self.__digestSha1:
             ret += "echo {DIGEST}\ \ {FILE} | sha1sum -c\n".format(DIGEST=self.__digestSha1, FILE=self.__fn)


pym/bob/scm/url.py
```diff
@@ -98,24 +98,45 @@ def asScript(self):
 cp -uT {URL} {FILE}
 """
         else:
+            tool = "curl"
+            options = ""
+            useTimeStamping = True
+            if which(tool) is not None:
```
jkloetzke: The whole descision must be done during runtime in the bash script. The python code runs where the recipes are parsed which is not necessarily the same machine as where they are executed in case of the Jenkins backend.

pym/bob/scm/url.py
```diff
@@ -98,24 +98,45 @@ def asScript(self):
...
+                    raise ParseError("Fallback to 'wget' impossible due to protocol limitations. Please install 'curl'!")
+            else:
+                raise ParseError("Don't know how to download. Please install 'curl'!")
             tpl = """
 {HEADER}
 mkdir -p {DIR}
 cd {DIR}
-if [ -e {FILE} ] ; then
-    curl {OPTIONS} -o {FILE} -z {FILE} {URL}
+if [ -e {FILE} -a {USETIMESTAMPING} = "True" ] ; then
```
jkloetzke: Now the code will always download the file if wget is used. This will typically be unncessary in case the checkout step is not determinstic as a whole. Instead I would suggest to download only if the destination does not exist in case of wget.

doc/tutorial/getting_started.rst
```diff
@@ -11,12 +11,18 @@ Python modules that are not in the standard library:
...
   that comes with your distribution (e.g. python3-pyparsing on Debian).
 * gcc
-* `python3-sphinx` Only needed for generating man pages.
+* `python3-sphinx`_. Only needed for generating man pages.
+
+Apart from the tutorial additional dependencies could arise, e.g.:
+
+* curl as the default URL SCM downloader (a limited fallback to wget is possible)
+* extractors based on the supported extensions (e.g. xz, 7z, ...)
+* source code management handlers (e.g. git, svn, ...)
```
jkloetzke: The generic documentation should be a separate commit. Except the "(a limited fallback to wget is possible)" it's completely unrelated to the change and usefull on it's own.

pym/bob/scm/url.py
JoeKar: The backend stuff was totally out of scope to me, sorry.  I will update asap.

pym/bob/scm/url.py
JoeKar: You're right, will be fixed as well.

doc/tutorial/getting_started.rst
JoeKar: Ok, I will prepare.
How about moving the whole "Getting Started" chapter into the common "Bob User Manual" in between of "Introduction" and "Concepts"? This is more or less the usual place for build-/run time dependencies and installation instructions. I wouldn't search in the tutorial in the first place.

rhubert:
What about using `bob _download` as curl replacement?

pym/bob/scm/url.py
```diff
@@ -98,24 +96,49 @@ def asScript(self):
...
         else:
+            # do command logic inside the shell script to leave it to the backend
             tpl = """
 {HEADER}
 mkdir -p {DIR}
 cd {DIR}
 if [ -e {FILE} ] ; then
-    curl {OPTIONS} -o {FILE} -z {FILE} {URL}
+    if [ -x "$(command -v curl)" ] ; then
+        OPTIONS="-sSgLf"
```
jkloetzke: If you set variables you will have to put that part in a sub-shell to prevent leaking the variable in the `checkoutScript` that is possibly concatenated. But as the list of options is fixed I would rather pass it directly into the script, e.g. as `{CURL_OPTIONS}` and `{WGET_OPTIONS}`.

pym/bob/scm/url.py
```diff
@@ -98,24 +96,49 @@ def asScript(self):
...
+        elif [ -x "$(command -v wget)" ] ; then
+            OPTIONS="-q --no-glob"
+            if [ {SKIPCERTCHECK} = True ] ; then
+                OPTIONS+=" --no-check-certificate"
+            fi
+            wget $OPTIONS -O $F {URL}
+        else
+            >&2 echo "\nERROR: Don't know how to download. Please install 'curl'!\n"
+            # cleanup
+            rm -f $F
```
jkloetzke: This is unneeded. The `trap ...` at the start of the sub-shell already takes care of this.

pym/bob/scm/url.py
```diff
@@ -98,24 +96,49 @@ def asScript(self):
...
+            # exit successful to prevent the error print of the script
+            exit 0
```
jkloetzke: That should be `exit 1`! Otherwise the download will silently fail and break the build because nothing was downloaded as expected.

JoeKar: I need to refer to the last two comments...
the #cleanup was necessary in case the exit with '0' has been used since this doesn't involve the trap and the reason was more or less written with the comment (I tried it with '1' before) above the exit.

I thought...
`ERROR: Don't know how to download. Please install 'curl'!

sha1sum: pkg-config-0.29.1.tar.gz: No such file or directory
sha1sum: WARNING: 1 listed file could not be read
Step failed with return status 1; Command: sha1sum -c`
...was sufficient, but you're right, it now complains about the step of checksum verification, which isn't really failing.

Ok, let's summarize using exit with '1' is the correct approach to let the current step fail...but even then I suggest to introduce the error echo. Because this directly tells what was going wrong including a sufficient proposal without reading the printed snippet of the script.

PR #242, opened 2019-05-07
ADD qtcreator support for windows by MSYS2

mahaase:
update of MR: https://github.com/BobBuildTool/bob/pull/235 

- I got no better solution that using pwd -W msys2 tool for getting the cwd/pwd in windows style.
- It is imported to start msys2 by "msys2_shell.cmd" to get WD environment-vaiable. I got no better solution to get install path of msys2 to execute bob by qt-creator.
- Replace "::" is important in some case, because of "C:/" shall not match and replaced to "C_/".

TODO

- finally there is one issue remaining: inside the qt-creator the project-explorer shows all directories starting at root point (e.g. "C:"), because of the source files not inside the directory of the project files. I have no idea how to solve this. (in linux it is solved by symlinks.)

### DIFF
--- a/pym/bob/generators/QtCreatorGenerator.py
+++ b/pym/bob/generators/QtCreatorGenerator.py
@@ -11,14 +11,24 @@ import os
 import shutil
 import stat
 import xml.etree.ElementTree
+import shutil
+import subprocess
 from os.path import expanduser
 from os.path import join
-from bob.utils import removePath
+from bob.utils import removePath, isWindows
 from bob.errors import BuildError
 from bob.utils import summonMagic, hashFile
 from collections import OrderedDict, namedtuple
 from bob.tty import colorize
 from pipes import quote
+
+# helper to get linux or windows (MSYS2 support) cwd
+pwd = None
+def cwd():
+    global pwd
+    if pwd == None and isWindows():
+        pwd = os.popen('pwd -W').read().strip()
+    return os.getcwd() if not isWindows() else pwd
 
 # scan package recursivelely with its dependencies and build a list of checkout dirs
 def getCheckOutDirs(package, excludes, dirs):
@@ -79,12 +89,18 @@ def addRunSteps(outFile, runTargets):
 
 def addBuildConfig(outFile, num, name, buildArgs, buildMeFile):
     outFile.write('<valuemap type="QVariantMap" key="ProjectExplorer.Target.BuildConfiguration.' + str(num) + '">\n')
-    outFile.write('<value type="QString" key="ProjectExplorer.BuildConfiguration.BuildDirectory">' + os.getcwd() + '</value>\n')
+    outFile.write('<value type="QString" key="ProjectExplorer.BuildConfiguration.BuildDirectory">' + cwd() + '</value>\n')
     outFile.write('<valuemap type="QVariantMap" key="ProjectExplorer.BuildConfiguration.BuildStepList.0">\n')
     outFile.write(' <valuemap type="QVariantMap" key="ProjectExplorer.BuildStepList.Step.0">\n')
     outFile.write('  <value type="bool" key="ProjectExplorer.BuildStep.Enabled">true</value>\n')
-    outFile.write('  <value type="QString" key="ProjectExplorer.ProcessStep.Arguments">' + buildArgs + '</value>\n')
-    outFile.write('  <value type="QString" key="ProjectExplorer.ProcessStep.Command">' + buildMeFile + '</value>\n')
+    if isWindows():
+        outFile.write('  <value type="QString" key="ProjectExplorer.ProcessStep.Arguments">-msys2 -defterm -no-start -use-full-path -here -c "PATH=' + os.getenv('PATH') + ' sh ' + buildMeFile + ' -v ' + buildArgs + '"</value>\n')
+        if os.getenv('WD') is None:
+            raise BuildError("Cannot create QtCreator project for Windows! MSYS2 must be started by msys2_shell.cmd script!")
+        outFile.write('  <value type="QString" key="ProjectExplorer.ProcessStep.Command">' + os.path.normpath(os.path.join(os.getenv('WD').replace('\\', '/'), '..', '..', 'msys2_shell.cmd')) + '</value>\n')
+    else:
+        outFile.write('  <value type="QString" key="ProjectExplorer.ProcessStep.Arguments">' + buildArgs + '</value>\n')
+        outFile.write('  <value type="QString" key="ProjectExplorer.ProcessStep.Command">' + buildMeFile + '</value>\n')
     outFile.write('  <value type="QString" key="ProjectExplorer.ProcessStep.WorkingDirectory">%{buildDir}</value>\n')
     outFile.write('  <value type="QString" key="ProjectExplorer.ProjectConfiguration.DefaultDisplayName">' + name + '</value>\n')
     outFile.write('  <value type="QString" key="ProjectExplorer.ProjectConfiguration.DisplayName"></value>\n')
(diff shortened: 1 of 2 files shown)


pym/bob/generators/QtCreatorGenerator.py
```diff
@@ -79,12 +81,16 @@ def addRunSteps(outFile, runTargets):
...
+    outFile.write('<value type="QString" key="ProjectExplorer.BuildConfiguration.BuildDirectory">' + (os.getcwd() if not isWindows() else os.popen('pwd -W').read().strip()) + '</value>\n')
     outFile.write('<valuemap type="QVariantMap" key="ProjectExplorer.BuildConfiguration.BuildStepList.0">\n')
     outFile.write(' <valuemap type="QVariantMap" key="ProjectExplorer.BuildStepList.Step.0">\n')
     outFile.write('  <value type="bool" key="ProjectExplorer.BuildStep.Enabled">true</value>\n')
-    outFile.write('  <value type="QString" key="ProjectExplorer.ProcessStep.Arguments">' + buildArgs + '</value>\n')
-    outFile.write('  <value type="QString" key="ProjectExplorer.ProcessStep.Command">' + buildMeFile + '</value>\n')
+    if isWindows():
+        outFile.write('  <value type="QString" key="ProjectExplorer.ProcessStep.Arguments">-msys2 -defterm -no-start -use-full-path -here -c "PATH=' + os.getenv('PATH') + ' sh ' + buildMeFile + ' -v ' + buildArgs + '"</value>\n')
+        outFile.write('  <value type="QString" key="ProjectExplorer.ProcessStep.Command">' + os.path.normpath(os.path.join(os.getenv('WD').replace('\\', '/'), '..', '..', 'msys2_shell.cmd')) + '</value>\n')
```
jkloetzke: That line feels magic. What's in `WD` and is the relative path to `msys2_shell.cmd` always the same? I'm just curious...

pym/bob/generators/QtCreatorGenerator.py
```diff
@@ -98,7 +104,7 @@ def addBuildConfig(outFile, num, name, buildArgs, buildMeFile):
     outFile.write('<value type="int" key="ProjectExplorer.BuildConfiguration.BuildStepListCount">1</value>\n')
     outFile.write('<value type="bool" key="ProjectExplorer.BuildConfiguration.ClearSystemEnvironment">false</value>\n')
     outFile.write('<valuelist type="QVariantList" key="ProjectExplorer.BuildConfiguration.UserEnvironmentChanges">\n')
-    outFile.write(' <value type="QString">' + "PATH=" + os.environ['PATH'] + '</value>\n')
+    outFile.write(' <value type="QString">' + 'PATH=' + os.environ['PATH'] + '</value>\n')
```
jkloetzke: Looks like an unneeded change.

pym/bob/generators/QtCreatorGenerator.py
```diff
@@ -200,11 +206,11 @@ def qtProjectGenerator(package, argv, extra):
     getCheckOutDirs(package, excludes, dirs)
     if not projectName:
         # use package name for project name
-        projectName = package.getName()
+        projectName = package.getName().replace('::', '__')
     if not destination:
         # use package name for project directory
-        destination = os.path.join(os.getcwd(), "projects", projectName)
-        destination = destination.replace(':', '_')
+        destination = os.path.join(os.getcwd() if not isWindows() else os.popen('pwd -W').read().strip(), "projects", projectName)
```
jkloetzke: Can you replace all `os.getcwd() if not isWindows() else os.popen('pwd -W').read().strip()` instances by invoking a helper?, e.g.:

```python
def cwd():
    return os.getcwd() if not isWindows() else os.popen('pwd -W').read().strip()

...

        destination = os.path.join(cwd(), "projects", projectName)
```

That'll remove the code duplication and make it more readable.

pym/bob/generators/QtCreatorGenerator.py
mahaase: that depends on this: https://github.com/msys2/MSYS2-packages/blob/master/filesystem/msys2_shell.cmd
It is important to start MSYS2 with this msys2_shell.cmd. I don't know a better solution.
I should log out an error, if WD is not set, with the information, that the user has to start MSYS2 by this script?!

pym/bob/generators/QtCreatorGenerator.py
mahaase: Correct, I will revert this with next commit.

mahaase:
We should write some documentation for this. The idea is, that we use natively on windows, incl. a native QtCreator installation (using a MSYS2 QtCreator installation isn't supported). So this plugin uses windows pathes for the project files, but linux/msys2 pathes for bob stuff by using the "msys2_shell.cmd" to start bob inside MSYS2.

pym/bob/generators/QtCreatorGenerator.py
jkloetzke: I think you must test it in advance. Because `os.getenv()` returns `None` if the variable is not set and then the code will crash. Better raise a `BuildError` with a proper error message before.

pym/bob/generators/QtCreatorGenerator.py
```diff
@@ -239,18 +249,28 @@ def qtProjectGenerator(package, argv, extra):
 
     # now go through all checkout dirs to find all header / include files
     for name,path in OrderedDict(sorted(dirs, key=lambda t: t[1])).items():
-        newPath = os.path.join(symlinkDir, name)
-        if not args.update: os.symlink(os.path.join(os.getcwd(), path), newPath)
+        name = name.replace('::', '__')
```
jkloetzke: Please move the replace in the `isWindows` path. The substitution is not needed on Linux and there we can show the correct package name which looks better.

jkloetzke:
> We should write some documentation for this. The idea is, that we use natively on windows, incl. a native QtCreator installation (using a MSYS2 QtCreator installation isn't supported). So this plugin uses windows pathes for the project files, but linux/msys2 pathes for bob stuff by using the "msys2_shell.cmd" to start bob inside MSYS2.

That would be indeed very good. A good place would be the bob-project manpage IMHO.

pym/bob/generators/QtCreatorGenerator.py
mahaase: done.

pym/bob/generators/QtCreatorGenerator.py
```diff
@@ -239,18 +251,28 @@ def qtProjectGenerator(package, argv, extra):
...
             if (((os.path.sep + '.git' + os.path.sep) in root) or
                 ((os.path.sep + '.svn' + os.path.sep) in root)):
                 continue
             for filename in filenames:
                 if source.match(filename) or cmake.match(filename) or filename == 'CMakeLists.txt':
-                    sList.append(os.path.join(os.getcwd(), os.path.join(root,filename)))
+                    if isWindows():
+                        sList.append(os.path.join(root,filename))
+                    else:
+                        sList.append(os.path.join(cwd(), os.path.join(root,filename)))
```
jkloetzke: Oh, I guess this will be slow on Windows with many files. What about calling `cwd()` once at the start of `qtProjectGenerator()` and use a variable to store the value?

mahaase: I am not a pyhton geek, would that be okay @jkloetzke?

`
# helper to get linux or windows (MSYS2 support) cwd
+pwd = None
 def cwd():
-    return os.getcwd() if not isWindows() else os.popen('pwd -W').read().strip()
+    global pwd
+    if pwd == None and isWindows():
+        pwd = os.popen('pwd -W').read().strip()
+    return os.getcwd() if not isWindows() else pwd
`

mahaase:
> BTW, could also please restructure the commits and align the commit messages with the usual style of the Bob repository? Just take a look at the history for some examples. I suggest to make two commits: one for the actual generator rework and one for the documentation update.

Okay, I will rebase the commit-messages in the end!

jkloetzke:
Thanks for contributing. :+1:

mahaase:
it's a pleasure :)

btw: I have a patch for fixing the path issue inside qtcreator's generic-project explorer:

https://bugreports.qt.io/browse/QTCREATORBUG-19454
but no idea, how to integrate this correctly, the patch is just the straight variant, working for usage the bob-project qt-creator plugin, **if u use it under windows platform**!

PR #253, opened 2019-06-02
Visual Studio 2019 generator

### LINKED ISSUE #248: bob project generator for Visual Studio
Hi,

this "issue" is an improvement ticket. I am searching for good ideas how to adapt Visual Studio to bob project to generate a **generic** Visual Studio project like the QtCreator plugin. The best would be, if Visual Studio also uses bob to build the software, debugging should be possible and build errors should be parsed the the source code.

I have no idea how to realize this - especially without using CMake generator.

You are welcome for brainstorming :)

jkloetzke:
The first implementation that works on MSYS2. Full support for VS would require that Bob runs natively on Windows which is not possible (yet).

Fixes #248.

### DIFF
--- a/pym/bob/generators/VisualStudio.py
+++ b/pym/bob/generators/VisualStudio.py
@@ -1,0 +1,330 @@
+# Bob build tool
+# Copyright (C) 2019  Jan Klötzke
+#
+# SPDX-License-Identifier: GPL-3.0-or-later
+
+from .common import CommonIDEGenerator
+from pathlib import Path, PureWindowsPath
+from pipes import quote
+from uuid import UUID
+from uuid import uuid4 as randomUuid
+from uuid import uuid5 as sha1NsUuid
+from xml.etree import ElementTree
+import os
+import sys
+
+SOLUTION_TEMPLATE = """\
+
+Microsoft Visual Studio Solution File, Format Version 12.00
+# Visual Studio Version 16
+VisualStudioVersion = 16.0.28803.352
+MinimumVisualStudioVersion = 10.0.40219.1
+{PROJECTS_LIST}
+Global
+	GlobalSection(SolutionConfigurationPlatforms) = preSolution
+		Build|x86 = Build|x86
+		Checkout+Build|x86 = Checkout+Build|x86
+	EndGlobalSection
+	GlobalSection(ProjectConfigurationPlatforms) = postSolution
+{PROJECTS_CFG}
+	EndGlobalSection
+	GlobalSection(SolutionProperties) = preSolution
+		HideSolutionNode = FALSE
+	EndGlobalSection
+	GlobalSection(ExtensibilityGlobals) = postSolution
+		SolutionGuid = {{{SOLUTION_GUID}}}
+	EndGlobalSection
+EndGlobal
+"""
+
+SOLUTION_PROJECT_TEMPLATE = """\
+Project("{{8BC9CEB8-8B4A-11D0-8D11-00A0C91BC942}}") = "{NAME}", "{NAME}\\{NAME}.vcxproj", "{{{GUID}}}"
+EndProject"""
+
+PROJECT_TEMPLATE = """\
+<?xml version="1.0" encoding="utf-8"?>
+<Project DefaultTargets="Build" xmlns="http://schemas.microsoft.com/developer/msbuild/2003">
+  <ItemGroup Label="ProjectConfigurations">
+    <ProjectConfiguration Include="Build|Win32">
+      <Configuration>Build</Configuration>
+      <Platform>Win32</Platform>
+    </ProjectConfiguration>
+    <ProjectConfiguration Include="Checkout+Build|Win32">
+      <Configuration>Checkout+Build</Configuration>
+      <Platform>Win32</Platform>
+    </ProjectConfiguration>
+  </ItemGroup>
+  <PropertyGroup Label="Globals">
+    <VCProjectVersion>16.0</VCProjectVersion>
+    <ProjectGuid>{{{PROJECT_GUID}}}</ProjectGuid>
+    <Keyword>Win32Proj</Keyword>
+  </PropertyGroup>
+  <Import Project="$(VCTargetsPath)\Microsoft.Cpp.Default.props" />
+  <PropertyGroup Condition="'$(Configuration)|$(Platform)'=='Build|Win32'" Label="Configuration">
+    <ConfigurationType>Makefile</ConfigurationType>
+    <UseDebugLibraries>true</UseDebugLibraries>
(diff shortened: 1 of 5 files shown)


pym/bob/generators/VisualStudio.py
```diff
@@ -0,0 +1,320 @@
...
+            for i in self.sources: ElementTree.SubElement(g, "ClCompile", {"Include" : str(i)})
+            items.append(ElementTree.tostring(g, encoding="unicode"))
+        # TODO: self.resources
+        if self.dependencies:
+            g = ElementTree.Element("ItemGroup")
+            incPaths = []
+            for i in sorted(self.dependencies):
+                dep = ElementTree.SubElement(g, "ProjectReference",
+                        {"Include" : "..\\{NAME}\\{NAME}.vcxproj".format(NAME=i)})
+                ElementTree.SubElement(dep, "Project").text = "{" + str(projects[i].uuid) + "}"
```
mahaase: If there is a module, which is splitted into multiple packages, the projects[i] will crash.
As workaround, I surrounded this lines by try-except and everything is fine.

pym/bob/generators/VisualStudio.py
```diff
@@ -0,0 +1,320 @@
...
+            ElementTree.SubElement(n, "NMakeBuildCommandLine").text = cmd + " -b " + self.packagePath
+            ElementTree.SubElement(n, "NMakeReBuildCommandLine").text = cmd + " -b -f --clean " + self.packagePath
+            # TODO: NMakeCleanCommandLine == "rm -rf dev/{build,dist}"?
+            items.append(ElementTree.tostring(n, encoding="unicode"))
+
+            n = ElementTree.Element("PropertyGroup", {"Condition" : "'$(Configuration)|$(Platform)'=='Checkout+Build|Win32'"})
+            ElementTree.SubElement(n, "NMakeOutput")
+            ElementTree.SubElement(n, "NMakeIncludeSearchPath").text = ";".join(incPaths)
+            ElementTree.SubElement(n, "NMakeBuildCommandLine").text = cmd + " " + self.packagePath
+            ElementTree.SubElement(n, "NMakeReBuildCommandLine").text = cmd + " -f --clean " + self.packagePath
```
mahaase: Here I added -v, I guess in QtCreator plugin -v is also default?
But this is something I can do: bring all build-modes from Qt-Creator plugin to here.

pym/bob/generators/VisualStudio.py
```diff
@@ -0,0 +1,320 @@
...
+            g = ElementTree.Element("ItemGroup")
+            for i in self.sources: ElementTree.SubElement(g, "ClCompile", {"Include" : str(i)})
+            items.append(ElementTree.tostring(g, encoding="unicode"))
+        # TODO: self.resources
+        if self.dependencies:
+            g = ElementTree.Element("ItemGroup")
+            incPaths = []
+            for i in sorted(self.dependencies):
+                dep = ElementTree.SubElement(g, "ProjectReference",
+                        {"Include" : "..\\{NAME}\\{NAME}.vcxproj".format(NAME=i)})
```
mahaase: Windows has trouble with "::" in filesystem, I replaces here by "__".

pym/bob/generators/VisualStudio.py
```diff
@@ -0,0 +1,320 @@
...
+        buildMe = os.path.join(self.destination, "buildme.sh")
+        buildMeCmd = baseBuildMe + " " + buildMe
+        if not self.args.update:
+            self.updateFile(buildMe, self.__generateBuildme(extra))
+
+        solutionProjectList = []
+        solutionProjectConfigs = []
+        for name,project in projects.items():
+            p = os.path.join(self.destination, name)
+            os.makedirs(p, exist_ok=True)
```
mahaase: Windows has trouble with "::" in filesystem, I replaces here by "__".

pym/bob/generators/VisualStudio.py
```diff
@@ -0,0 +1,320 @@
...
+            self.updateFile(os.path.join(p, name+".vcxproj"), project.generateProject(projects, buildMeCmd),
+                    encoding="utf-8", newline='\r\n')
+            self.updateFile(os.path.join(p, name+".vcxproj.filters"), project.generateFilters(),
+                    encoding="utf-8", newline='\r\n')
+
+            solutionProjectList.append(SOLUTION_PROJECT_TEMPLATE.format(NAME=name, GUID=str(project.uuid).upper()))
+            solutionProjectConfigs.append("\t\t{{{GUID}}}.Build|x86.ActiveCfg = Build|Win32".format(GUID=str(project.uuid).upper()))
```
mahaase: Just for the root-project we have to add "\t\t{{{GUID}}}.Build|x86.Build.0", so the "Projektmappe" get a default project, which will be built by clicking <build "Projektmappe">.

jkloetzke:
Most stuff should be addressed. There are still many TODOs and FIXMEs, though...

pym/bob/generators/VisualStudio.py
mahaase: fixed!

pym/bob/generators/VisualStudio.py
mahaase: fixed.

pym/bob/generators/VisualStudio.py
mahaase: fixed.

pym/bob/generators/VisualStudio.py
mahaase: fixed

pym/bob/generators/VisualStudio.py
mahaase: also fixed.

mahaase:
I tested the new fixes - well done!! No problems with "::" anymore. Default project can be built by "Projektmappe erstellen". No crashes while creating the vs project. 👍 

Also the exclude feature works well - important in VS for projects with boost ;-)

The only thing remaining is the project tree, where the source code is just splitted into source and header files.

jkloetzke:
> I tested the new fixes - well done!! No problems with "::" anymore. Default project can be built by "Projektmappe erstellen". No crashes while creating the vs project. 👍
> Also the exclude feature works well - important in VS for projects with boost ;-)

Nice to hear.

> The only thing remaining is the project tree, where the source code is just splitted into source and header files.

That should be fixed too. Hopefully I didn't break anything. 😉 

> btw: vs2019, it is also working with vs2017 so far.

Could be but I guess there may be subtle differences in the generated project. At least CMake has distinct generators for them. As I don't have vs2017 to test I would rather not advertise it officially. At least the toolset has a different version...

PR #258, opened 2019-06-18
Native windows support

jkloetzke:
This is a heavily work-in-progress PR that I still wanted to share early to get some feedback. The branch enables the execution of Bob on native Windows (no MSYS!). Bob gained another scripting language (namely Powershell) that can be used in parallel to Bash. When run natively on Windows only Powershell will work, though.

### What is needed?

1. Install Windows Python and make sure it's added to PATH
2. clone Bob

That's it. There is a `bob.cmd` file so that the Windows shell understands the "bob" command.

### What is working so far?

Basic git, url SCM and scripting is working. The following kind-of-builds given that `git` was added to `PATH` and Visual Studio is installed.

https://gist.github.com/jkloetzke/d760c5800e97cbbe9486b2152ffdb164

The CMakeFile.txt enables Fortran that fails. I've yet to find a simpler example.

You can also create a Visual Studio project via

    bob project root vs2019

that works even with debugging the application!

Recipes can define dedicated scripts for each language (e.g. `buildScriptBash` and `buildScriptPwsh`) that is selected depending on the Platform Bob runs on. 

### What's to be done?

A lot of things. This will take a couple of weeks to get into shape. What I would like to know if it's worth the effort? Is PowerShell a good idea (cmd.exe scripts are unusable IMHO)? What else is needed to get a proper Windows support?

**ATTENTION: the branch will be rebased from time to time!**


jkloetzke:
@mahaase @kolewu @ThomasFeher Would be nice if you'd have a look on it 😉

gruberchr:
What about using MINGW32/64 Bash instead of Powershell?

- it supports several UNIX commands in a native Windows environment, so scripts might not have to be changed
- Bob runs in Windows Python
- native Windows path names are used
- native CMake can be used
- you can create native Visual Studio projects

kolewu:
Why not use python as scripting language? This could be useful for all platforms and should be available wherever there is a compiler.

And yes, batch-scripting is unspeakably bad to write and totally useless for someone who knows real scripting.

kolewu:
Yes, batch-scripting is unspeakably bad to write and totally useless for someone who knows real scripting.

mahaase:
I also would prefer the bash-scripts. I have completely lo experiences to powershell scripting. And I would like it, if my bash-scripts, which already there, would be supported also for the native-windows-bob.
But generally the idea, using python also for the scripts, is in my opinion the cleanest way.

jkloetzke:
> What about using MINGW32/64 Bash instead of Powershell?

Well, that's the current solution you have I guess. Because the python interpreter is not the problem. It's the external tools that are called that need to get native paths but the MINGW bash is still a unix tool with unix paths. I briefly experimented with that approach and found it to be not really usable. When I tried to call the Microsoft compiler with some paths I got mysterious error messages.

> Why not use python as scripting language? This could be useful for all platforms and should be available wherever there is a compiler.

I really though hard about this approach. But in the end Bob is there to call external tools. And there Python does not really shine. While a `subprocess.check_call(['cmake', ...])` is ok I have some other typical problems in the recipes that are not so nice to write:

* `os.path.join("a", "b", ...)` for a portable python script?
* file globbing (`os.walk`?)
* manipulate/pass environment variables
* pushd/popd

That's purely subjective but my feeling is that the Python scripts won't be very readable. And scripts that support both Linux and Windows will deteriorate into a large `if windows else linux` anyway. Maybe it really depends on the use case. My feeling was that for Windows-Only projects PowerShell might be a better fit. But once multiple scripting languages can be supported adding another one should be relatively easy.

jkloetzke:
> I also would prefer the bash-scripts. I have completely lo experiences to powershell scripting. And I would like it, if my bash-scripts, which already there, would be supported also for the native-windows-bob.

The problem is there is no bash AFAIK that works with Windows native paths. The unix tools that are called from the bash scripts always work on unix paths. That's why I think it is necessary to use something else.

> But generally the idea, using python also for the scripts, is in my opinion the cleanest way.

It would be good to know how hard the conversion really is for your use case and if the scripts are maintainable in the end. You could rewrite your recipes to just call python directly with an inline script. That should be close enough to try it out.

jkloetzke:
Thinking more about the remarks I have the feeling that Bob should get mostly independent of the scripting language that is used in the recipes instead of requiring specific ones. Currently the problem is that Bob generates itself scripts that are concatenated with the user supplied scripts. This ties Bob to the scripting language of the recipes which is actually not needed. So I'm thinking currently in the following direction:

* move all Bob generated scripts to separately executed portable python scripts
* only execute {checkout,build,package,fingerprint}Script with the chosen scripting language

In the end Bob only needs to know very little about the scripting language in the recipes: where is the interpreter, how to call it and some tiny header that is added to each script (error handling, setting special BOB_ variables). That way almost any language can be supported with minimal effort.

mahaase:
Okay. I talked to a college. He said PowerShell (OO scripting language) is very powerful and usable for windows. And I had a look into jenkins today - jenkins also supports PowerShell in a plugin. So it would also be possible to run a native windows jenkins. That would be very nice. 

Learning a new script-language maybe is the smallest pain :)

jkloetzke:
Ok, this is now almost finished. Just the documentation and some minor problems are remaining. :wink: 

The following things are now working:
* Run Bob native on Windows with PowerShell scripts
* Run Bob native on Windows with MSYS2 bash. This only works if `bash.exe` is in `%PATH%` at the *end*. Bob must be run with a native Python installation.
* Run Bob on Windows in MSYS2
* Use PowerShell on Linux as scripting language

Any feedback is welcome.

mahaase:
Thanks!!! We will test this these days!

sbixl:
Hi together and thanks for this very nice feature. We will test this in the next two weeks too.

I'm not sure if it's a problem or if I used it wrong but if I try to checkout a URL from a network drive under windows like e.g. "//dwxyz/my-pkg.zip" if fails. A similar issue happens if I use "d:/my-pkg.zip" where bob complains about "Unsupported URL scheme: d". I have already checked the invoke function in file [url.py](https://github.com/jkloetzke/bob/blob/native-windows/pym/bob/scm/url.py) and it looks like it is not considered / does not fit to the concept?

I would also share some hints about improving build performance under Windows 10 (tested with the initial commit of this PR from June 2019 and F-Secure Internet Security):

- First of all you should deactivate all antivirus scans on a specific directory in which bob will build your artifacts/projects. That alone halves the build time (at least in our environment). 
- Don't build in your user directory under windows because the internal windows indexer mechanism is scanning this directory. You can of course deactivate indexing [manage-windows-10-search-indexing](https://www.ghacks.net/2017/08/10/manage-windows-10-search-indexing/) if required.
- Furthermore some Antivirus-programs seems to sporadic slow down the compiler access if the compiler is installed on the local filesystem (e.g. IAR-Compiler). If you use packed pre-compiled toolchain binaries like gcc this effect does not occur.

jkloetzke:
> I'm not sure if it's a problem or if I used it wrong but if I try to checkout a URL from a network drive under windows like e.g. "//dwxyz/my-pkg.zip" if fails. A similar issue happens if I use "d:/my-pkg.zip" where bob complains about "Unsupported URL scheme: d". I have already checked the invoke function in file [url.py](https://github.com/jkloetzke/bob/blob/native-windows/pym/bob/scm/url.py) and it looks like it is not considered / does not fit to the concept?

Currently Windows paths are not handled and there is only a shortcut for Unix paths. Using `file://...` doesn't work correctly either. :thinking: I'll look into this...

jkloetzke:
Ok, Windows paths should work now as URLs. The following formats are supported:

* C:\foo.bar
* C:/foo.bar
* file:///C:/foo.bar
* file:///C:\foo.bar
* \\\\server\path
* file:///\\\\server\path

sbixl:
Hi, I have tested the changes but it looks like there's still some small change missing.

For example if I use the following [1] `url: C:/X/Y/my-pkg.zip` everything is working like expected.
Buf if I use [2] `url: C:\\X\\Y\\my-pkg.zip` the archive is not downloaded. Within the function [async def invoke(self, invoker)](https://github.com/jkloetzke/bob/blob/99b8457d3646d4f30dd5f69027060731849b8eb3/pym/bob/scm/url.py#L239) no download is triggered because the destination is not correct (it points to the destination of the source directory [2] we want to download from). You can see this too if the function [def __init__(self, spec, overrides=[], tidy=None)](https://github.com/jkloetzke/bob/blob/99b8457d3646d4f30dd5f69027060731849b8eb3/pym/bob/scm/url.py#L161) is called and you check the spec parameter. The property `fileName` contains not the file name 'my-pkg.zip' instead it contains the full path to the source [2].

Another "cosmetic" question is, whether it's suitbable for a windows user to add a url like C:\X\Y\my-pkg.zip instead of C:\\\X\\\Y\\\my-pkg.zip to the recipes?

jkloetzke:
> Hi, I have tested the changes but it looks like there's still some small change missing.

:+1: 

> For example if I use the following [1] `url: C:/X/Y/my-pkg.zip` everything is working like expected.
> Buf if I use [2] `url: C:\\X\\Y\\my-pkg.zip` the archive is not downloaded. Within the function [async def invoke(self, invoker)](https://github.com/jkloetzke/bob/blob/99b8457d3646d4f30dd5f69027060731849b8eb3/pym/bob/scm/url.py#L239) no download is triggered because the destination is not correct (it points to the destination of the source directory [2] we want to download from). You can see this too if the function [def **init**(self, spec, overrides=[], tidy=None)](https://github.com/jkloetzke/bob/blob/99b8457d3646d4f30dd5f69027060731849b8eb3/pym/bob/scm/url.py#L161) is called and you check the spec parameter. The property `fileName` contains not the file name 'my-pkg.zip' instead it contains the full path to the source [2].

Nice catch. Should be fixed now.

> Another "cosmetic" question is, whether it's suitbable for a windows user to add a url like `C:\X\Y\my-pkg.zip` instead of `C:\\X\\Y\\my-pkg.zip` to the recipes?

This depends on how you write your recipes. The backslash escaping is part of the YAML syntax. If you put the file name in quotes then you have to escape the backslash:

```yaml
url: "C:\\X\\Y\\my-pkg.zip"
```

If you leave out the quotes or use a block string then you do not need to escape the backslashes:

```yaml
url: C:\X\Y\my-pkg.zip
---
url: |
    C:\X\Y\my-pkg.zip
```

sbixl:
Hi, downloading is working now 👍. I have briefly examined the formats you mentioned and they are working fine. 

Regarding:

`If you leave out the quotes or use a block string then you do not need to escape the backslashes.`

I already tried `url: C:\X\Y\my-pkg.zip` but bob complains about the following:

`Error: URL must be a fully qualified path name
Error: Recipe tools::my-pkg in checkoutSCM: dir:., url:C:XYmy-pkg.zip in checkoutSCM: dir:., url:C:XYmy-pkg.zip failed`

Somewhere the slashed are lost 🤔

jkloetzke:
> > If you leave out the quotes or use a block string then you do not need to escape the backslashes.
> 
> I already tried `url: C:\X\Y\my-pkg.zip` but bob complains about the following:
> 
> `Error: URL must be a fully qualified path name Error: Recipe tools::my-pkg in checkoutSCM: dir:., url:C:XYmy-pkg.zip in checkoutSCM: dir:., url:C:XYmy-pkg.zip failed`
> 
> Somewhere the slashed are lost thinking

:man_facepalming: I totally forgot that the `url` (like almost any other property) is subject to [string substitution](https://bob-build-tool.readthedocs.io/en/latest/manual/configuration.html#string-substitution) in Bob. So there are two layers of (possible) escaping that need to be considered: YAML *and* Bob.

mahaase:
Hey Jan,

thank u very much :) (btw: sry 4 late reply.)

following stuff i tested:
* vs2019
   - build (sub-) run targets 👍 
   - start debugger for module with exe result (automatic detected the executable) 👍 
   - handling of vs start-project feature 👍 
   - catch breakpoints
   - error resolving in issue-list
   - jump to errors
   - jump to include files
   - ...

I guess, the most important features for normal software developer should be fine now. 

My focus was Windows in MSYS2 system with bash script language.

I detected no issues so far, I am waiting the reintegration into master :)

jkloetzke:
I'm currently refactoring things a bit, hopefully the last time. :wink:

I tried to create some cross-patform tutorial recipes but the language decision in `config.yaml` does not scale well. Instead I'm trying to see if a per-recipe language decision (where classes may support multiple languages) works better...

sbixl:
Hi Jan,

while using bob native under windows in combination with symlinks, an internal exception occurs when using python versions < 3.8. It looks like the problem has been fixed by [issue37834](https://bugs.python.org/issue37834) in Python Version 3.8. I think we should add this information to the documentation for windows?

Bob version 0.16+unknown
  File "c:\projects\bob\pym\bob\scripts.py", line 141, in catchErrors
    ret = fun(*args, **kwargs)
  File "c:\projects\bob\pym\bob\scripts.py", line 242, in cmd
    ret = cmd(args.args, bobRoot)
  File "c:\projects\bob\pym\bob\scripts.py", line 29, in __develop
    doDevelop(*args, **kwargs)
  File "c:\projects\bob\pym\bob\cmds\build\build.py", line 288, in doDevelop
    commonBuildDevelop(parser, argv, bobRoot, True)
  File "c:\projects\bob\pym\bob\cmds\build\build.py", line 234, in commonBuildDevelop
    builder.cook(backlog, True if args.build_mode == 'checkout-only' else False)
  File "c:\projects\bob\pym\bob\cmds\build\builder.py", line 640, in cook
    raise self.__buildErrors[0]
  File "c:\projects\bob\pym\bob\cmds\build\builder.py", line 542, in __taskWrapper
    ret = await coro()
  File "c:\projects\bob\pym\bob\cmds\build\builder.py", line 735, in _cookStep
    await self._cookPackageStep(step, checkoutOnly, depth, mayUpOrDownload, buildId)
  File "c:\projects\bob\pym\bob\cmds\build\builder.py", line 1052, in _cookPackageStep
  File "c:\projects\bob\pym\bob\cmds\build\builder.py", line 54, in hashWorkspace
    os.path.join(step.getWorkspacePath(), "..", "cache.bin"))
  File "c:\projects\bob\pym\bob\utils.py", line 405, in hashDirectory
    return DirHasher(index, ignoreDirs).hashDirectory(path)
  File "c:\projects\bob\pym\bob\utils.py", line 385, in hashDirectory
    return self.__hashDir(os.fsencode(path))
  File "c:\projects\bob\pym\bob\utils.py", line 375, in __hashDir
    for (e, f, s) in entries
  File "c:\projects\bob\pym\bob\utils.py", line 375, in <listcomp>
    for (e, f, s) in entries
  File "c:\projects\bob\pym\bob\utils.py", line 327, in __hashEntry
    digest = self.__hashDir(prefix, entry)
  File "c:\projects\bob\pym\bob\utils.py", line 375, in __hashDir
    for (e, f, s) in entries
  File "c:\projects\bob\pym\bob\utils.py", line 375, in <listcomp>
    for (e, f, s) in entries
  File "c:\projects\bob\pym\bob\utils.py", line 329, in __hashEntry
    digest = self.__index.check(prefix, entry, s, DirHasher.__hashLink)
  File "c:\projects\bob\pym\bob\utils.py", line 294, in check
    digest = process(os.path.join(prefix, name) if name else prefix)
  File "c:\projects\bob\pym\bob\utils.py", line 344, in __hashLink
    m.update(os.readlink(path))
TypeError: readlink() argument 1 must be str, not bytes

jkloetzke:
@sbixl The latest problem should be fixed. What worries me more is that the handling of symlinks is not consistent on Windows. If you have the privileges you can create them. OTOH MSYS2 is just copying the file when calling `ln -s`. This will lead to problems with binary artifacts...

I've refactored things quite a bit and it is finished from my POV. Please test it thoroughly. Unless something is really broken I will merge it next week.

Note that there are still issues left. The remaining TODOs are tracked here: https://github.com/BobBuildTool/bob/projects/4.

kbartecki:
Hey Jan,

u have to uncomment
#MSYS=winsymlinks:nativestrict

in e.g. using msys2.exe, than msys2.ini or if u use msys2_shell.cmd, that inside the cmd file.

if option disabled, msys uses built-in copy functionality.

BR.

jkloetzke:
Thanks @kbartecki for the hint. Actually I'm more concerned that you need a special privilege (SeCreateSymbolicLinkPrivilege) to create a symlink. It also looks like even the Python tarfile implementation will fall back to copies if a symlink cannot be created (for whatever reason). So extracting the same archive or running the same script (with `ln -s`) depends on your local settings. This will certainly lead to "Restart build due to wrongly predicted sources" problems.

While this won't affect this PR we will have to solve it before the next release.

pym/bob/input.py
```diff
@@ -108,10 +108,12 @@ def sliceHost(digest):
         """Extract host fingerprint digest part (if any)."""
         return digest[20:]
 
-def fetchScripts(recipe, prefix, resolveBash = lambda x, y: x):
+def fetchScripts(recipe, prefix, resolveBash = lambda x, y: x, resolvePwsh = lambda x, y: x):
     return {
         ScriptLanguage.BASH : resolveBash(recipe.get(prefix + "ScriptBash",
             recipe.get(prefix + "Script")), prefix + "Script[Bash]"),
+        ScriptLanguage.PWSH : resolvePwsh(recipe.get(prefix + "ScriptPwsh"),
+            prefix + "ScriptPwsh"),
```
sbixl: Hi,
I suppose that's what the expression should be: `ScriptLanguage.PWSH : resolvePwsh(recipe.get(prefix + "ScriptPwsh", recipe.get(prefix + "Script")), prefix + "Script[Pwsh]")` ? Without this only the package step is executed because no scripts for checkout and build step can't be fetched. I tested the change in my environment (native windows) and everything is building fine again.

jkloetzke: Actually not. The idea was that `*Script` is just an alias for `*ScriptBash`. If you want to use PowerShell scripts they always have to be written in `*SriptPwsh`.

*BUT*, I think your idea makes sense. Then the `*Script` will be taken for whatever language is used to build the recipe. Classes can use `*ScriptBash` and `*scriptPwsh` if they like to support both. I will just have to update the documentation...

mahaase:
got an issue! I have a recipe with:

provideTools:
    codegen:
        path: "usr/bin"
        libs: ["/usr/lib/pym/codegen"]

with the bob of native-windows branch the usage of codegen fails, because LD_LIBRARY_PATH doesn't exists. (windows, msys2)

jkloetzke:
> with the bob of native-windows branch the usage of codegen fails, because LD_LIBRARY_PATH doesn't exists. (windows, msys2)
> LD_LIBRARY_PATH is completely missing (also in env file).

It was missing deliberately because I thought that LD_LIBRARY_PATH is not used on Windows. Apparently it is. 😉

Should be fixed.

jkloetzke:
Unless somebody objects I will merge it at the end of the day...

pym/bob/input.py
```diff
@@ -2919,7 +2965,14 @@ def __loadPlugin(self, mangledName, fileName, name):
         projectGenerators = manifest.get('projectGenerators', {})
         if not isinstance(projectGenerators, dict):
             raise ParseError("Plugin '"+fileName+"': 'projectGenerators' has wrong type!")
-        self.__projectGenerators.update(projectGenerators)
+        if projectGenerators:
+            if compareVersion(apiVersion, "0.16.1.dev33") < 0:
+                # cut off extra argument for old generators
+                projectGenerators = {
+                    name : lambda package, args, extra, bobRoot: generator(package, args, extra)
```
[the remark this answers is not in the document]
jkloetzke: Actually not. Starting with commit 98aa3506df2ffaac4f4e79256602827fa66954dc the generator plugin will get the `bobRoot` path. This code is there to support plugins that were written before that and that do not expect a fourth argument.

jkloetzke:
Thanks everybody for the testing and feedback.

PR #270, opened 2019-10-03
builder/project: add packages download option

### LINKED ISSUE #264: bob project --download no
bob project includes "--download no" implicitly. I doesn't got, where this come from.

IMO this is not, what we want. Especially also the --exclude flat will be ignored, e.g. we exclude boost and qt lib, than I don't want do download and build **the sources**.

rhubert:
With download modules a regex can be used to specify which packages
should be downloaded.

Fixes #264

### DIFF
--- a/doc/manpages/bob-build-dev.rst
+++ b/doc/manpages/bob-build-dev.rst
@@ -48,7 +48,7 @@ Options
     built package(s).
 
 ``--download MODE``
-    Download from binary archive (yes, no, deps, forced, forced-deps)
+    Download from binary archive (yes, no, deps, forced, forced-deps, packages)
 
     no
       build given module and it's dependencies from sources
@@ -66,6 +66,8 @@ Options
     forced-fallback
       combination of forced and forced-deps modes: if forced fails fall back to
       forced-deps
+    packages=<packages regex>
+      download modules that match a given regular expression, build all other.
 
 ``--incremental``
     Reuse build directory for incremental builds.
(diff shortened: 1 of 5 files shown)


doc/manpages/bob-build-dev.rst
```diff
@@ -66,6 +66,8 @@ Options
     forced-fallback
       combination of forced and forced-deps modes: if forced fails fall back to
       forced-deps
+    modules=<modules regex>
+      download modules that match a given regular expression, build all other.
```
jkloetzke: I would rather call it `packages=` to be consistent with the usual naming.

pym/bob/cmds/build/build.py
```diff
@@ -38,6 +38,13 @@ def runHook(recipes, hook, args):
     return ret
 
 def commonBuildDevelop(parser, argv, bobRoot, develop):
+    def _downloadArgument(arg):
+        if arg.startswith('modules='):
+            return arg[8:]
```
jkloetzke: I don't think it's a good idea to cut away the prefix here already. Imagine a package is called `yes`. I would rather cut the prefix when compiling the regex.

pym/bob/cmds/build/builder.py
```diff
@@ -355,9 +356,11 @@ def setDownloadMode(self, mode):
             self.__archive.wantDownload(True)
             self.__downloadDepth = 0
             self.__downloadDepthForce = 1
-        else:
-            assert mode == 'no'
+        elif mode == 'no':
             self.__archive.wantDownload(False)
+        else:
+            self.__archive.wantDownload(True)
+            self.__downloadModules = re.compile(mode)
```
jkloetzke: We should catch a possible invalid regex here.

```suggestion
            try:
                self.__downloadModules = re.compile(mode)
            except re.error as e:
                raise BuildError("Invalid download regex '{}': {}".format(e.pattern), e)
```

jkloetzke:
Thanks. :+1:

PR #276, opened 2019-10-31
jenkins: cleanup @tmp directories before build

### LINKED ISSUE #273: @tmp directory in workspace break some features
already known problem with git-client jenkins plugin reported here by rhubert: 
https://github.com/jenkinsci/git-client-plugin/pull/353

issues like:
 - restarting cause of wrongly predicted sources
 - no artifact download if sources checked out
 - ...

rhubert:
Some jenkins plugins create @tmp directories in the workspace for some
reasons. Remove them as they influence the buildId calculation.

Fixes #273

### DIFF
--- a/pym/bob/cmds/jenkins.py
+++ b/pym/bob/cmds/jenkins.py
@@ -848,6 +848,10 @@ class JenkinsJob:
                         AUDIT=JenkinsJob._auditName(d),
                         WSP_PATH=d.getWorkspacePath()))
 
+        prepareCmds.append("# remove @tmp directories created by some jenkins plugins")
+        for d in sorted(self.__checkoutSteps.values()):
+            prepareCmds.append("rm -rf {}".format(" ".join(quote(d.getWorkspacePath() + "/" + s + "@tmp") for s in d.getScmDirectories())))
+
         # Create first "prepare" shell action. The actual command is set at the
         # end because the prepare commands are generated throughout the
         # generating process.


pym/bob/cmds/jenkins.py
```diff
@@ -848,6 +848,11 @@ def ensureFingerprint(step):
                         AUDIT=JenkinsJob._auditName(d),
                         WSP_PATH=d.getWorkspacePath()))
 
+        # remove @tmp directory generated by some jenkins plugins
+        for d in sorted(self.__checkoutSteps.values()):
+            prepareCmds.append("# remove @tmp directories created by some jenkins plugins")
```
jkloetzke: The comment should probably be added only once...

pym/bob/cmds/jenkins.py
```diff
@@ -848,6 +848,11 @@ def ensureFingerprint(step):
...
+            prepareCmds.append("rm -rf {}".format(" ".join(d.getWorkspacePath() + "/" + s + "@tmp" for s in d.getScmDirectories())))
```
jkloetzke: Better quote the argument to `rm -rf`...

```suggestion
            prepareCmds.append("rm -rf {}".format(" ".join(quote(d.getWorkspacePath() + "/" + s + "@tmp") for s in d.getScmDirectories())))
```

pym/bob/cmds/jenkins.py
jkloetzke: But you still add the comment for each SCM again...

rhubert: now I got you point :sleepy:

PR #279, opened 2019-11-13
Fix detection of closing single quotation mark in StringParser

gruberchr:
There is a bug in function `getSingleQuoted()` of class `StringParser`, which ignores a closing single quotation mark, if it is at the end of the string. This bug is fixed.

### DIFF
--- a/pym/bob/stringparser.py
+++ b/pym/bob/stringparser.py
@@ -81,13 +81,12 @@ class StringParser:
         i = self.index
         while i < self.end:
             if self.text[i] == "'":
-                i += 1
                 break
             i += 1
         if i >= self.end:
             raise ParseError("Missing closing \"'\"")
-        ret = self.text[self.index:i-1]
-        self.index = i
+        ret = self.text[self.index:i]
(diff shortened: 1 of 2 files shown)


gruberchr:
By the way, there is no test case at the moment, which checks, if variable and command substitution is disabled in substrings quoted with single quotes.

jkloetzke:
Good catch. I took the freedom to reorder the commits so that a `git bisect` will not trip on the test case without the fix. I also adjusted the commit messages slightly.

> By the way, there is no test case at the moment, which checks, if variable and command substitution is disabled in substrings quoted with single quotes.

You're right. Do you plan to add some test cases? I could also put them on my todo list, though...

gruberchr:
> You're right. Do you plan to add some test cases? I could also put them on my todo list, though...

No, we don't. Put it on your todo list.

jkloetzke:
> > You're right. Do you plan to add some test cases? I could also put them on my todo list, though...
> 
> No, we don't. Put it on your todo list.

Ok, no problem.

PR #280, opened 2019-11-13
Use function quote() from module shlex instead of pipes

gruberchr:
Function `pipes.quote()` is deprecated since Python 2.7

see https://docs.python.org/2/library/pipes.html


jkloetzke:
Thanks. :+1:

PR #283, opened 2019-11-19
builder: do not cancel buildId calculation

### LINKED ISSUE #271: Parallel build does not store full state on error
If one package fails in the parallel build the other running steps are not remembered if they finish successfully. A subsequent `--resume` will build them again even though they are already done.

rhubert:
If a job fails the overall execution is stopped and no more tasks are
started. Since buildId calculation is also a executed as a task the
buildId of a parallel running successfull job was not calculated and the
build was executed in the next bob build.

Handle this by adding a ignoreExecutionStop argument to __yieldJobWhile.
This argument should be set to true for all 'cleanup' tasks.

Fixes #271.


jkloetzke:
Thanks! :+1:

PR #288, opened 2019-11-25
generators: vs2019 add also local includes to include search paths

mahaase:
Hi,

I added the local include paths of the module also to the include search paths for intellisense.
There was the problem, that !!SOME!! includes were not accessable by pressing F12 or STRG + left mouse button. (behavior was a bit strange, because some includes worked - but with this fix, all includes work fine).

BR.


jkloetzke:
Thanks. :+1:

PR #294, opened 2019-12-18
query-path: Show message on empty output

Ferruck:
In case that no paths for the given query could be found, present a
message to the user stating exactly that. Bob's silence in these cases
could be quite confusing.

Note that since query-path is mainly used for scripting purposes and
these scripts may rely on the empty output, the message is printed to
stderr.

Update the copyright statement whilst we're at it.

### DIFF
--- a/pym/bob/cmds/build/query.py
+++ b/pym/bob/cmds/build/query.py
@@ -1,5 +1,5 @@
 # Bob build tool
-# Copyright (C) 2016-2018  TechniSat Digital GmbH
+# Copyright (C) 2016-2019, The BobBuildTool Contributors
 #
 # SPDX-License-Identifier: GPL-3.0-or-later
 
@@ -9,6 +9,7 @@ from ...utils import processDefines
 from string import Formatter
 import argparse
 import os
+import sys
 
 from .builder import LocalBuilder
 from .state import DevelopDirOracle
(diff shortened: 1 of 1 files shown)


jkloetzke:
Hmm, I think there are two distinct cases that are now not distinguishable:

* The query matched no package and there was no path that could be queried
* The query actually matched one or more packages but the requested information could not be retrieved because the workspace is not existing or unknown

From a users perspective the added error message is only for the first case. But the latter case should get a proper message too which was the original intention I guess.

Ferruck:
That's a point. I'll have a look into it again.

Before we take another round: Is printing to stderr a proper solution in your eyes?

jkloetzke:
> Before we take another round: Is printing to stderr a proper solution in your eyes?

Yes. I guess adding a `-q` option makes some sense to suppress the messages. And also a `--fail` option like curl could be handy to exit with an error if the requested information could not be retrieved...

Ferruck:
> Yes. I guess adding a -q option makes some sense to suppress the messages. And also a --fail option like curl could be handy to exit with an error if the requested information could not be retrieved...

I'll tackle this in a second commit since I'll first have to learn how command line options are handled inside Bob.

Is the message stuff OK now?

Ferruck:
> [...] exit with an error if the requested information could not be retrieved...

Does this include queries that matched no packages or only non-present paths?

jkloetzke:
> > [...] exit with an error if the requested information could not be retrieved...
> 
> Does this include queries that matched no packages or only non-present paths?

I would say only the latter. You might enter a query that matches zero packages. This is also not an error in `bob dev/build`. Consistency FTW... :wink:

PR #296, opened 2020-01-16
input: add support for if-expressions

rhubert:
Support real expressions for the if property. This makes the if syntax
more readable.

E.g. instead of:
```
   if: "$(or,$(eq,${FOO},bar),${BAZ})"
```

you can now write:
```
   if: !expr |
      "${FOO}" == "bar" || "${BAZ}"
```

### DIFF
--- a/pym/bob/input.py
+++ b/pym/bob/input.py
@@ -1816,7 +1816,7 @@ class Recipe(object):
                 use = dep.get("use", use)
                 newCond = dep.get("if")
                 if newCond is not None:
-                    cond = "$(and,{},{})".format(cond, newCond) if cond is not None else newCond
+                    cond = cond + [newCond] if cond is not None else [ newCond ]
                 name = dep.get("name")
                 if name:
                     if "depends" in dep:
@@ -2207,7 +2207,15 @@ class Recipe(object):
             env.setFunArgs({ "recipe" : self, "sandbox" : bool(sandbox) and sandboxEnabled,
                 "__tools" : tools })
 
-            if not env.evaluate(dep.condition, "dependency "+dep.recipe): continue
+            if dep.condition:
+                conditionsTrue = False
+                for cond in dep.condition:
+                    if not env.evaluate(cond, "dependency "+dep.recipe): break
+                else:
+                    conditionsTrue = True
+
+                if not conditionsTrue: continue
+
             r = self.__recipeSet.getRecipe(dep.recipe)
             try:
                 if r.__packageName in stack:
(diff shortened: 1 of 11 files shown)


pym/bob/input.py
```diff
@@ -2203,11 +2203,20 @@ def prepare(self, inputEnv, sandboxEnabled, inputStates, inputSandbox=None,
...
-            if not env.evaluate(dep.condition, "dependency "+dep.recipe): continue
+            if dep.condition:
+                conditionsTrue = False
+                for cond in dep.condition:
+                    if not env.evaluate(cond, "dependency "+dep.recipe): break
+                else:
+                    conditionsTrue = True
+
+                if not conditionsTrue: continue
+
```
jkloetzke: I guess this could be simplified as

```python
if not all(env.evaluate(cond, "dependency "+dep.recipe) for cond in dep.condition): continue
```

pym/bob/input.py
```diff
@@ -2801,7 +2811,7 @@ def updateArchive(x): self.__archive = x
             ),
             "scmOverrides" : BuiltinSetting(
                 schema.Schema([{
-                    schema.Optional('match') : schema.Schema({ str: str }),
+                    schema.Optional('match') : schema.Schema({ str: str}),
```
jkloetzke: Stray whitespace change

pym/bob/scm/scm.py
```diff
@@ -38,7 +38,7 @@ def __doesMatch(self, scm, env):
         for (key, value) in self.__match.items():
             if key not in scm: return False
             value = env.substitute(value, "svmOverride::match")
-            if not fnmatch.fnmatchcase(scm[key], value): return False
+            if not fnmatch.fnmatchcase(str(scm[key]), value): return False
```
jkloetzke: I think a comment would be nice to explain why this is needed...

pym/bob/stringparser.py
```diff
@@ -196,13 +198,169 @@ def getCommand(self):
...
+            self.opStr, self.right)
+
+    def evalExpression(self, env):
+        return self.op(self.left.evalExpressionToString(env), self.right.evalExpressionToString(env))
+
+class IfExpressionParser:
+    def __init__(self):
+
+        # create parsing grammer
+        locationPath = pyparsing.Forward()
```
jkloetzke: Unused variable

pym/bob/stringparser.py
```diff
@@ -196,13 +198,169 @@ def getCommand(self):
...
+        return ret[0].evalExpression(env)
+
 
 class Env(MutableMapping):
     def __init__(self, other={}):
         self.data = dict(other)
         self.funs = []
         self.funArgs = {}
         self.touched = [ set() ]
+        self.__expressionParser = IfExpressionParser()
```
jkloetzke: Env objects are created quite often. AFAICS we only ever need one instance of the `IfExpressionParser`, right? So I think we can lazily create one global instance and use it in `evaluate()` when needed.

jkloetzke:
:+1:

PR #297, opened 2020-01-17
input: add support for 'if' in scmOverrides

rhubert:
Add 'if' keyword to scmOverrides to selectively enable or disable
scmOverrides.

### DIFF
--- a/doc/manual/configuration.rst
+++ b/doc/manual/configuration.rst
@@ -1742,9 +1742,12 @@ are applied after string substitution. The general syntax looks like the followi
           url:
             pattern: "foo"
             replacement: "bar"
+        if: !expr |
+          "${BOB_RECIPE_NAME}" == "foo"
 
 The ``scmOverrides`` key takes a list of one or more override specifications.
-The override is first matched via patterns that are in the ``match`` section.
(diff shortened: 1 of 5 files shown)


doc/manual/configuration.rst
```diff
@@ -1742,9 +1742,12 @@ are applied after string substitution. The general syntax looks like the followi
           url:
             pattern: "foo"
             replacement: "bar"
+        if: !expr |
+          "${USE_OVERRIDES} == 1"
```
jkloetzke: I think the condition is wrong: it is just a single string that is interpreted as `True`. Maybe it would be nice to have a more useful example too. Something like:

```suggestion
          "${BOB_RECIPE_NAME}" == "foo"
```

pym/bob/scm/scm.py
```diff
@@ -35,6 +36,7 @@ def __setstate__(self, s):
         self.__init()
 
     def __doesMatch(self, scm, env):
+        if self.__if and not env.evaluate(self.__if, "scmOverride::if"): return False
```
jkloetzke: Shouldn't that be

```suggestion
        if self.__if is not None and not env.evaluate(self.__if, "scmOverride::if"): return False
```

? Because an empty string should be considered `False`.

PR #300, opened 2020-02-01
Query path options

Ferruck:
This is a follow-up on #294 where we agreed on having both a `-q` and a `--fail` option for `bob query-path`.

### DIFF
--- a/test/query-path/run.sh
+++ b/test/query-path/run.sh
@@ -17,6 +17,18 @@ test -z "$(run_bob query-path --release root)"
 test -z "$(run_bob query-path --release root/interm1/child)"
 test -z "$(run_bob query-path --release root/interm2/child)"
 
+# Still, must report error on stderr as long as we don't provide -q
+test -n "$(run_bob query-path --release root 2>&1 | grep 'not present')"
+test -n "$(run_bob query-path --release root/interm1/child 2>&1 | grep 'not present')"
+test -n "$(run_bob query-path --release root/interm2/child 2>&1 | grep 'not present')"
+test -z "$(run_bob query-path --release -q root 2>&1 | grep 'not present')"
+test -z "$(run_bob query-path --release -q root/interm1/child 2>&1 | grep 'not present')"
+test -z "$(run_bob query-path --release -q root/interm2/child 2>&1 | grep 'not present')"
+
+# Also, errors on non-present packages should be reported
+test -n "$(run_bob query-path --release notpresent 2>&1 | grep 'Naptime')"
+test -z "$(run_bob query-path --release -q notpresent 2>&1 | grep 'Naptime')"
+
 # Perform a full release build. Must report paths for everything.
 run_bob build root
 set -- $(run_bob query-path --release root)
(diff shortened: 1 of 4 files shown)


Ferruck:
I'll have a look into the unit tests and add one or two.

jkloetzke:
I guess adding some simple tests to the `query-path` black box test should be sufficient...

Ferruck:
From my side this is done. What's your opinion?

test/query-path/run.sh
```diff
@@ -17,9 +17,27 @@ test -z "$(run_bob query-path --release root)"
...
+test -n "$(run_bob query-path --release root/interm1/child 2>&1 | grep 'not present')"
+test -n "$(run_bob query-path --release root/interm2/child 2>&1 | grep 'not present')"
+test -z "$(run_bob query-path --release -q root 2>&1 | grep 'not present')"
+test -z "$(run_bob query-path --release -q root/interm1/child 2>&1 | grep 'not present')"
+test -z "$(run_bob query-path --release -q root/interm2/child 2>&1 | grep 'not present')"
+
+# And must fail if --fail is provided
+test !$(run_bob query-path --release --fail root)
+test !$(run_bob query-path --release --fail root/interm1/child)
+test !$(run_bob query-path --release --fail root/interm2/child)
```
jkloetzke: I'm not sure if this is really testing what is should. I think the `!` is subject to history expansion. :thinking: Have you tested the test? :wink: 

I would suggest `expect_fail run_bob query-path --release --fail root`...

Ferruck: Well, maybe I'm misreading [man test](https://linux.die.net/man/1/test), or maybe I'm even referring to the wrong man page, or - and that's the most probable case - my novice shell knowledge isn't sufficient, but

> ! EXPRESSION
>
>     EXPRESSION is false 

:thinking: May you enlighten me?

Yes, I've run the tests, but maybe I've misinterpreted their behaviour... :flushed:

I'll change it to your suggestion as soon as I get to it. :male_shrug:

jkloetzke: The `test` command gets the "expression" as argument. As such it only looks at strings that are passed to it. Apart from the different operators an empty string is "false" and everything else is "true". The exit status of `test` is 0 if the evaluated expression is true otherwise 1.

You wanted to test is the exit status of `run_bob`. What you instead tested is the output of `run_bob`. See this example:

```sh
#!/bin/bash -e

if test !$(/bin/false) ; then
   echo false is true
fi
if test !$(/bin/true) ; then
   echo true is true
fi
```

Output:

```
false is true
true is true
```

Typing the above example on the command line is another story and won't work for a different reason. The "!" is the history expansion meta character. See `HISTORY EXPANSION` in the bash man page for the gory details...

ThomasFeher:
This and #294 is very much appreciated. It helps a lot with building tools around bob.

Thank you!

test/query-path/run.sh
Ferruck: Bash, nuff said...

But thanks for your comprehensive explanation. As I said, I'll change it once I get to it.

jkloetzke:
Thanks 👍

PR #319, opened 2020-04-12
cvs scm: make command invocation work again

stefanreuther:
The rework in 4e488a75 forgot to 'await' the commands, causing a warning to
be generated and the command being not executed:

    cvs.py:74: RuntimeWarning: coroutine 'Invoker.checkCommand' was never awaited

----

In case this was a canary whether I'm still using it: of course I do! I even build MS-DOS programs with it now! :-)

### DIFF
--- a/pym/bob/scm/cvs.py
+++ b/pym/bob/scm/cvs.py
@@ -63,14 +63,14 @@ class CvsScm(Scm):
             # We therefore use a 'cvs up' after the initial 'cvs co', to get the same behaviour for the initial and subsequent builds.
             if re.match('^:ext:', rootarg) and self.__dir == '.':
                 os.symlink(".", invoker.join("__tmp$$"))
-                invoker.checkCommand(["cvs", "-qz3", "-d", rootarg, "co"] + revarg +
+                await invoker.checkCommand(["cvs", "-qz3", "-d", rootarg, "co"] + revarg +
                     ["-d", "__tmp$$", self.__module], env=env)
                 os.unlink(invoker.join("__tmp$$"))
             else:
-                invoker.checkCommand(["cvs", "-qz3", "-d", rootarg, "co"] + revarg +
+                await invoker.checkCommand(["cvs", "-qz3", "-d", rootarg, "co"] + revarg +
                     ["-d", self.__dir, self.__module], env=env)
 
-        invoker.checkCommand(["cvs", "-qz3", "-d", rootarg, "up", "-dP"] + revarg +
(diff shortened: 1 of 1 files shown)


jkloetzke:
> The rework in [4e488a7](https://github.com/BobBuildTool/bob/commit/4e488a7599f090c78cf3d63180416b123b244944) forgot to 'await' the commands, causing a warning to
> be generated and the command being not executed:
> 
> ```
> cvs.py:74: RuntimeWarning: coroutine 'Invoker.checkCommand' was never awaited
> ```

Thanks for the fix.

> In case this was a canary whether I'm still using it: of course I do!

:laughing: :+1:  Well, in fact I thought that we have a smoke test to catch such embarrassing errors. :see_no_evil: 

> I even build MS-DOS programs with it now! :-)

Wow. I'm surprised this is even possible to cross compile. Or have you gone native? :wink:

stefanreuther:
> Well, in fact I thought that we have a smoke test to catch such embarrassing errors.

If I recall correctly, the reason for not doing that was that the CI pipeline has no `cvs` installed.

> Wow. I'm surprised this is even possible to cross compile. Or have you gone native?

I am using dosemu ([class](https://github.com/stefanreuther/bob-recipes/blob/master/classes/dos.yaml), [recipe](https://github.com/stefanreuther/bob-recipes/blob/master/recipes/planets/dos/env.yaml)).

Scripting this stuff in bob means I can finally put that old source tree at rest that originated on a 386 and has been carefully salvaged from one computer to the next. Plus, cloning a 120 MB tree to try something out now is actually easy whereas that 386 had only 100 MB total disk space :)

jkloetzke:
> > Well, in fact I thought that we have a smoke test to catch such embarrassing errors.
> 
> If I recall correctly, the reason for not doing that was that the CI pipeline has no `cvs` installed.

Could be. Anyway, `cvs` is available on the CI pipeline because there are already some tests that needed them. I've added them some time ago. 

> > Wow. I'm surprised this is even possible to cross compile. Or have you gone native?
> 
> I am using dosemu ([class](https://github.com/stefanreuther/bob-recipes/blob/master/classes/dos.yaml), [recipe](https://github.com/stefanreuther/bob-recipes/blob/master/recipes/planets/dos/env.yaml)).
> 
> Scripting this stuff in bob means I can finally put that old source tree at rest that originated on a 386 and has been carefully salvaged from one computer to the next. Plus, cloning a 120 MB tree to try something out now is actually easy whereas that 386 had only 100 MB total disk space :)

:sunglasses: Nice!

PR #322, opened 2020-04-15
cvs scm: add a black-box test

stefanreuther:
Tests that initial checkout, update, and change to a branch work as
expected.

---

Followup to https://github.com/BobBuildTool/bob/pull/319 - if we can use CVS, let's test it.

### DIFF
--- a/test/cvs-scm/run.sh
+++ b/test/cvs-scm/run.sh
@@ -1,0 +1,71 @@
+#!/bin/bash -e
+#
+#  Quick functionality test for CVS
+#
+. ../test-lib.sh 2>/dev/null || { echo "Must run in script directory!" ; exit 1 ; }
+
+dir=$(mktemp -d)
+echo "Using scratch dir: $dir"
+trap 'rm -rf "$dir"; cleanup' EXIT
+cleanup
+
+# Bob recipes
+bob=$dir/bob
+mkdir -p "$bob/recipes"
+cp recipe1.yaml "$bob/recipes/t.yaml"
+
+# Directory to play in
+work=$dir/work
+mkdir "$work"
+echo init > "$work/file.txt"
+
+# CVS repository: contains a single repository "mod" with our file
+cvsroot=$dir/cvsroot
+mkdir "$cvsroot"
+cvs -d "$cvsroot" init
+(
+  cd "$work"
+  cvs -d "$cvsroot" import -m "message" mod vendor release
+  cvs -d "$cvsroot" co -d . mod
+)
+  
+
+##
+##  First check: check out initial state
+##
+run_bob -C "$bob" dev -DCVSROOT="$cvsroot" t
+result=$(run_bob -C "$bob" query-path -DCVSROOT="$cvsroot" -f {dist} t)
+test -n "$result"
+diff -q "$bob/$result/file.txt" "$work/file.txt"
+
+
+##
+##  Second check: modify the file and check that it is picked up by bob
+##  (incremental update works)
+##
+echo modif > "$work/file.txt"
+(
+  cd "$work"
+  cvs ci -m "update message" file.txt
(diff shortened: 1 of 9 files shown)


test/cvs-scm/run.sh
```diff
@@ -0,0 +1,94 @@
...
+root: true
+checkoutSCM:
+  scm: cvs
+  cvsroot: \${CVSROOT}
+  module: mod
+buildScript: |
+  cp "\$1/file.txt" .
+packageScript: |
+  cp "\$1/file.txt" .
+EOF
```
jkloetzke, line 23: Why not just put the recipes directly in the test directory like the other black box tests?

test/cvs-scm/run.sh
```diff
@@ -0,0 +1,94 @@
+#!/bin/bash -e
+#
+#  Quick functionality test for CVS
+#
+. ../test-lib.sh 2>/dev/null || { echo "Must run in script directory!" ; exit 1 ; }
+
+dir=$(mktemp -d)
```
jkloetzke, line 7: It's a good habit to insert a `trap 'rm -rf "$dir"' EXIT` directly afterwards. This will make sure that the temporary directory is removed regardless how the script exits.

test/cvs-scm/run.sh
```diff
@@ -0,0 +1,94 @@
...
+packageScript: |
+  cp "\$1/file.txt" .
+EOF
+
+run_bob -C "$bob" dev -DCVSROOT="$cvsroot" t
+diff -q "$bob/$result/file.txt" "$work/file.txt"
+
+# Clean up
+rm -rf "$dir"
+cleanup
```
jkloetzke, line 94: The `rm` is not needed if the `trap` is used above. The `cleanup` should actually be done as first step in the test to evict any stale state from the last run.

jkloetzke (commented):
Thanks for the test. :+1: Just some minor comments...

test/cvs-scm/run.sh
stefanreuther: Because I'm updating the recipes during the test to check the "change the branch" usecase. Creating the recipes from scratch was the simplest, safest I came up with.

test/cvs-scm/run.sh
stefanreuther: The ones I looked at didn't have that :)
But yes, that's a good idea.

stefanreuther:
By the way, making an identical test for git and svn doesn't sound too hard, so I could add that as well. Or do we already have something that fills that role?

test/cvs-scm/run.sh
jkloetzke: Oh, that's unfortunate. Looks like there is still room for improvement. :wink:

test/cvs-scm/run.sh
jkloetzke: Normally I would say you could use variable substitution and `-D...` when invoking bob. But it looks like it really makes a difference if the `rev` property on the SCM exists or not. :thinking: So maybe it's really the best option.

jkloetzke:
There is the `status` black box test that comes close. But it does many other things and is geared towards the `status` command. Having a pure SCM smoke test for all SCMs looks like a good idea. So yes, feel free to add them too. :+1:

test/cvs-scm/run.sh
stefanreuther: I wanted as little logic as possible in the test, to exercise more paths in the code. For CVS, a

     rev: ${REV}

would have been possible, passing "HEAD" for "not-from-a-branch".

I have now placed the recipes as separate files to save us a layer of quoting, but they are still copied into a fresh workspace.

jkloetzke:
Thanks. :+1:

Issue #257, opened 2019-06-15
Implement job server


jkloetzke:
Bob should understand and provide a GNU make compatible job server. Currently a system can be overloaded when Bob builds packages in parallel where each package invokes a parallel make.

PR #333, opened 2020-05-15
builder: add simple make jobserver

rhubert:
The make jobserver mechanism need two pipes. As long as it can read from
the request pipe new jobs are started. Once a job has finished a byte is
written back into the response pipe.

fixes: #257

### DIFF
--- a/pym/bob/cmds/build/build.py
+++ b/pym/bob/cmds/build/build.py
@@ -10,8 +10,10 @@ from ...tty import setVerbosity, setTui
 from ...utils import copyTree, processDefines, EventLoopWrapper
 import argparse
 import datetime
+import re
 import os
 import subprocess
+import stat
 import sys
 import time
 
@@ -130,6 +132,7 @@ def commonBuildDevelop(parser, argv, bobRoot, develop):
         else:
             cfg = recipes.getCommandConfig().get('build', {})
 
+        noJobs = args.jobs == None
         defaults = {
                 'destination' : '',
                 'force' : False,
@@ -155,6 +158,24 @@ def commonBuildDevelop(parser, argv, bobRoot, develop):
             args.jobs = os.cpu_count()
         elif args.jobs <= 0:
             parser.error("--jobs argument must be greater than zero!")
+
+        # parse MAKEFLAGS environment variable to setup number of jobs
+        # when called from make
+        makeFlags = os.environ.get('MAKEFLAGS')
+        makeFds = None
+        if makeFlags is not None:
+            jobs = re.search(r'-j([0-9]*)', makeFlags)
+            fds = re.search(r'--jobserver-auth=([0-9]*),([0-9]*)', makeFlags)
+            if jobs and fds and jobs.group(1) and fds.group(1) and fds.group(2):
+                if noJobs:
+                    args.jobs = int(jobs.group(1))
+                    makeFds = [int(fds.group(1)), int(fds.group(2))]
+                    try:
+                        if not all(stat.S_ISFIFO(os.stat(f).st_mode) for f in makeFds): makeFds = None
+                    except OSError:
+                        makeFds = None
+                else:
+                    print("warning: -j" + str(args.jobs) + " forced: resetting jobserver mode.");
 
         envWhiteList = recipes.envWhiteList()
         envWhiteList |= set(args.white_list)
@@ -184,6 +205,7 @@ def commonBuildDevelop(parser, argv, bobRoot, develop):
         builder.setAlwaysCheckout(args.always_checkout + cfg.get('always_checkout', []))
         builder.setLinkDependencies(args.link_deps)
         builder.setJobs(args.jobs)
+        builder.setMakeFds(makeFds)
         builder.setKeepGoing(args.keep_going)
         builder.setAudit(args.audit)
         if args.resume: builder.loadBuildState()
--- a/pym/bob/cmds/build/builder.py
+++ b/pym/bob/cmds/build/builder.py
@@ -104,6 +104,79 @@ class RestartBuildException(Exception):
 class CancelBuildException(Exception):
     pass
 
+class ExternalJobServer:
+    def __init__(self, makeFds):
+        self.__makeFds = makeFds
+
+    def getMakeFd(self):
+        return self.__makeFds
+
+class InternalJobServer:
+    def __init__(self, jobs):
+        self.__rfd, self.__wfd = os.pipe()
+        os.write(self.__wfd, bytes(jobs))
+
+    def getMakeFd(self):
+        return [ self.__rfd, self.__wfd ]
+
+class JobServerSemaphore:
+    def __init__(self, fds, recursive):
+        self.__sem = asyncio.Semaphore(0)
+        self.__waitersCnt = 0
+        self.__fds = fds
+        os.set_blocking (self.__fds[0], False);
+        self.__tokens = []
+        self.__recursive = recursive
+        self.__acquired = 0
+
+    @staticmethod
+    def jobavailableCallback(self):
+        try:
+            while self.__waitersCnt:
+                self.__tokens.append(os.read(self.__fds[0], 1))
+                self.__waitersCnt -= 1
+                self.__sem.release()
+        except BlockingIOError:
+            pass
+        finally:
+            if self.__waitersCnt == 0:
+                asyncio.get_event_loop().remove_reader(self.__fds[0])
+
+    async def acquire(self):
+        if self.__recursive and self.__acquired == 0:
+            self.__acquired += 1
+            return
+        try:
+            self.__tokens.append(os.read(self.__fds[0], 1))
+        except BlockingIOError:
+            if self.__waitersCnt == 0:
+                asyncio.get_event_loop().add_reader(self.__fds[0],
+                    JobServerSemaphore.jobavailableCallback, self)
+            self.__waitersCnt += 1
+            await self.__sem.acquire()
+            pass
+        self.__acquired += 1
+
+    async def __aenter__(self):
+        await self.acquire()
+        return None
+
+    def release(self):
+        if self.__acquired == 0:
+            raise ValueError ("BoundedSemaphore released too many times")
+        if self.__waitersCnt != 0:
+           self.__waitersCnt -= 1;
+           self.__sem.release()
+           if self.__waitersCnt == 0:
+               asyncio.get_event_loop().remove_reader(self.__fds[0])
+        else:
+            if not self.__recursive or self.__acquired > 1:
+                os.write(self.__fds[1], self.__tokens.pop())
+        self.__acquired -= 1
+
+    async def __aexit__(self, exc_type, exc, tb):
+        self.release()
+
 class LocalBuilderStatistic:
     def __init__(self):
         self.__activeOverrides = set()
@@ -190,6 +263,7 @@ cd {ROOT}
         self.__buildOnly = buildOnly
         self.__preserveEnv = preserveEnv
         self.__envWhiteList = set(envWhiteList)
+        self.__jobServer = None
         self.__archive = DummyArchive()
         self.__downloadDepth = 0xffff
         self.__downloadDepthForce = 0xffff
@@ -203,6 +277,7 @@ cd {ROOT}
         self.__alwaysCheckout = []
         self.__linkDeps = True
         self.__jobs = 1
+        self.__makeFds = None
         self.__bufferedStdIO = False
         self.__keepGoing = False
         self.__audit = True
@@ -259,6 +334,9 @@ cd {ROOT}
 
     def setJobs(self, jobs):
         self.__jobs = max(jobs, 1)
+
+    def setMakeFds(self, makeFds):
+        self.__makeFds = makeFds
 
     def enableBufferedIO(self):
         self.__bufferedStdIO = True
@@ -485,6 +563,8 @@ cd {ROOT}
         invoker = Invoker(spec, self.__preserveEnv, self.__noLogFile,
             self.__verbose >= INFO, self.__verbose >= NORMAL,
             self.__verbose >= DEBUG, self.__bufferedStdIO)
+        if step.jobServer() and self.__jobServer:
+            invoker.setMakeParameters(self.__jobServer.getMakeFd(), self.__jobs)
         ret = await invoker.executeStep(InvocationMode.CALL, cleanWorkspace)
         if not self.__bufferedStdIO: ttyReinit() # work around MSYS2 messing up the console
         if ret == -int(signal.SIGINT):
@@ -643,7 +723,15 @@ cd {ROOT}
             self.__fingerprintTasks = {}
             self.__allTasks = set()
             self.__buildErrors = []
-            self.__runners = asyncio.BoundedSemaphore(self.__jobs)
+            if sys.platform == "win32" or self.__jobs == 1:
+                self.__runners = asyncio.BoundedSemaphore(self.__jobs)
+            else:
+                if self.__makeFds:
+                    self.__jobServer = ExternalJobServer(self.__makeFds)
+                    self.__runners = JobServerSemaphore(self.__jobServer.getMakeFd(), True)
+                else:
+                    self.__jobServer = InternalJobServer(self.__jobs)
+                    self.__runners = JobServerSemaphore(self.__jobServer.getMakeFd(), False)
             self.__tasksDone = 0
             self.__tasksNum = 0
 
--- a/pym/bob/input.py
+++ b/pym/bob/input.py
@@ -914,6 +914,9 @@ class CoreStep(CoreItem):
     def isFingerprinted(self):
         return self.fingerprintMask != 0
 
+    @property
+    def jobServer(self):
+        return self.corePackage.recipe.jobServer
 
 class Step:
     """Represents the smallest unit of execution of a package.
@@ -1250,6 +1253,11 @@ class Step:
     def isRelocatable(self):
         """Returns True if the step is relocatable."""
         return False
+
+    def jobServer(self):
+        """Returns True if the jobserver should be used to schedule
+        builds for this step."""
+        return self._coreStep.jobServer()
 
     def _getProvidedDeps(self):
         p = self.__package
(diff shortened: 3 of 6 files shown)


rhubert:
Well, this basically works... But I'm unsure if the `__runners` semaphore should also be converted to a read on the pipe?

rhubert:
>  Now the tricky part is where the pipe is empty. In this case the task must be put to sleep and the pipe must be watched with [loop.add_reader()](https://docs.python.org/3.5/library/asyncio-eventloop.html#asyncio.AbstractEventLoop.add_reader) to get a callback when the pipe is readable again.

In this case the `read` is already blocking? I pushed some changes to make use of the Jobserver for bob tasks.

> Even make will only forward these if it understands that the callee understands the job protocol (see Job Slots).

Are you referring to `First, only command lines that make understands to be recursive invocations of make..` ? I have not fully understood what the author wanted to say with this paragraph. But a simple Makefile
```
all: foo.sh
	./foo.sh
```
and foo.sh as
``` 
#!/bin/sh
printenv MAKEFLAGS
```
produces:
```bash
$ make -j 8
./foo.sh
 -j8 --jobserver-auth=3,4
```

> So we need something like a jobServer recipe property to selectively enable the support. Otherwise we may break existing recipes.

I think this is indeed need. Otherwise existing recipes with `make -j $(nprocs)` produce a warning like: `make: warning: -j8 forced in submake: resetting jobserver mode.`

But I'm unsure if this should be a property or a policy?

jkloetzke:
> > Now the tricky part is where the pipe is empty. In this case the task must be put to sleep and the pipe must be watched with [loop.add_reader()](https://docs.python.org/3.5/library/asyncio-eventloop.html#asyncio.AbstractEventLoop.add_reader) to get a callback when the pipe is readable again.
> 
> In this case the `read` is already blocking?

Yes, but it must not block. Bob uses coroutines internally. When Bob blocks on `read()` then all internal processing and dispatching will stop. Instead the currently active coroutine must be suspended so that another coroutine can be scheduled.

> I pushed some changes to make use of the Jobserver for bob tasks.

I'm a bit confused about the thread that shuffles bytes out and into the pipe. I don't think this is necessary.

> > Even make will only forward these if it understands that the callee understands the job protocol (see Job Slots).
> 
> Are you referring to `First, only command lines that make understands to be recursive invocations of make..` ? I have not fully understood what the author wanted to say with this paragraph. But a simple Makefile

Yes, exactly. I was also confused initially but it makes sense. The `MAKEFLAGS` variable is always there. What is passed conditionally are the file descriptors. Without the `+` the file descriptors are closed. You can see that by inspecting `/proc/self/fd` from the shell script script.

I would not do it this way in Bob, though. If job mode is enabled the `MAKEFLAGS` are set and the file descriptors are passed. Otherwise neither is done.

> > So we need something like a jobServer recipe property to selectively enable the support. Otherwise we may break existing recipes.
> 
> I think this is indeed need. Otherwise existing recipes with `make -j $(nprocs)` produce a warning like: `make: warning: -j8 forced in submake: resetting jobserver mode.`
> 
> But I'm unsure if this should be a property or a policy?

It should be a property. It should be selectively enabled by the `make` class which can then refrain from passing `-j`. Some recipes require serial builds for their packages so they need to disable job mode explicitly. A policy would be too broad I guess.

rhubert:
OK, I reworked the semaphore to avoid blocking and added a property which now affects checkout, build and package step in the same way. Not sure if this is a good idea :thinking: 

> I'm a bit confused about the thread that shuffles bytes out and into the pipe. I don't think this is necessary.

I'm open for other ideas.... :grinning: 

> I'm not sure what should happen if the recipes define `MAKEFLAGS` or if Bob is called from a recursive make where it's already set.

Append `--jobserver-auth` to the existing `MAKEFLAGS` envirionment? Maybe we can also check if  `jobserver-auth` is already in `MAKEFLAGS` and throw an error in this case? 

If bob is called from a recursive make as a user I would assume the provided fd's are used and passed. But is this really a use case or just something to note in the documentation?

pym/bob/input.py
```diff
@@ -1559,6 +1567,7 @@ def __init__(self, recipe, tools, diffTools, sandbox, diffSandbox,
         self.states = states
         self.pkgId = pkgId
         self.fingerprintMask = fingerprintMask
+        self.jobServer = useJobServer
```
jkloetzke, line 1570: There is no need to store the value in the package because it will always be the same as the recipe value. Just access `self.recipe...` to get the value.

pym/bob/cmds/build/builder.py
```diff
@@ -104,6 +108,100 @@ class RestartBuildException(Exception):
...
+        os.write(self.__w_server, bytes(jobs))
+        self.thread.start()
+
+    def getMakeFd(self):
+        return [ self.__r_make, self.__w_make ]
+
+    def stop(self):
+        self.__running = False
+        os.write(self.__w_make, bytes(0x30))
+        self.thread.join()
```
jkloetzke, line 130: Why not just one pipe, load it with the initial bytes and we're good?

pym/bob/cmds/build/builder.py
```diff
@@ -104,6 +108,100 @@ class RestartBuildException(Exception):
...
+
+
+class JobServerSemaphore:
+    def __init__(self, value, fds):
+        if sys.platform == "win32":
```
jkloetzke, line 135: I would make the distinction in `Builder.cook()` on Windows we create the `asyncio.BoundedSemaphore ` as before, otherwise the new `JobServerSemaphore`. Then this class here does not need to make this distinction and will get smaller.

pym/bob/cmds/build/builder.py
```diff
@@ -104,6 +108,100 @@ class RestartBuildException(Exception):
...
+            self.__sem = asyncio.BoundedSemaphore(value)
+        else:
+            self.__sem = asyncio.Semaphore(0)
+            self.__lock = threading.Lock()
+            self.__waitersCnt = 0
+            self.__jobsAquired = 0
+            self.__fds = fds
+            fcntl.fcntl(self.__fds[0], fcntl.F_SETFL,
+                    fcntl.fcntl(self.__fds[0], fcntl.F_GETFL) | os.O_NONBLOCK)
```
jkloetzke, line 144: `os.set_blocking`

pym/bob/cmds/build/builder.py
```diff
@@ -104,6 +108,100 @@ class RestartBuildException(Exception):
...
+    async def acquire(self):
+        if sys.platform == "win32":
+            await self.__sem.acquire()
+        else:
+            ret = None
+            try:
+                ret = os.read(self.__fds[0], 1)
+            except BlockingIOError:
+                pass
+            if not ret:
```
jkloetzke, line 169: This will evaluate to `True` if `ret == 0`. Did you mean `ret is None`?

pym/bob/cmds/build/builder.py
```diff
@@ -33,6 +33,10 @@
 import stat
 import sys
 import tempfile
+import threading
+
+if sys.platform is not "win32":
```
jkloetzke, line 38: While it will technically work should compare by value (`==`) and not by identity.

pym/bob/cmds/build/builder.py
```diff
@@ -104,6 +108,100 @@ class RestartBuildException(Exception):
...
+        try:
+            ret = os.read(self.__fds[0], 1)
+        except BlockingIOError:
+            pass
+        if ret:
+            with self.__lock:
+                self.__waitersCnt -= 1
+                if self.__waitersCnt == 0:
+                    asyncio.get_event_loop().remove_reader(self.__fds[0])
+            self.__sem.release()
```
jkloetzke, line 158: What about this?

```suggestion
        try:
            while __waitersCnt:
                os.read(self.__fds[0], 1)
                self.__waitersCnt -= 1
                self.__sem.release()
        except BlockingIOError:
            pass
        if self.__waitersCnt == 0:
            asyncio.get_event_loop().remove_reader(self.__fds[0])
```

What is still not right is that we have to remember the values that were extracted from the pipe and use them when writing back to the pipe.

jkloetzke:
> OK, I reworked the semaphore to avoid blocking and added a property which now affects checkout, build and package step in the same way. Not sure if this is a good idea thinking

I don't see a problem with this at the moment. We can always add more fine grained options afterwards.

> > I'm not sure what should happen if the recipes define `MAKEFLAGS` or if Bob is called from a recursive make where it's already set.
> 
> Append `--jobserver-auth` to the existing `MAKEFLAGS` envirionment? Maybe we can also check if `jobserver-auth` is already in `MAKEFLAGS` and throw an error in this case?

I would just append it. If the user defines `MAKEFLAGS` with something incompatible then it's his fault.

> If bob is called from a recursive make as a user I would assume the provided fd's are used and passed. But is this really a use case or just something to note in the documentation?

We could just use the fds instead of creating a pipe on our own. This would imply that we somehow have to cope with the implicit token that Bob will get, though.

I've skimmed through the changes. It's no exhaustive review but this is definitely going in the right direction...

pym/bob/cmds/build/builder.py
rhubert: Haven't tested this but I think it will not work as os.read is not blocking? So when ever a token is back in the pipe all waiters are released, aren't they? At least the return of `os.read` must be checked?

jkloetzke: My assumption was that either we get exactly one byte from the pipe (and hence we got a ticket) or `read` will throw `BlockingIOError` because the pipe was empty and the call would have blocked. If this assumption does not hold then the code above is indeed garbage.

pym/bob/cmds/build/builder.py
```diff
@@ -126,6 +129,79 @@ def stop(self):
...
+                    os.write(self.__fds[1], bytes(1))
+                    self.__jobsAquired -= 1
+
+    async def __aexit__(self, exc_type, exc, tb):
+        if sys.platform == "win32":
+            await self.__sem.__aexit__(exc_type, exc, tb)
+        else:
+            with self.__lock:
+                if self.__waitersCnt != 0:
+                    self.__waitersCnt = -1;
```
rhubert, line 198: :facepalm: :facepalm: :facepalm: :facepalm: 
it took me several hours to find the reason why bob sometimes didn't finish... 
:angry: :angry: :angry: :angry:

rhubert:
Thanks for the review - again a lot of code could be removed :) 

The use of make provided file descriptors is still missing. 
 > Even make will only forward these if it understands that the callee understands the job protocol (see Job Slots). 

Do you know how make makes this decision? If make foes not understand bob understands the job protocol we don't need to implement this? 

I'm also thinking about a test for this but I have no idea how to count the running make jobs. ATM I'm just using a Makefile which starts a few sleep's and monitor top...

jkloetzke:
> Thanks for the review - again a lot of code could be removed :)

You're welcome. :)

> The use of make provided file descriptors is still missing.
> 
> > Even make will only forward these if it understands that the callee understands the job protocol (see Job Slots).
> 
> Do you know how make makes this decision? If make foes not understand bob understands the job protocol we don't need to implement this?

AFAICT it consideres invocations of "make" or "$(MAKE)" and every command that is prefixed by a "+". I could imagine someone writes a recursive makefile where Bob is only a part of it. So I think it's worth the effort.

> I'm also thinking about a test for this but I have no idea how to count the running make jobs. ATM I'm just using a Makefile which starts a few sleep's and monitor top...

Hmm, a black box test will be difficult. I would just do a smoke test there. Instead a unit test may be a better approach where the relevant parts can be mocked.

pym/bob/input.py
```diff
@@ -2212,6 +2225,9 @@ def coDet(r):
             self.__relocatable = self.__recipeSet.getPolicy('allRelocatable') \
                 or not self.__provideTools
 
+        if self.__jobServer is None:
```
jkloetzke, line 2228: I'm sure there was class inheritance code before. Looks like it got lost in the last refactoring. Currently the property will not work when set in a (anonymous) class.

pym/bob/invoker.py
```diff
@@ -177,6 +178,10 @@ async def __runCommand(self, args, cwd, stdout=None, stderr=None,
             _env.update(env)
             env = _env
 
+        if len(self.__makeFds) == 2:
+            env["MAKEFLAGS"] = "".join([env.get("MAKEFLAGS") if env.get("MAKEFLAGS") is not None else ""]) + \
+                "-j --jobserver-auth=" + ",".join([str(fd) for fd in self.__makeFds])
```
jkloetzke, line 183: Shouldn't there be a space between multiple options? I also thought that make forwards the number of jobs, e.g. `-j4` instead of an unbounded `-j`.

pym/bob/invoker.py
```diff
@@ -305,6 +307,12 @@ async def executeStep(self, mode, clean=False, keepSandbox=False):
             elif clean and mode != InvocationMode.SHELL:
                 emptyDirectory(self.__spec.workspaceWorkspacePath)
 
+            if len(self.__makeFds) == 2:
+                self.__spec.env["MAKEFLAGS"] = ("".join([self.__spec.env.get("MAKEFLAGS") if
+                        self.__spec.env.get("MAKEFLAGS") is not None else ""])
+                    + " -j" + str(self.__makeJobs) + " --jobserver-auth="
+                    + ",".join([str(fd) for fd in self.__makeFds]))
```
jkloetzke, line 314: Hmm, I've played a bit around and it looks like `make` will parse the `MAKEFLAGS` that it gets, forwards some of the options it knows and overwrites the rest. Maybe we should do the same?

rhubert: I don't think we should remove or edit the `MAKEFLAGS` defined in the recipe or somewhere else? What sounds reasonable is to overwrite `-j` and `--jobserver-auth` if they are already in the `MAKEFLAGS`?

rhubert:
I added some code to make use of `MAKEFLAGS` provided to bob. I'm not sure if it's correct to use `-j` from `MAKEFLAGS` instead of `-j` from cmdline in case we have both?

pym/bob/invoker.py
jkloetzke: I'm not sure yet how to handle it properly. By default I would probably only forward options that are understood by Bob. For example if you define variables on the command line, e.g. `make -j3 FOO=bar`, you will get `MAKEFLAGS= -j3 --jobserver-auth=3,4 -- FOO=bar`. That will influence the build and we should probably strip the assignment out.

OTOH you are probably right that we should not overwrite what a user did specify in the recipe or on the command line.

So maybe we should just use `-j` and `--jobserver-auth` from the OS environment. But if the user specified something in the recipes then we amend `MAKEFLAGS` and keep the rest...

jkloetzke:
> I added some code to make use of `MAKEFLAGS` provided to bob. I'm not sure if it's correct to use `-j` from `MAKEFLAGS` instead of `-j` from cmdline in case we have both?

I made a small test and if I pass `-j2` to a sub-make when the top-make is run with `-j3` then I get the following:

```
$ LANG= make -j3 FOO=bar
sh -c 'echo "top: MAKEFLAGS=$MAKEFLAGS"'
top: MAKEFLAGS= -j3 --jobserver-auth=3,4 -- FOO=bar
make -C subdir -j2
make[1]: warning: -j2 forced in submake: resetting jobserver mode.
make[1]: Entering directory '/home/jkloetzke/tmp/subdir'
sh -c 'echo "sub: MAKEFLAGS=$MAKEFLAGS"'
sub: MAKEFLAGS=w -j2 --jobserver-auth=5,6 -- FOO=bar
make[1]: Leaving directory '/home/jkloetzke/tmp/subdir'
```

I think we should do the same, including the warning.

rhubert:
:bellhop_bell:

pym/bob/cmds/build/build.py
```diff
@@ -156,6 +158,20 @@ def _downloadArgument(arg):
...
+        if makeFlags is not None:
+            jobs = re.search(r'-j([0-9]*)', makeFlags)
+            fds = re.search(r'--jobserver-auth=([0-9]*),([0-9]*)', makeFlags)
+            if jobs and fds and jobs.group(1) and fds.group(1) and fds.group(2):
+                if noJobs:
+                    args.jobs = int(jobs.group(1))
+                    makeFds = [int(fds.group(1)), int(fds.group(2))]
+                else:
+                    print("warning: -j" + str(args.jobs) + " forced: resetting jobserver mode.");
+
```
jkloetzke, line 174: I think we have to check if the file descriptors are actually open. If the `bob` call is not prefixed with `+` in the `Makefile` then `make` will still set `MAKEFLAGS` but the file descriptors are not open. Something along these lines should work (untested):

```python
try:
  if not all(stat.S_ISFIFO(os.stat(f).st_mode) for f in makeFds): makeFds = None
except OSError:
  makeFds = None
```

pym/bob/cmds/build/builder.py
```diff
@@ -104,6 +104,79 @@ class RestartBuildException(Exception):
...
     pass
 
+class ExternalJobServer:
+    def __init__(self, makeFds):
+        self.__makeFds = makeFds
+
+    def getMakeFd(self):
+        return self.__makeFds
+
+class JobServer:
```
jkloetzke, line 114: Maybe name it `InternalJobServer` for consistency?

```suggestion
class InternalJobServer:
```

test/test_cmd_build_builder_jobserver.py
```diff
@@ -0,0 +1,34 @@
...
+
+class TestJobserverTest (TestCase):
+    def setUp (self):
+        self.__t = TestJobServer()
+
+    def testAcquire(self):
+        _run(self.__t.testAcquire())
+        self.__t.release()
+        _run(self.__t.testAcquire(2))
+        self.__t.release()
```
jkloetzke, line 30: Is the release deliberately unbalanced?

jkloetzke (commented):
Sorry. I didn't see that you've already pushed weeks ago. :see_no_evil: Anyway, just some last minor comments...

rhubert:
Don't stress. I was also away a couple of weeks ... :)

I pushed a new commit for the latest issues. If it's ok I'll squash them..

jkloetzke:
> Don't stress. I was also away a couple of weeks ... :)

:laughing: 

> I pushed a new commit for the latest issues. If it's ok I'll squash them..

Looks good. :+1:  I'll merge after the squash...

jkloetzke:
:tada: Thanks!

PR #336, opened 2020-05-25
scm/url: Include User-Agent

Ferruck:
Apparently, some hosts filter out the User-Agent set by urllib by
default, so we should include our own. For debugging purposes it's
probably a good idea to also include the current version, to identify
misbehaving changes.

The host this issue aroused with in this particular case is
https://swupdate.openvpn.org/, but of course this could happen
anywhere else, too.

### DIFF
--- a/pym/bob/scm/url.py
+++ b/pym/bob/scm/url.py
@@ -1,5 +1,5 @@
 # Bob build tool
-# Copyright (C) 2016  Jan Klötzke
+# Copyright (C) 2016-2020 The BobBuildTool Contributors
 #
 # SPDX-License-Identifier: GPL-3.0-or-later
 
--- a/pym/bob/scm/url.py
+++ b/pym/bob/scm/url.py
@@ -3,6 +3,7 @@
 #
 # SPDX-License-Identifier: GPL-3.0-or-later
 
+from .. import BOB_VERSION
 from ..errors import BuildError, ParseError
 from ..stringparser import IfExpression
 from ..utils import asHexStr, hashFile, isWindows
@@ -207,6 +208,7 @@ class UrlScm(Scm):
         signal.signal(signal.SIGINT, signal.SIG_DFL)
 
         headers = {}
+        headers["User-Agent"] = "BobBuildTool/{}".format(BOB_VERSION)
         context = None if self.__sslVerify else ssl.SSLContext(ssl.PROTOCOL_SSLv23)
         if os.path.isfile(destination) and self.__url.startswith("http"):
             # Try to avoid download if possible
--- a/pym/bob/archive.py
+++ b/pym/bob/archive.py
@@ -1,5 +1,5 @@
 # Bob build tool
-# Copyright (C) 2016  Jan Klötzke
+# Copyright (C) 2016-2020 The BobBuildTool Contributors
 #
 # SPDX-License-Identifier: GPL-3.0-or-later
 
--- a/pym/bob/archive.py
+++ b/pym/bob/archive.py
@@ -19,6 +19,7 @@ it must not be overwritten. The backend should make sure that even on
 concurrent uploads the artifact must appear atomically for unrelated readers.
 """
 
+from . import BOB_VERSION
 from .errors import BuildError
 from .tty import stepAction, stepMessage, \
     SKIPPED, EXECUTED, WARNING, INFO, TRACE, ERROR, IMPORTANT
@@ -572,7 +573,8 @@ class SimpleHttpArchive(BaseArchive):
     def __openDownloadFile(self, buildId, suffix):
         connection = self._getConnection()
         url = self._makeUrl(buildId, suffix)
-        connection.request("GET", url)
+        connection.request("GET", url,
+                headers={ 'User-Agent' : 'BobBuildTool/{}'.format(BOB_VERSION) })
         response = connection.getresponse()
         if response.status == 200:
             return SimpleHttpDownloader(self, response)
@@ -596,7 +598,8 @@ class SimpleHttpArchive(BaseArchive):
         url = self._makeUrl(buildId, suffix)
 
         # check if already there
-        connection.request("HEAD", url)
+        connection.request("HEAD", url,
+                headers={ 'User-Agent' : 'BobBuildTool/{}'.format(BOB_VERSION) })
         response = connection.getresponse()
         response.read()
         if response.status == 200:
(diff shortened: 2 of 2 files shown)


pym/bob/scm/url.py
```diff
@@ -203,10 +203,13 @@ def getProperties(self):
         return ret
 
     def _download(self, destination):
+        from .. import BOB_INPUT_HASH
```
jkloetzke, line 206: No need to lazily import it. You can move that to the import block at the top of the file.

pym/bob/scm/url.py
```diff
@@ -203,10 +203,13 @@ def getProperties(self):
...
+
         # restore signals to default so that Ctrl+C kills us
         signal.signal(signal.SIGINT, signal.SIG_DFL)
 
         headers = {}
+        headers["User-Agent"] = "BobBuildTool (version {})".format(BOB_INPUT_HASH)
```
jkloetzke, line 212: Shouldn't we use the real version here?

```suggestion
        headers["User-Agent"] = "BobBuildTool (version {})".format(BOB_VERSION)
```

Ferruck:
> Nice catch. I'm wondering if we should not do the same thing in `pym/bob/archive.py`..? 🤔

I think so, yes, but it's probably not that urgent, since you're controlling both ends of the pipeline here most of the time.

pym/bob/scm/url.py
Ferruck: So this really lazy-imports? I thought this would be more about namespacing... 🤔 😁  Will do that, of course.

pym/bob/scm/url.py
Ferruck: Yes, probably. 😂 TBH, I didn't really grasp what you were doing in that `__init__.py` file...

pym/bob/scm/url.py
Ferruck: When thinking about it again, it's actually quite obvious that this lazy imports, given that Python is interpreted... 🤦

Ferruck:
I noticed a really slow connection to sourceforge with my last test in the office, maybe this is also related to this issue here. I will investigate that further.

Ferruck:
I've include the change into `archive.py`, as well. For now, I left the cURL calls as they were, since I was unsure how to handle them.
I've tested those new changes with our local HTTP archive using the unit tests within the `archive` folder.

Ferruck:
The problem with sourceforge seems unrelated for now, trying to download something from there fails with cURL, too.

Ferruck:
> I've tested those new changes with our local HTTP archive using the unit tests within the archive folder.

I have to retract that statement, I was not aware of your JIT `default.yaml` generation. I've now re-run the test with the correct backend and ended up with:

```
>> root-bravo                                                                                                                                                   
   FNGRPRNT  root-bravo .. 'Bob'                                                                                                                                
   DOWNLOAD  work/root-bravo/dist/1/workspace .. ok                                                                                                             
Build result is in work/root-bravo/dist/1/workspace                                                                                                             
Duration: 0:00:00.392519, 0 checkouts (0 overrides active), 0 packages built, 1 downloaded.                                                                     
Expected command to fail: run_bob build --download=forced root-bravo                                                                                            
ERROR: Command return 1 failed with code 1
```

Is this expected behavior?

As I side note, I am unable to use `bob archive clean` with that backend, the `-v` switch isn't really helpful, actually...

jkloetzke:
> > I've tested those new changes with our local HTTP archive using the unit tests within the archive folder.
> 
> I have to retract that statement, I was not aware of your JIT `default.yaml` generation. I've now re-run the test with the correct backend and ended up with:
> 
> ```
> >> root-bravo                                                                                                                                                   
>    FNGRPRNT  root-bravo .. 'Bob'                                                                                                                                
>    DOWNLOAD  work/root-bravo/dist/1/workspace .. ok                                                                                                             
> Build result is in work/root-bravo/dist/1/workspace                                                                                                             
> Duration: 0:00:00.392519, 0 checkouts (0 overrides active), 0 packages built, 1 downloaded.                                                                     
> Expected command to fail: run_bob build --download=forced root-bravo                                                                                            
> ERROR: Command return 1 failed with code 1
> ```
> 
> Is this expected behavior?

What did you do exactly? The unit tests passed for your latest changes...

> As I side note, I am unable to use `bob archive clean` with that backend, the `-v` switch isn't really helpful, actually...

That's because `bob archive` only works locally where the artifacts are stored. It's poorly documented in the manpage.

Ferruck:
> What did you do exactly? The unit tests passed for your latest changes...

Since I wanted to test my changes against a real server, I `cd`'d to the `test/archive`, modified the `run.sh` to produce a `default.yaml` that uses our backend as follows:

```yaml
archive:
  -
    backend: http
    url: http://192.168.2.200/bob/archive/
```

and ran `run.sh`. As I mentioned above, that was not the first test, I already ran it without changes to the `run.sh` so it used the local file backend, if that is of any relevance here.

But looking further below in `run.sh` with your second comment in mind, I suspect that running these unit tests against a remote archive will never work...

When I just run the unmodified tests within `archive` as well as `test_archive.py`, it works for me, too, but as I said, I'd have preferred to test against a real server.

> That's because bob archive only works locally where the artifacts are stored. It's poorly documented in the manpage.

Well, yes, indeed. Is this just because nobody cared to implement a remote `bob archive` or are there technical reasons?

jkloetzke:
> > What did you do exactly? The unit tests passed for your latest changes...
> 
> Since I wanted to test my changes against a real server, I `cd`'d to the `test/archive`, modified the `run.sh` to produce a `default.yaml` that uses our backend as follows:
> ...
> But looking further below in `run.sh` with your second comment in mind, I suspect that running these unit tests against a remote archive will never work...

No this is the black box test for the "bob archive" command. It won't work with remote repositories yet. The various archive up-/download backends are tested in the `test_archive.py` unit tests.

> When I just run the unmodified tests within `archive` as well as `test_archive.py`, it works for me, too, but as I said, I'd have preferred to test against a real server.

Then we first have to implement that in `bob archive`.

> > That's because bob archive only works locally where the artifacts are stored. It's poorly documented in the manpage.
> 
> Well, yes, indeed. Is this just because nobody cared to implement a remote `bob archive` or are there technical reasons?

There was no need yet. Until now it was sufficient to have a cron-job on the artifact server that works locally. Technically it is possible to implement that but it will require the HTTP endpoint to be a WebDAV resource (needs `PROPFIND` to list the directories). The `shell` backend might need a `ls` command too.

jkloetzke:
Thanks! :+1:

PR #342, opened 2020-06-03
Small bob archive improvements

### LINKED ISSUE #339: bob-archive man page needs clarification
> As I side note, I am unable to use `bob archive clean` with that backend, the `-v` switch isn't really helpful, actually...

That's because `bob archive` only works locally where the artifacts are stored. It's poorly documented in the manpage.

_Originally posted by @jkloetzke in https://github.com/BobBuildTool/bob/pull/336#issuecomment-634872928_

Ferruck:
This adds both a warning message when there haven't been any archive files found and a `--fail` option analogous to `bob query-path`.

I'm a bit unsure whether or not the `--fail` option should be a bit more aggressive, though, for example while detecting "wrong" files. I am happily awaiting feedback on this point.

Fixes #339

### DIFF
--- a/pym/bob/cmds/archive.py
+++ b/pym/bob/cmds/archive.py
@@ -14,6 +14,7 @@ import pickle
 import pyparsing
 import re
 import sqlite3
+import sys
 import tarfile
 
 # need to enable this for nested expression parsing performance
@@ -77,6 +78,7 @@ class ArchiveScanner:
 
     def scan(self, verbose):
         try:
+            found = False
             self.__db.execute("BEGIN")
             for l1 in os.listdir("."):
                 if not self.__dirSchema.fullmatch(l1): continue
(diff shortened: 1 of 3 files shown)


pym/bob/cmds/archive.py
```diff
@@ -86,11 +88,16 @@ def scan(self, verbose):
...
                         if not m: continue
+                        found = True
                         self.__scan(os.path.join(l2, l3), verbose)
         except OSError as e:
             raise BobError("Error scanning archive: " + str(e))
         finally:
             self.__db.execute("END")
+            if verbose and not found:
+                print("Your archive seems to be empty. "
+                      "Are you running 'bob archive' from within the correct directory?")
```
jkloetzke, line 99: I guess that should go to `stderr` because it's nothing parseable...

Ferruck:
> Hmm, I'm not sure about the `--fail` option. What if the archive is really empty?

IMHO that's no problem since you have to explicitly enable `--fail`. It's meant to be used in scripts or alike where you have certain expectations, for example that there actually *is* something in your archive. I would agree if it'd be enabled by default...

> Being more aggressive is not an option at all. There are small `.buildid` and `.fprnt` files that are not handled but are created by Bob.

Sounds more like `bob archive` should handle these?!

In my opinion the *common* case is that you have a directory containing nothing more than Bob's archive and that you actually *want* it this way. Thus, Bob should report (and fail, if `--fail` is given) when there's something else lingering around.

pym/bob/cmds/archive.py
Ferruck: In fact that's what I had initially... :joy:

PR #352, opened 2020-06-26
Feature: Visual Studio Code Project Generator

mahaase:
Hi,

this is a initial commit for a Visual Studio Code Project generator. Mostly copied and pasted from qt-creator generator.
Looks okay for me so far.

WIP: The launch.json file is still missing.

Feedback welcome.


jkloetzke:
Nice. Unfortunately you picked the wrong template. :see_no_evil: Starting with the Visual Studio generator there is some common generator code (`pym/bob/generators/common.py`) that handles the dirty details. New generators should be based on this to remove duplicated code. I just did not have the time to port the old generators to the new framework. So please have a look on `VisualStudio.py` for a start.

Regarding the generated json files I would propose to use the standard Python `json` module. It will handle all the encoding details and should reduce the boilerplate code.

mahaase:
there is a new version in pipeline, based on the correct template. 
so I will close this one.

Issue #353, opened 2020-06-27
git: add support for git submodules


jkloetzke:
Submodules should basically be treated like the main module.

* Update them if the main module references a new commit but only if they are unmodified
* What should happen if a submodule disappears? What if a submodule switches the url?
* Audit trail updates
* Support `--shallow-submodules` cloning

rhubert:
- scmOverrides for submodules :roll_eyes:

jkloetzke:
Do you have an idea how this should look like?

rhubert:
For the `scmOverrides` I think we need a additional match property?

```yaml
scmOverrides:
 -
   match:
      url: repo
      submodule-path: submodule-foo
   del: [commit, tag]
   set: branch: foo
```

In addition to this I have a repo with many submodules where we need only one and thus forked only that to our servers. The current solution is a ` git submodule update --init <submodule-foo>` instead of `git submodule update --init`. Not sure how to handle this in the `git-scm`?

```yaml
checkoutSCM:
 - scm: git
   url: foo
   submodules-active:
      - !*
      - submodule-foo
``` 
????

jkloetzke:
Thanks for the input. I think it really makes sense to allow to clone a subset of the submodules. Maybe something like:

```yaml
checkoutSCM:
    - scm: git
      url: foo@bar.test
      submodules: True # recursively clone all submodules
    - scm: git
      url: something@else.test
      submodules:
        - dir
        - sub/dir
```

I would not allow wildcards, though. They are not supported natively by git and it would require some parsing on our side. Just passing the submodule path looks easy.

But I'm not so sure about the `scmOverrides`. Technically a master repository references only commits. Where the commits are coming from is recorded in the `.gitmodules` file. As far as Bob is concerned only commits will be fetched. If the user wants to checkout any branch in the submodule he has to do it manually anyway.

PR #363, opened 2020-09-06
Submodules

jkloetzke:
First shot for supporting submodules. This is way more heavy than I anticipated. @rhubert please have a look if there is something obvious missing. In principle only the `submodules` and `recurseSubmodules` properties need to be set and you're good. :crossed_fingers:


rhubert:
Found some time to test this today...Works in general, but cloning a subset as discussed in https://github.com/BobBuildTool/bob/issues/353#issuecomment-652584240  is not supported?

jkloetzke:
> Found some time to test this today...Works in general, but cloning a subset as discussed in [#353 (comment)](https://github.com/BobBuildTool/bob/issues/353#issuecomment-652584240) is not supported?

Uh, I skipped that part initially and then apparently I did not remember that you might need that. I guess this can be added with some small additions. If this feature is needed then just open a ticket and attach it to the 0.18 milestone...

PR #366, opened 2020-09-24
Add "show" command

jkloetzke:
The show command can be used to show all properties of a package as
they were computed by Bob when evaluating the recipes. The output can
be filtered and formatted in various ways. There is a special 'diff'
output format that produces a diff view of two packages.

- [x] Add documentation
- [x] Tests

### DIFF
--- a/contrib/bash-completion/bob
+++ b/contrib/bash-completion/bob
@@ -29,7 +29,8 @@ __bob_complete_dir()
                  compgen -d -P "$2" -S / -- "$1" ) )
 }
 
-__bob_commands="build dev clean graph help jenkins ls project status  query-scm query-recipe query-path query-meta"
+__bob_commands="build dev clean graph help jenkins ls project status  query-scm \
+                query-recipe query-path query-meta show"
 
 # Complete a Bob path
 #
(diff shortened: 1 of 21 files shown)


jkloetzke:
I think this is basically finished. If there are no objections or suggestions I would merge it...

rhubert:
:+1: I'm missing the recipe / packageName in the output?

jkloetzke:
Good point. I'll add it to the ``meta`` section...

rhubert:
:+1: looks good now. Only the `#0` appended to the `recipe` property of `checkoutSCM` is somehow strange / unexpected at first glace..

jkloetzke:
> Only the `#0` appended to the `recipe` property of `checkoutSCM` is somehow strange / unexpected at first glace..

That was one of the not so good ideas back at the beginning. It was introduced in ef840103a for `query-scm` to let scripts know where the SCM came from. It should have been called `origin` or something and the `#<num>` is pointless in the face of `multiPackage`. I guess we could remove it...

Ferruck:
Looks generally good, the only thing I'm missing is some kind of heading if you're querying multiple packages at once, e.g. by using the bobpaths syntax.

jkloetzke:
Good point. I think we could add a comment in the yaml and flat outputs. But json does not allow comments...

```yaml
---
# buildall/utils::e2fsprogs
buildTools:
    make: buildall/devel::make
    pkg-config: buildall/devel::pkg-config
    target-toolchain: buildall/devel::host-compat-toolchain
buildToolsWeak:
    make: buildall/devel::make
    pkg-config: buildall/devel::pkg-config
...
```

```
# buildall/utils::e2fsprogs
[19].buildTools.make=buildall/devel::make
[19].buildTools.pkg-config=buildall/devel::pkg-config
[19].buildTools.target-toolchain=buildall/devel::host-compat-toolchain
[19].buildToolsWeak.make=buildall/devel::make
[19].buildToolsWeak.pkg-config=buildall/devel::pkg-config
```

Ferruck:
Then maybe adding another property would be the more consistent variant? Something like `packagePath`?!

rhubert:
Not sure if anybody needs json / yaml output on stdout? Maybe it's a option to write them to a file where the filename matches the recipe name?

Ferruck:
Well, piping the output to something else is a valid use case, IMHO...

jkloetzke:
> Then maybe adding another property would be the more consistent variant? Something like `packagePath`?!

This is already there: `meta.package`. It's just no at the top because of the sorting...

> Not sure if anybody needs json / yaml output on stdout? Maybe it's a option to write them to a file where the filename matches the recipe name?

The output was intentionally thought to be parseable by other scripts. Therefore also the json and flat formats. And the yaml output naturally looks like the recipes so it was a good match too..

Ferruck:
Sorry, I missed `meta.package`. This is enough, at least for me. LGTM

PR #367, opened 2020-09-27
Add support for cloning a subset of submodules

jkloetzke:
Instead of cloning all submodules it is possible to give a list of paths
to the "submodules" property. Only these submodules will be cloned and
all others will be left uninitialized. This is not possible for
sub-submodules.

@rhubert Is this sufficient for your use case?

### DIFF
--- a/doc/manpages/bob-status.rst
+++ b/doc/manpages/bob-status.rst
@@ -117,12 +117,16 @@ Status codes can be interpreted as follows:
 - ``O`` = overridden. This SCM is affected by a
   :ref:`configuration-config-scmOverrides`. Pass ``--show-overrides`` to force
   the output of a ``STATUS`` line if the SCM is otherwise unmodified.
-- ``S`` = switched. The commit/tag/branch/URL is different from the recipe.
+- ``S`` = switched. The commit/tag/branch/URL is different from the recipe. In
+  case of submodules the current commit is different from the one tracked in
+  the parent tree.
 - ``U`` = unpushed commits on configured branch. Git only. Some commits were made
   locally to the configured branch but they have not yet been pushed to the remote.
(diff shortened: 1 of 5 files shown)


rhubert:
Thanks! Worked for my use-case :+1: 

`bob status` shows `unpushed local commits` for submodules?

with

```yaml
 root : True
 checkoutSCM:
  -
    scm: git
    url: https://git.codelabs.ch/muen.git
    branch: devel
    submodules:
        - tools/mugenschedcfg
        - tools/sbs

 buildScript: "true"
 packageScript: "true"
```

```
   STATUS u    dev/src/muen/1/workspace/.
   > submodule './tools/mugenschedcfg' unpushed local commits:
      fa04009 (grafted, HEAD) Minor improvements to plan plotting
   > submodule './tools/sbs' unpushed local commits:
      eebbc8f (grafted, HEAD) Extend sbs_inspect tool and provide stream verifier
```

jkloetzke:
> Thanks! Worked for my use-case +1

Cool.

> `bob status` shows `unpushed local commits` for submodules?

Not so cool. :wink: I'll look into it. This is certainly wrong.

jkloetzke:
> > `bob status` shows `unpushed local commits` for submodules?
> 
> Not so cool. wink I'll look into it. This is certainly wrong.

This is an artefact of the shallow and single-branch cloning. At least the `tools/mugenschedcfg` submodule checks out a commit from the "devel" branch while the default branch of the submodule is "master". Setting `shallowSubmodules: False` will "fix" the problem but it will clone much more.

In principle this problem can happen for regular checkouts with `shallow: 1` too when using `commit: ...`.

Not sure what is the best approach here. Suppress the unpushed commits check on shallow clones? Or keep it and amend the manpage?

rhubert:
I'd keep it and make a note in the manpage.

PR #373, opened 2020-10-27
builder: add option to force download of deps in different layers

danielroedl:
With --download=forced-deps-layers dependencies are forced to be
downloaded if they are in a different layer than the package
given on the command line.

### DIFF
--- a/doc/manpages/bob-build-dev.rst
+++ b/doc/manpages/bob-build-dev.rst
@@ -14,8 +14,8 @@ Options
 
     This option will just make sure that the sources of matching packages are
     checked out. Bob will still try to find matching binary artifacts to skip
-    the actual compilation of these packages. See the ``--download`` option
-    to control what is built and what is downloaded.
+    the actual compilation of these packages. See the ``--download`` and
+    ``--download-layer`` option to control what is built and what is downloaded.
 
 ``--audit``
     Generate an audit trail when building.
@@ -73,6 +73,22 @@ Options
       forced-deps
     packages=<packages regex>
       download modules that match a given regular expression, build all other.
+
+``--download-layer MODE``
+    Download from binary archive for layer (yes, no, forced)
+
+    Acts like ``--download`` but only for the modules of the layer that match a
+    given regular expression (``--download`` option will be overwritten for
+    matching modules).
+    Can be used multiple times (if regex is used also multiple times the last mode wins).
+    A sub layer is separated with a ``/``.
+
+    no=<layer regex>
+      build modules of a layer that match a given regular expression from sources
+    yes=<layer regex>
+      download modules of a layer that match a given regular expression, if download fails - build it from sources
+    forced=<layer regex>
+      like 'yes' above, but fail if any download fails
 
 ``--incremental``
     Reuse build directory for incremental builds.
@@ -203,7 +219,7 @@ Options
     current workspace. Because of the pipe-usage many tools like gcc,
     ls, git detect they are not running on a tty and disable output
     coloring. Disable the logfile generation to get the colored output
-    back. 
+    back.
 
 ``-p, --with-provided``
     Build provided dependencies too. In combination with ``--destination`` this
--- a/pym/bob/cmds/build/builder.py
+++ b/pym/bob/cmds/build/builder.py
@@ -267,6 +267,7 @@ cd {ROOT}
         self.__archive = DummyArchive()
         self.__downloadDepth = 0xffff
         self.__downloadDepthForce = 0xffff
+        self.__downloadLayerModes = []
         self.__downloadPackages = None
         self.__bobRoot = bobRoot
         self.__cleanBuild = cleanBuild
@@ -317,6 +318,18 @@ cd {ROOT}
         else:
             assert mode == 'no'
             self.__archive.wantDownload(False)
+
+    def setDownloadLayerMode(self, modes):
+        self.__downloadLayerModes = []
+        self.__archive.wantDownload(True)
+        for mode in modes:
+            regex = ""
+            modeKey, _, modeLayer = mode.partition("=")
+            try:
+                regex = re.compile(modeLayer)
+            except re.error as e:
+                raise BuildError("Invalid download layer regex '{}': {}".format(e.pattern, e))
+            self.__downloadLayerModes.append((regex, modeKey))
 
     def setUploadMode(self, mode):
         self.__archive.wantUpload(mode)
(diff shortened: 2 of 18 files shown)


jkloetzke:
I'm pondering about the feature since yesterday and my gut feeling tells me that a more general approach may be even better. Could you elaborate a bit more on the use case? Maybe it helps to get the semantics right.

Technically I think the layer specification is almost orthogonal to the current depth-based download specification. I could also image the use case where one wants to build everything from one or more layers and (force) download from another layer(s). So maybe something like a `--download-layer <no/yes/force>=<regex>` switch that can be given multiple times is more appropriate? Your PR is then something like `--download=deps --download-layer yes= --download-layer force=".*"` (assuming you want to build the root layer and download every other layer). The `--download-layer` matches would be done in the order that is given and the first match determines the no/yes/force disposition.

The idea is not fully fleshed out because the interaction between `--download` and `--download-layer` is not really clear. Is it an `and` or an `or` composition? What happens if one says `no` and the other switch says `force` to a package?

What do you think? Would it still match your use case?

danielroedl:
The use case is that I want to develop a component of my main-project separately (in a let say component-tree) and use it in the main-project-tree. The new switch should prevent a rebuild the component in the main-project build process without knowledge of this (because e.g. testing was done in the develop process before).
If the artifact of this component is not found the build should fail. This could be the cause if configuration/toolchain/dependencies in the main-project-tree are different to the configuration/toolchain/dependencies in the component-tree.

A possible layer combination:
```
utility-tree --use--> toolchain-layer, utility-layer
component-tree --use--> toolchain-layer, component-layer, utility-layer
main-project-tree --use--> toolchain-layer, component-layer, utility-layer
```

Different versions of the utility-layer in the component-tree and main-project-tree can trigger a build of the component in the main-project build process.

I think the `--download-layer` option would also fit my use case. In addition it has the nice advantage that one can have multiple component layers in one project and define dynamically which layer should be built/downloaded.

For the combination of `--download` and `--download-layer` I would prefer that `--download-layer` overrides `--download` option for packages in the matching layers.

I try to implement and it and update the PR.

danielroedl:
I have implemented the `--download-layer` argument and some tests for it.

Please take a look and give me a feedback what I need to do additionally to get this feature ready for integration.

Thanks!

doc/manpages/bob-build-dev.rst
```diff
@@ -74,6 +74,22 @@ Options
...
       download modules that match a given regular expression, build all other.
 
+``--download-layer MODE``
+    Download from binary archive for layer (yes, no, forced)
+
+    Acts like ``--download`` but only for the modules of the layer that match a
+    given regular expression (``--download`` option will be overwritten for
+    matching modules).
+    Can be used multiple times (if regex is used also multiple times the last mode wins).
+    A sub layer is separated with a dot.
```
jkloetzke, line 84: I would rather separate layers with `/`. The dot is a valid character for a package, so this may lead to ambiguous matches. It's also a regex meta character that would need escaping to match precisely.

pym/bob/cmds/build/builder.py
```diff
@@ -318,6 +319,19 @@ def setDownloadMode(self, mode):
...
 
+    def setDownloadLayerMode(self, modes):
+        self.__downloadLayerModes = []
+        self.__archive.wantDownload(True)
+        for mode in modes:
+            regex = ""
+            m = re.match('^(\w+)(=(.*))?$', mode)
+            modeKey = m.group(1)
+            try:
+                regex = re.compile(m.group(3))
```
jkloetzke, line 330: No need for a regex. A `str.partition()` is sufficient:

```suggestion
            modeKey, _, modeLayer = mode.partition("=")
            try:
                regex = re.compile(modeLayer)
```

pym/bob/cmds/build/builder.py
```diff
@@ -1124,8 +1138,17 @@ async def _downloadPackage(self, packageStep, depth, packageBuildId):
         workspaceChanged = False
         wasDownloaded = False
         packageDigest = packageStep.getVariantId()
-        if depth >= self.__downloadDepth or (self.__downloadPackages and
-                self.__downloadPackages.search(packageStep.getPackage().getName())):
+        layer = ".".join(packageStep.getPackage().getRecipe().getLayer())
```
jkloetzke, line 1141:
```suggestion
        layer = "/".join(packageStep.getPackage().getRecipe().getLayer())
```

test/layer-download/config.yaml
```diff
@@ -0,0 +1,5 @@
+bobMinimumVersion: "0.16"
```
jkloetzke, line 1: Not that it makes a difference but new tests should use the current version:

```suggestion
bobMinimumVersion: "0.18"
```

test/layer-download/run.sh
```diff
@@ -0,0 +1,103 @@
...
+checkBuildDirExist 1 0 0 0 0
+checkDistDirExist 1 1 1 1 0
+
+cleanup
+run_bob build --download deps --download-layer no=componentA --download-layer no=componentB root
+checkBuildDirExist 1 1 0 0 0
+checkDistDirExist 1 1 1 1 0
+
+cleanup
+run_bob build --download deps --download-layer no=component.* root
```
jkloetzke, line 59: The `*` should be quoted to prevent some accidental matching by the shell of files:
```suggestion
run_bob build --download deps --download-layer no="component.*" root
```

test/layer-download/run.sh
```diff
@@ -0,0 +1,103 @@
...
+checkBuildDirExist 0 0 0 0 0
+checkDistDirExist 1 0 0 0 0
+
+cleanup
+run_bob build --download yes --download-layer no=componentA root/A::A
+checkBuildDirExist 1 0 0 0 0
+checkDistDirExist 1 0 0 1 0
+
+cleanup
+run_bob build --download yes --download-layer no=component.* root/A::A
```
jkloetzke, line 75: ditto

jkloetzke (changes requested):
This looks very good and is almost ready to merge. :+1: Just some minor nitpicks...

doc/manpages/bob-build-dev.rst
danielroedl: Changed.

pym/bob/cmds/build/builder.py
danielroedl: Changed.

test/layer-download/config.yaml
danielroedl: Changed.

test/layer-download/run.sh
danielroedl: Changed.

test/layer-download/run.sh
danielroedl: Changed.

pym/bob/cmds/build/builder.py
danielroedl: Changed.

danielroedl:
I don't get the point why the "codecov/patch" check failed and what I can do to get this working.

jkloetzke (approved):

jkloetzke:
> I don't get the point why the "codecov/patch" check failed and what I can do to get this working.

Never mind. This check is that the diff itself should have a coverage above the current project coverage value. Actually I don't care about this as long as the whole project coverage does not go down...

jkloetzke:
Thanks. 👍

PR #390, opened 2020-12-22
[RFC] doc: some notes about offline installations

rhubert:
how to install `bob` on a machine without internet access??

I haven't found a easy solution to do this but it's working with the steps described in the PR. :man_shrugging:


jkloetzke:
As I don't have a better idea either I'm inclined to merge this. Any objections?

PR #405, opened 2021-03-18
Local shared packages

jkloetzke:
This has been tested moderately. Even though it *should* work on Windows I did not test it there yet. A couple of test are also missing.

@mahaase @danielroedl This might be interesting for you I guess. Any testing and feedback is welcome.

### DIFF
--- a/pym/bob/cmds/build/share.py
+++ b/pym/bob/cmds/build/share.py
@@ -1,0 +1,314 @@
+# Bob build tool
+# Copyright (C) 2021  Jan Klötzke
+#
+# SPDX-License-Identifier: GPL-3.0-or-later
+
+from ...errors import BuildError
+from ...tty import Warn, WarnOnce
+from ...utils import asHexStr, hashDirectoryWithSize, isWindows
+import errno
+import os, os.path
+import json
+import shutil
+import sys
+import tempfile
+
+warnRepoSize = WarnOnce("The shared repository is over its quota. Run 'bob clean --shared' to free disk space!")
+warnGcDidNotHelp = WarnOnce("The automatic garbage collection of the shared repository was unable to free enough space. Run 'bob clean --shared' manually.")
+warnNoShareConfigured = Warn("No shared directory configured! Nothing cleaned.")
+
+if sys.platform == 'win32':
+    import msvcrt
+    def lockFile(fd, exclusive):
+        msvcrt.locking(fd.fileno(), msvcrt.LK_LOCK if exclusive else msvcrt.LK_RLCK, 0x100000)
+    def unlockFile(fd):
+        fd.seek(0)
+        msvcrt.locking(fd.fileno(), msvcrt.LK_UNLCK, 0x100000)
+else:
+    import fcntl
+    def lockFile(fd, exclusive):
+        fcntl.flock(fd, fcntl.LOCK_EX if exclusive else fcntl.LOCK_SH)
+    def unlockFile(fd):
+        fcntl.flock(fd, fcntl.LOCK_UN)
+
+# The first two generations were used by the Jenkins backend.
+SHARED_GENERATION = '-3'
+
+class OpenLocked:
+
+    def __init__(self, fileName, mode, exclusive):
+        self.fileName = fileName
+        self.mode = mode
+        self.exclusive = exclusive
+
+    def __enter__(self):
+        self.fd = open(self.fileName, self.mode)
+        try:
+            lockFile(self.fd, self.exclusive)
+        except:
+            self.fd.close()
+            raise
+        return self.fd
+
+    def __exit__(self, exc_type, exc_value, traceback):
+        try:
+            unlockFile(self.fd)
+        finally:
+            self.fd.close()
+
+
+def sameWorkspace(link, sharePath):
+    """Is the workspace shared and points to sharePath?
+
+    Depending on the OS the shared package could be a symlink to the shared
+    place or have a place-holder file with the path. The workspace could also
+    be something completely different by now.
+    """
+    try:
+        if os.path.islink(link):
+            dst = os.readlink(link)
+        elif os.path.isfile(link):
+            with open(link) as f:
+                dst = f.read(0x10000)
+        else:
+            return False
+        return os.path.samefile(dst, sharePath)
+    except OSError as e:
+        raise BuildError("Error inspecting workspace: " + str(e))
+
+def checkUnused(pkgMeta, pkgPath):
+    pkgWorkspace = os.path.join(pkgPath, "workspace")
+    return all((not sameWorkspace(user, pkgWorkspace)) for user in pkgMeta.get("users", []))
+
+class NullShare:
+    def __init__(self):
+        self.quota = 0
+
+    def remoteName(self, buildId):
+        return ""
+
+    def canInstall(self):
+        return False
+
+    def useSharedPackage(self, workspace, buildId):
+        return None, None
+
+    def installSharedPackage(self, workspace, buildId, sharedHash, mayMove):
+        return None, False
+
+    def gc(self, pruneUsed, pruneUnused, dryRun=False, progress=lambda x: None, newPkg=None):
+        warnNoShareConfigured.warn()
+        return None
+
+
+class LocalShare:
+    UNITS = [("KiB", 1024**1), ("MiB", 1024**2), ("GiB", 1024**3), ("TiB", 1024**4),
+             ("K",   1024**1), ("M",   1024**2), ("G",   1024**3), ("T",   1024**4),
+             ("KB",  1000**1), ("MB",  1000**2), ("GB",  1000**3), ("TB",  1000**4)]
+
+    def __init__(self, spec):
+        self.__path = os.path.abspath(os.path.expanduser(spec['path']))
+        quota = spec.get('quota')
+        if isinstance(quota, str):
+            for ending, factor in self.UNITS:
+                if quota.endswith(ending): break
+            else:
+                ending = None
+                factor = 1
+
+            if ending is not None:
+                quota = quota[:-len(ending)]
+            quota = int(quota) * factor
+        self.__quota = quota
+        self.__autoClean = spec.get('autoClean', True)
+
+    def __buildPath(self, buildId):
+        buildId = asHexStr(buildId) + SHARED_GENERATION
+        return os.path.join(*[self.__path, buildId[0:2], buildId[2:4], buildId[4:]])
+
+    def __addPackage(self, buildId, size):
+        def update(f):
+            meta = json.load(f)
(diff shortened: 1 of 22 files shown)


pym/bob/cmds/build/share.py
```diff
@@ -0,0 +1,296 @@
...
+                            pkgMeta = json.load(pf)
+                            pkgTime = os.fstat(pf.fileno()).st_mtime_ns
+                            pkgUnused = checkUnused(pkgMeta, pkgPath) and (pkgPath != newPkg)
+                        if pkgUnused or pruneUsed:
+                            candidates.append((pkgUnused, pkgTime, size, pkg))
+                    except FileNotFoundError:
+                        pass
+
+                # Move all candidates to the attic until we are under the quota
+                for pkgUnused, _, pkgSize, pkgBuildId in sorted(candidates):
```
Ferruck, line 277: The manpage states that packages are deleted *from oldest usage to newest*, now you're ignoring `pkgTime`. Is this intentional? Is the behaviour ensured implicitly?

Ferruck (approved):
Looks good to me. I'll test it today and report if I encounter any problems.

pym/bob/cmds/build/share.py
```diff
@@ -0,0 +1,296 @@
...
+import os, os.path
+import json
+import shutil
+import sys
+import tempfile
+
+warnRepoSize = WarnOnce("The shared repository is over its quota. Run 'bob clean --shared' to free disk space!")
+warnGcDidNotHelp = WarnOnce("The automatic garbage collection of the shared repository was unable to free enough space. Run 'bob clean --shared' manually.")
+
+if sys.platform == 'win32':
```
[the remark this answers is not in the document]
jkloetzke: Because `isWindows()` includes the MSYS platform where the `msvcrt` module is not available. MSYS does implement the `fcntl.flock()` function so it should work... :crossed_fingers:

pym/bob/cmds/build/share.py
jkloetzke: This is taken care of by the `sorted()` on the right side. It sorts the tuples by comparing the elements. So the unused packages go first, then they are sorted by time (and then size and buildid but who cares ;))...

Ferruck: Of course it does :exploding_head:

mahaase:
thank you! i will share it with employees and keep it in mind for future! 👍

sbixl:
Thanks for the feature! 😊 I will try to test it on windows at the next weekend.

jkloetzke:
I am thinking about to remove the new `--install` and `--shared` switches and move these settings into `default.yaml`. Is there a need to specify on the command line, per invocation, to use and/or install shared packages? I have the feeling that it would be sufficient to configure this per user or per project. Any feedback is welcome.

The reason is because there are already so many command line switches to `bob dev/build` that I'm reluctant to add more...

Ferruck:
> 
> 
> I am thinking about to remove the new `--install` and `--shared` switches and move these settings into `default.yaml`. Is there a need to specify on the command line, per invocation, to use and/or install shared packages? I have the feeling that it would be sufficient to configure this per user or per project. Any feedback is welcome.
> 
> The reason is because there are already so many command line switches to `bob dev/build` that I'm reluctant to add more...

Sounds reasonable, though I'd like to have a switch to temporarily *disable* sharing then, for example because I'm trying something in my workspace that I'd rather not see in the shared location...

jkloetzke:
> > I am thinking about to remove the new `--install` and `--shared` switches and move these settings into `default.yaml`. Is there a need to specify on the command line, per invocation, to use and/or install shared packages? I have the feeling that it would be sufficient to configure this per user or per project. Any feedback is welcome.
> > The reason is because there are already so many command line switches to `bob dev/build` that I'm reluctant to add more...
> 
> Sounds reasonable, though I'd like to have a switch to temporarily _disable_ sharing then, for example because I'm trying something in my workspace that I'd rather not see in the shared location...

Good point. I think I'll leave the switches.

pym/bob/cmds/build/share.py
```diff
@@ -0,0 +1,301 @@
...
+        # Prepare everyting in temporary directory next to the shared packages
+        # to atomically "install" the whole package with a single move. Can
+        # still lose the race at the final rename!
+        repoSize = 0
+        try:
+            os.makedirs(os.path.dirname(sharedPath), exist_ok=True)
+            with tempfile.TemporaryDirectory(dir=self.__path) as tmpDir:
+                tmpSharedPath = os.path.join(tmpDir, "pkg")
+                os.mkdir(tmpSharedPath)
+                shutil.copyfile(os.path.join(workspace, "..", "audit.json.gz"),
```
sbixl, line 200: If audit is disabled, the copy operation will fail. I guess we should add a checker if audit is active or not?

jkloetzke: I'm more inclined to prohibit installation of packages that were built without audit trail. Right now such packages are not uploaded either. Would that be a problem?

sbixl: Ok, that makes sense of course. I just stumbled across it while testing with disabled audit trail.

sbixl:
Hi Jan, I think I have found a little bug in the switching process between "shared" and "no-shared". I think it is because the two packages (tool4 and tool5) are both marked as shared and tool5 has a dependency to tool4. There may also be two issues see **3)** and **5)**. I did not test it on linux so maybe it is only a windows topic.

Steps to reproduce (native windows, no msys):

[repo.case.zip](https://github.com/BobBuildTool/bob/files/6255411/repo.case.zip)

Please set a proper path to the shared directory (its hardcoded)!

**1.)** `bob dev buildall`
**2.)** `bob dev buildall -b --no-shared`
**3.)** `bob clean --shared --dry-run` --> bob fails with (link to **tool1** why ever fails) :

Bob version 0.18.2.dev80+g445f54f
  File "d:\tools\bob\pym\bob\scripts.py", line 147, in catchErrors
    ret = fun(*args, **kwargs)
  File "d:\tools\bob\pym\bob\scripts.py", line 248, in cmd
    ret = cmd(args.args, bobRoot)
  File "d:\tools\bob\pym\bob\scripts.py", line 34, in __clean
    doClean(*args, **kwargs)
  File "d:\tools\bob\pym\bob\cmds\build\clean.py", line 164, in doClean
    lambda x: delPaths.append(x))
  File "d:\tools\bob\pym\bob\cmds\build\share.py", line 271, in gc
    pkgUnused = checkUnused(pkgMeta, pkgPath) and (pkgPath != newPkg)
  File "d:\tools\bob\pym\bob\cmds\build\share.py", line 73, in checkUnused
    return all((not sameWorkspace(user, pkgWorkspace)) for user in pkgMeta.get("users", []))
    return all((not sameWorkspace(user, pkgWorkspace)) for user in pkgMeta.get("users", []))
  File "d:\tools\bob\pym\bob\cmds\build\share.py", line 65, in sameWorkspace
    with open(link) as f:
PermissionError: [Errno 13] Permission denied: 'D:\\projects\\temp\\repo.case\\dev\\dist\\tool1\\1\\workspace'

**4.)** `bob dev buildall` --> As you can see all tools, except tool4 becomes shared again.
**5.)** `bob clean --shared --dry-run` --> bob fails with (now link to **tool4** fails) :

Bob version 0.18.2.dev80+g445f54f
  File "d:\tools\bob\pym\bob\scripts.py", line 147, in catchErrors
    ret = fun(*args, **kwargs)
  File "d:\tools\bob\pym\bob\scripts.py", line 248, in cmd
    ret = cmd(args.args, bobRoot)
  File "d:\tools\bob\pym\bob\scripts.py", line 34, in __clean
    doClean(*args, **kwargs)
  File "d:\tools\bob\pym\bob\cmds\build\clean.py", line 164, in doClean
    lambda x: delPaths.append(x))
  File "d:\tools\bob\pym\bob\cmds\build\share.py", line 271, in gc
    pkgUnused = checkUnused(pkgMeta, pkgPath) and (pkgPath != newPkg)
  File "d:\tools\bob\pym\bob\cmds\build\share.py", line 73, in checkUnused
    return all((not sameWorkspace(user, pkgWorkspace)) for user in pkgMeta.get("users", []))
  File "d:\tools\bob\pym\bob\cmds\build\share.py", line 73, in <genexpr>
    return all((not sameWorkspace(user, pkgWorkspace)) for user in pkgMeta.get("users", []))
  File "d:\tools\bob\pym\bob\cmds\build\share.py", line 65, in sameWorkspace
    with open(link) as f:
PermissionError: [Errno 13] Permission denied: 'D:\\projects\\temp\\repo.case\\dev\\dist\\tool4\\1\\workspace'

jkloetzke:
Thanks for testing. The crashes were unrelated to Windows and should (hopefully) be fixed...

jkloetzke:
Unless something else pops up I will probably merge this tomorrow evening and create 0.19.0-rc1...

sbixl:
The crashes are fixed now 👍! But there seems to be still a problem with the package "tool4" in [repo.case.zip](https://github.com/BobBuildTool/bob/files/6255411/repo.case.zip). 

1) If I build the project from scratch with 'bob dev buildall', all packages marked as "shared" are installed successfully to the shared location.
2) If I build the second time with 'bob dev buildall -b --no-shared', all packages marked as "shared" are unshared like expected.
3) If I build now with 'bob dev buildall -b', all packages **except tool4** are shared again.

Since package "tool4" is a direct dependency of "tool5", it looks like that dependencies are not forwarded to the shared package management. I have tried to narrow it down in the python code and it seems the missing trigger is already very high up in the function chain, since "_useSharedPackage" in "builder.py" does not see the packageStep of tool4.

sbixl:
If I run for example `bob clean --shared --all-unused`, all shared packages are deleted from the shared location like expected. 
However, a few artifacts remain in the shared location like e.g. ".bob.cache/**03/04**". Is this intentional, that the directories "**03/04**" are not deleted?

jkloetzke:
> Since package "tool4" is a direct dependency of "tool5", it looks like that dependencies are not forwarded to the shared package management.

This is the lazy evaluation of Bob that causes this behaviour. "tool4" is only used as build dependency of "tool5". But the package of "tool5" can be used from the shared repository. There is no need to build/use "tool4". Hence the workspace is not touched.

> If I run for example `bob clean --shared --all-unused`, all shared packages are deleted from the shared location like expected.
> However, a few artifacts remain in the shared location like e.g. ".bob.cache/**03/04**". Is this intentional, that the directories "**03/04**" are not deleted?

Yes, these empty directories are not removed. The first two levels of directories are there so that there are not too many packages in the same directory. The cleaning process does only delete the actual package directory. The problem is that a `bob clean --shared` and a `bob dev --install` could run concurrently on the same machine for the same top level directories. Keeping these empty directories makes installation easier because it does not need to cope with vanishing directories.

PR #417, opened 2021-06-07
python-3.9 + msys: platform string "msys" renamed by "cygwin"

### LINKED ISSUE #416: MSYS2: bob crashes with python3.9
```
  File "<string>", line 1, in <module>
  File "/usr/lib/python3.9/multiprocessing/forkserver.py", line 258, in main
    fds = reduction.recvfds(s, MAXFDS_TO_SEND + 1)
  File "/usr/lib/python3.9/multiprocessing/reduction.py", line 164, in recvfds
    raise RuntimeError('received %d items of ancdata' %
RuntimeError: received 0 items of ancdata
```

reason:
```
$ python3
Python 3.9.5 (default, May 15 2021, 20:34:13)
[GCC 10.2.0] on cygwin
Type "help", "copyright", "credits" or "license" for more information.
>>> import sys
>>> sys.platform
'cygwin'
```

looks like `msys` is renamed by `cygwin`.

fix1:

```
--- a/pym/bob/utils.py
+++ b/pym/bob/utils.py
@@ -641,7 +641,7 @@ class EventLoopWrapper:
             origSigInt = signal.getsignal(signal.SIGINT)
             signal.signal(signal.SIGINT, signal.SIG_IGN)

-            method = 'fork' if sys.platform == 'msys' else 'forkserver'
+            method = 'fork' if sys.platform in ('msys', 'cygwin') else 'forkserver'

             # Hack to get coverage data with the multiprocessing module. Up to
             # date coverage.py cannot trace cross fork/exec. Taken and adapted
```

what shall we do?! with the `BOB_HOST_PLATFORM`?

```
--- a/pym/bob/input.py
+++ b/pym/bob/input.py
@@ -3430,7 +3430,8 @@ class RecipeSet:
             env.update(self.__defaultEnv)
         env.setFuns(self.__stringFunctions)
         env.update(envOverrides)
-        env["BOB_HOST_PLATFORM"] = platform
+        env["BOB_HOST_PLATFORM"] = platform if platform != 'cygwin' else 'msys'
+        print(env["BOB_HOST_PLATFORM"])
         self.__rootEnv = env

         # resolve recipes and their classes
```

or we need another indicator in recipes to handle between linux and MSYS2-system?!

VS2019 generator:
```
--- a/pym/bob/generators/VisualStudio.py
+++ b/pym/bob/generators/VisualStudio.py
@@ -278,7 +278,7 @@ class Vs2019Generator(CommonIDEGenerator):

         # gather root paths
         bobPwd = Path(os.getcwd())
-        if sys.platform == 'msys':
+        if sys.platform in ('msys', 'cygwin'):
             if os.getenv('WD') is None:
                 raise BuildError("Cannot create Visual Studio project for Windows! MSYS2 must be started by msys2_shell.cmd script!")
             msysRoot = PureWindowsPath(os.getenv('WD')) / '..' / '..'
```
... in the `__init__.py` too. there are some more locations. 

shouldn't we use the `isWindows` function?

mahaase:
in MSYS2 with the update to python3.9 the platform string isn't "msys" anymore
the string is replaced by "cygwin".

we use isWindows or isMsys to check for the systems now.

.FIXES #416

### DIFF
--- a/pym/bob/generators/VisualStudio.py
+++ b/pym/bob/generators/VisualStudio.py
@@ -4,7 +4,7 @@
 # SPDX-License-Identifier: GPL-3.0-or-later
 
 from .common import CommonIDEGenerator
-from ..utils import quoteCmdExe
+from ..utils import quoteCmdExe, isMsys
 from pathlib import Path, PureWindowsPath
 from shlex import quote as quoteBash
 from uuid import UUID
@@ -278,7 +278,7 @@ class Vs2019Generator(CommonIDEGenerator):
 
         # gather root paths
         bobPwd = Path(os.getcwd())
-        if sys.platform == 'msys':
+        if isMsys():
             if os.getenv('WD') is None:
                 raise BuildError("Cannot create Visual Studio project for Windows! MSYS2 must be started by msys2_shell.cmd script!")
             msysRoot = PureWindowsPath(os.getenv('WD')) / '..' / '..'
@@ -331,7 +331,7 @@ class Vs2019Generator(CommonIDEGenerator):
 
     def __generateBuildme(self, extra, projectRoot, bobRoot):
         buildMe = []
-        if sys.platform == 'msys':
+        if isMsys():
             extra = " ".join(quoteBash(e) for e in extra)
             buildMe.append("#!/bin/sh")
             buildMe.append("export PATH=" + quoteBash(os.environ["PATH"]))
--- a/pym/bob/input.py
+++ b/pym/bob/input.py
@@ -13,7 +13,7 @@ from .state import BobState
 from .stringparser import checkGlobList, Env, DEFAULT_STRING_FUNS, IfExpression
 from .tty import InfoOnce, Warn, WarnOnce, setColorMode
 from .utils import asHexStr, joinScripts, compareVersion, binStat, \
-    updateDicRecursive, hashString, getPlatformTag, isWindows
+    updateDicRecursive, hashString, getPlatformTag, isWindows, getPlatformString
 from itertools import chain
 from os.path import expanduser
 from string import Template
@@ -3434,7 +3434,7 @@ class RecipeSet:
         else:
             return schema.validate(default)
 
-    def parse(self, envOverrides={}, platform=sys.platform):
+    def parse(self, envOverrides={}, platform=getPlatformString()):
         if not os.path.isdir("recipes"):
             raise ParseError("No recipes directory found.")
         self.__cache.open()
--- a/pym/bob/utils.py
+++ b/pym/bob/utils.py
@@ -15,6 +15,7 @@ import shutil
 import stat
 import struct
 import sys
+import sysconfig
 
 def hashString(string):
     h = hashlib.md5()
@@ -178,6 +179,19 @@ def compareVersion(origLeft, origRight):
                             .format(origLeft, origRight))
     return ret
 
+
+def getPlatformString():
+    return __platformString
+
+def isMsys():
+    return __isMsys
+
+if sys.platform.startswith('msys') or sysconfig.get_platform().startswith('msys'):
+    __isMsys = True
+    __platformString = 'msys'
+else:
+    __isMsys = False
+    __platformString = sys.platform
 
 def isWindows():
     """Check if we run on a windows platform.
@@ -641,7 +655,7 @@ class EventLoopWrapper:
             origSigInt = signal.getsignal(signal.SIGINT)
             signal.signal(signal.SIGINT, signal.SIG_IGN)
 
-            method = 'fork' if sys.platform == 'msys' else 'forkserver'
+            method = 'fork' if isWindows() else 'forkserver'
 
             # Hack to get coverage data with the multiprocessing module. Up to
             # date coverage.py cannot trace cross fork/exec. Taken and adapted
--- a/doc/manual/configuration.rst
+++ b/doc/manual/configuration.rst
@@ -667,6 +667,7 @@ the variable list:
 
   * ``linux``: Linux
   * ``msys``: Windows/MSYS2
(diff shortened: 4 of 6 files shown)


pym/bob/generators/VisualStudio.py
```diff
@@ -278,7 +278,7 @@ def generate(self, extra, bobRoot):
 
         # gather root paths
         bobPwd = Path(os.getcwd())
-        if sys.platform == 'msys':
+        if isMsys:
```
jkloetzke, line 281: You must actually call the function:

```suggestion
        if isMsys():
```

Also see the other changes with the same problem. You also forgot to import it so this will certainly fail at runtime.

pym/bob/input.py
```diff
@@ -3465,7 +3465,7 @@ def __parse(self, envOverrides, platform):
             env.update(self.__defaultEnv)
         env.setFuns(self.__stringFunctions)
         env.update(envOverrides)
-        env["BOB_HOST_PLATFORM"] = platform
+        env["BOB_HOST_PLATFORM"] = platform if not isMsys else 'msys'
```
jkloetzke, line 3468: This is the wrong spot. In case of Jenkins the `platform` variable describes the Jenkins architecture which might be different from the one where Bob is running. Actually the

```python
    def parse(self, envOverrides={}, platform=sys.platform):
```

above must be changed to something like

```python
    def parse(self, envOverrides={}, platform=getPlatformString()):
```

where `getPlatformString()` needs to be defined in utils.py.

pym/bob/utils.py
```diff
@@ -179,6 +179,9 @@ def cmp(l, r):
     return ret
 
 
+def isMsys():
```
jkloetzke, line 182: I think this is wrong because Cygwin is not exactly MSYS2. It should only return `True` if we're really running in MSYS2.

pym/bob/utils.py
```diff
@@ -641,7 +641,7 @@ def __init__(self):
             origSigInt = signal.getsignal(signal.SIGINT)
             signal.signal(signal.SIGINT, signal.SIG_IGN)
 
-            method = 'fork' if sys.platform == 'msys' else 'forkserver'
+            method = 'fork' if isMsys else 'forkserver'
```
jkloetzke, line 644: `isWindows()`?

jkloetzke (changes requested):
According to the [MSYS2 commit](https://github.com/msys2/MSYS2-packages/commit/2b65957fb905ae901cb6ac2fae6de6445dae7e05#diff-11a4a60b518bf24989d481468076e5d5982884626aed9faeb35b8576fcd223e1) that changed this we can still test for MSYS2 via `sysconfig.get_platform().startswith("msys")`. So if we find that `sys.platform` is `cygwin` the above test can be used to differentiate between a real Cygwin or a recent MSYS2. Let's hope they keep this scheme...

pym/bob/utils.py
mahaase: i have no idea, currently for win32 and linux we used 'forkserver' and just for msys we used 'fork' ?!
if i use `isWindows()` now, i will change the behavior.

pym/bob/generators/VisualStudio.py
mahaase: ohh?! i tested it successfully, so this was just luck? :D

pym/bob/utils.py
mahaase: i have no idea about cygwin, that's a msys, with a windows cross gcc, isn't it?

pym/bob/generators/VisualStudio.py
jkloetzke: A function is always True ;):

```
>>> def foo():
...   pass
... 
>>> bool(foo)
True
```

pym/bob/utils.py
jkloetzke: No. The `win32` case is already tested above. This leaves us either with a POSIX emulation on Windows (MSYS2 or Cygwin) or a native POSIX system. This was never tested on Cygwin but it should also use the `fork` method.

mahaase:
i hope i got it?! 

recheck pls :)

jkloetzke (commented):
Please have a look at the CI failures.

pym/bob/utils.py
```diff
@@ -241,17 +253,16 @@ def getPlatformTag():
...
         return __platformTag
 
-    p = sys.platform
-    if p == 'win32':
-        ret = b'w'
-    elif p in ('msys', 'cygwin'):
+    if isMsys() or isCygwin():
         ret = b'm'
+    elif isWindows():
+        ret = b'w'
```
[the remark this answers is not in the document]
mahaase: i tried to bring the inforation to a single point.

e.g. currently we check for native windows in different ways:
sys.platform == win32
os.name != posix

pym/bob/utils.py
```diff
@@ -215,7 +229,7 @@ def canSymlink():
     # it is possible to create symlinks. Try to create a symlink to see if we
     # have the privilege. Either the symlink() call fails directly or MSYS
     # silently creates a copy (unless MSYS=winsymlinks:nativestrict is set).
-    if sys.platform in ('msys', 'cygwin', 'win32'):
+    if isWindows():
```
jkloetzke, line 232: Either you revert this particular change too or you adapt the failing unit tests by mocking `isWindows()` in them.

jkloetzke (commented):
Looks good IMHO but there are still failing unit tests. I think I found the reason... (see below)

pym/bob/utils.py
mahaase: reverted.

jkloetzke (approved):

jkloetzke:
Thanks :)

PR #423, opened 2021-06-18
Python language support

jkloetzke:
This is the first draft to support Python as scripting language. Some parts are still missing and it's almost untested...

### DIFF
--- a/pym/bob/languages.py
+++ b/pym/bob/languages.py
@@ -157,7 +157,7 @@ class IncludeResolver(ABC):
         self.__incDigests = [ hashlib.sha1(origText.encode('utf8')).digest().hex() ]
         self.__incFiles = {}
 
-    def __nextIncludeFile(self, content):
+    def _nextIncludeFile(self, content):
         name = "{}_{}".format(self.varBase, self.count).translate(INVALID_CHAR_TRANS)
         self.count += 1
         self.__incFiles[name] = content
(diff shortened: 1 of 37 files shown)


pym/bob/languages.py
```diff
@@ -599,9 +567,150 @@ def setupFingerprint(spec, env):
...
+        return repr(content.decode('utf8'))
+
+
+class PythonLanguage:
+    index = ScriptLanguage.PYTHON
+    glue = "\nos.chdir(os.environ['BOB_CWD'])\n"
+    Resolver = PythonResolver
+
+    HELPERS = dedent("""\
+        from subprocess import run, call, check_call, check_output
```
Ferruck, line 584: Not sure if I'm fully understanding what's going on here, but it looks like these are inserted into the final script. If this is the case: do you really want to clutter every user-provided script with those functions? Especially `run` and `call` are pretty generic and might be used by the user. I'd prefer `import subprocess` like the ones below, even if that means more work here.

Ferruck (commented):
:+1:

pym/bob/languages.py
jkloetzke: Yeah, that's something where I'm not yet sure about. At least some minimal imports are required by the script prologue (os, os.path, sys). Almost any script *will* use the subprocess main functions anyway. I first found it convenient to implicitly import them always. But thinking more about this it might not be a good idea and leave this entirely to the user. That can be handled by some class(es) and `{checkout,build,package}SetupPython` snippets.

Issue #429, opened 2021-08-27
git default branch name


rhubert:
I just stumbled upon this when I tried to clone a newly created repo with `bob`. Looks like GitHub, Gitlab,.. changed the default name to `main`....

https://about.gitlab.com/blog/2021/03/10/new-git-default-branch-name/
https://github.com/github/renaming

What options do we have?
1. leave it `master`
2. remove `master` and use `HEAD`
3. policy based change from `master` to `main` / `HEAD`?
4. config-variable?
5. try `master` first, fallback to `main`?
6. combination of 2,3,(4) ?
7. ...?

I somehow prefer option 2. According to the [manual](https://nixos.org/manual/nix/stable/) (search for `builtins.fetchGit`) `nix` uses `HEAD` as default.

mahaase:
hmm, the HEAD thing i do not understand.

> The special symbol "HEAD" can always be used to refer to the current branch.
> https://git-scm.com/docs/user-manual.html

this would change the behavior in some cases. so i guess we should do the policy thing (3).

jkloetzke:
I kind of incline towards option 4. We could introduce an option in `config.yaml` or `default.yaml` that sets the default git branch globally. Of course it needs to stay at `master` by default to retain backwards compatibility.

The problem with option 2. and 5. is that you don't know in advance what branch you will get. It's also something that may change with time, depending on the server configuration. Some of the earliest versions of Bob did in fact use plain "git clone" if no branch was given. It turned our to be too hard to implement correctly because you will have to handle the case where the default branch on the server is changed and you have to update your working copy accordingly.

PR #433, opened 2021-09-24
git: add defaultBranchName

rhubert:
Fixes: #429

Not sure if this fits. The higher layer would override the setting of a lower layer... :/

### DIFF
--- a/doc/manual/configuration.rst
+++ b/doc/manual/configuration.rst
@@ -852,7 +852,7 @@ If the package is built from multiple modules you can give a list of SCMs::
             dir: src/bar
 
 There are three common (string) attributes in all SCM specifications: ``scm``,
-``dir`` and ``if``. Additionally to the string type, the ``if`` property may
+``dir`` (\*) and ``if``. Additionally to the string type, the ``if`` property may
 be given as IfExpression (see :ref:`configuration-principle-booleans`). By
 default the SCMs check out to the root of the workspace. You may specify any
 relative path in ``dir`` to checkout to this directory.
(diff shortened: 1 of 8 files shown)


jkloetzke:
> Not sure if this fits. The higher layer would override the setting of a lower layer... :/

I think it's the other way around. A sub-layer could override the setting of the upper layer. This is certainly not a good idea.

Given that we have `scmOverrides` in `default.yaml` it feels more natural to move this setting there. And while we're at it, we might add the possibility to change other defaults too. Maybe something like:

```yaml
scmDefaults:
    git:
        branch: main
        singleBranch: True
```

What do you think?

doc/manual/configuration.rst
```diff
@@ -876,31 +876,31 @@ cvs    CVS repository               | ``cvsroot``: repository location ("``:ext:
                                     | ``module``: module name
                                     | ``rev``: revision, branch, or tag name (optional)
 git    `Git`_ project               | ``url``: URL of remote repository
-                                    | ``branch``: Branch to check out (optional, default: master)
+                                    | ``branch`` (*): Branch to check out (optional, default: master)
```
jkloetzke, line 879: The single asterisk must be escaped => `(\*)`

doc/manual/configuration.rst
```diff
@@ -2079,6 +2079,14 @@ the anonymous access to the container can be used like this::
...
 in case other backends are configured too.
 
+scmDefaults
+~~~~~~~~~~~
+
+Type: Dict of SCM dicts
+
+Default settings for SCMs are applied if the value is not set in the
+recipe. Useable values are marked with `(*)` in (:ref:`configuration-recipes-scm`).
+
```
jkloetzke, line 2089: It would be good to have some example to see how the syntax looks exactly.

pym/bob/input.py
```diff
@@ -389,6 +389,13 @@ def Scm(spec, env, overrides, recipeSet):
         except schema.SchemaError as e:
             raise ParseError("Error validating SCM after applying scmOverrides: {}".format(str(e)))
 
+    # apply scmDefaults
+    defaults = recipeSet.scmDefaults().get(spec['scm'])
+    if defaults:
+        for d in defaults:
+            if not spec.get(d):
+                spec[d] = defaults[d]
```
jkloetzke, line 397: I think this is equivalent to:

```suggestion
    for k, v in recipeSet.scmDefaults().get(spec['scm'], {}).items():
        spec.setdefault(k, v)
```

I guess we'll also need some test that uses the feature.

jkloetzke (commented):
This looks quite good IMO. Just some minor comments...

pym/bob/input.py
rhubert: It's always amazing ... :+1:

doc/manual/configuration.rst
```diff
@@ -852,7 +852,7 @@ If the package is built from multiple modules you can give a list of SCMs::
             dir: src/bar
 
 There are three common (string) attributes in all SCM specifications: ``scm``,
-``dir`` and ``if``. Additionally to the string type, the ``if`` property may
+``dir`` (\*) and ``if``. Additionally to the string type, the ``if`` property may
```
jkloetzke, line 855: Looks like the `cvs` SCM still missing this change...

pym/bob/input.py
jkloetzke: :grinning:

doc/manual/configuration.rst
rhubert: you are right.. The unit test will also follow..

test/test_input_recipeset.py
```diff
@@ -1390,3 +1390,171 @@ def testNewBehaviour(self):
 
         pkg = self.generate().walkPackagePath("root")
         self.assertTrue(pkg.getCheckoutStep().getScmList()[0].getProperties(False)["prune"])
+
+
+class TestScmDefaults(RecipesTmp, TestCase):
+    """ Test behaviour of pruneImportScm policy"""
```
jkloetzke, line 1396: Copy'n'paste

test/test_input_recipeset.py
```diff
@@ -1390,3 +1390,171 @@ def testNewBehaviour(self):
...
+
+    def testScmDefaultsGit(self):
+        """Test scmDefaults for gitscm"""
+        with TemporaryDirectory() as tmp:
+            os.chdir(tmp)
+            os.mkdir("recipes")
```
jkloetzke, line 1402: No need to do this. The `RecipesTmp` base class does exactly this already.

test/test_input_recipeset.py
```diff
@@ -1390,3 +1390,171 @@ def testNewBehaviour(self):
...
+                   scm: git
+                   url: foo@bar
+                buildScript: "true"
+                packageScript: "true"
+                """)
+            pkg = self.generate().walkPackagePath("root")
+            self.assertEqual(pkg.getCheckoutStep().getScmList()[0].getProperties(False)["branch"],
+                    "master")
+            self.assertEqual(pkg.getCheckoutStep().getScmList()[0].getProperties(False)["sslVerify"],
+                    None)
```
jkloetzke, line 1416: How about this to reduce the redundancy?

```suggestion
            props = pkg.getCheckoutStep().getScmList()[0].getProperties(False)
            self.assertEqual(props["branch"], "master")
            self.assertEqual(props["sslVerify"], None)
```
...

test/test_input_recipeset.py
```diff
@@ -1390,3 +1390,171 @@ def testNewBehaviour(self):
...
+            self.assertEqual(pkg.getCheckoutStep().getScmList()[0].getProperties(False)["shallow"],
+                    None)
+            self.assertEqual(pkg.getCheckoutStep().getScmList()[0].getProperties(False)["singleBranch"],
+                    None)
+            self.assertFalse(pkg.getCheckoutStep().getScmList()[0].getProperties(False)["submodules"])
+            self.assertFalse(pkg.getCheckoutStep().getScmList()[0].getProperties(False)["recurseSubmodules"])
+            self.assertTrue(pkg.getCheckoutStep().getScmList()[0].getProperties(False)["shallowSubmodules"])
+
+            with open("default.yaml", "w") as f:
```
jkloetzke, line 1425: Maybe you could add a `writeDefault` (like the existing `writeConfig`) helper in the `RecipesTmp` class?

jkloetzke:
Thanks :+1:

PR #434, opened 2021-09-26
invoker: do not fail on environment variable collisions

### LINKED ISSUE #432: Option -E leads to collision regarding TMP environment variables on MSYS
When using `bob dev/build` with option `-E` (preserve environment) on MSYS with a current Bob version, e.g. 0.19.2, the following error message is thrown:

```
Error: Colliding environment variables: TEMP ~= temp, tmp ~= TMP
```

I use this option quite rarely, but in earlier Bob versions it worked fine. It seems to got broken in some of the newer versions.

jkloetzke:
On Windows environment variables are case-insensitive but
case-preserving. Unfortunately MSYS2 creates variables that only differ
in case (e.g. TMP vs. tmp). Remove the hard failure from the check and
convert it into a warning because Bob versions before 0.17 worked
without failure in this case. Also fix the code to catch all possible
collisions, including ones that are created by the recipes itself.

Fixes #432.

### DIFF
--- a/pym/bob/invoker.py
+++ b/pym/bob/invoker.py
@@ -95,6 +95,7 @@ class Invoker:
         self.__trace = trace
         self.__sandboxHelperPath = None
         self.__stdioBuffer = io.BytesIO() if redirect else None
+        self.__warnedDuplicates = { '' }
 
         # Redirection is a bit complicated. We have to consider two levels: the
         # optional log file and the console.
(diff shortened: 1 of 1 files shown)


gruberchr:
I checked the changes and invoking Bob with `-E` does not generate any failure or warning anymore. :+1:

I also checked, that adding a duplicated (lower case) environment variable in one of the recipes still generates a warning. Is it still desired, that the failure is converted to a warning as discussed in #432? Or should the warning be converted back to a failure after checking only the `definedEnvVars`?

Finally I had a look on the code and it looks fine to me.

jkloetzke:
Thanks for the review.

> I also checked, that adding a duplicated (lower case) environment variable in one of the recipes still generates a warning. Is it still desired, that the failure is converted to a warning as discussed in #432? Or should the warning be converted back to a failure after checking only the `definedEnvVars`?

I would keep this as a warning. Technically you could do this if you only call MSYS2 executables and it was working before Bob 0.17.

PR #436, opened 2021-09-30
Add out-of-tree build support

jkloetzke:
Adds a new "init" command which initializes a build tree that points to
the project root directory where the recipes are read from. The build
tree might have it's own default.yaml too.

```
bob init /path/to/recipes build
cd build
bob dev ...
```

Still missing:

- [ ] Tests
- [ ] Amend plugin API to get project root directory
- [ ] Documentation

Please give it a thorough testing...


mahaase:
great! in first tests that looks very well! 👍 
maybe we should add the autocompletion stuff.

PR #437, opened 2021-10-01
Git references

rhubert:
This build on top of #433 ...

### DIFF
--- a/doc/manual/configuration.rst
+++ b/doc/manual/configuration.rst
@@ -887,6 +887,17 @@ git    `Git`_ project               | ``url``: URL of remote repository
                                     | ``submodules`` (\*): Whether to clone all / a subset of submodules. (optional)
                                     | ``recurseSubmodules`` (\*): Recusively clone submodules (optional, defaults to false)
                                     | ``shallowSubmodules`` (\*): Clone submodules shallowly (optional, defaults to true)
+                                    | ``references`` (\*): Git reference. A local reference repo to be used as
+                                    |       alternate (see man git-clone).
+                                    |       A list of strings or a dictionaries with
+                                    |        ``url``: (optional, Regex-String, default: ``.*``). The matching part
+                                    |           of the remote URL is replaced by
+                                    |        ``repo``: (String) local storage path.
+                                    |        ``optional``: (Boolean, default True). Marks the reference as
+                                    |           optional if true. Otherwise a error is raised if the
+                                    |           local reference repo didn't exitst.
+                                    |   Note: ``references`` are not used for submodules.
+                                    | ``disassociate``: (Boolean, default false). Diasassociate the reference.
 import Import directory from        | ``url``: Directory path relative to project root.
        project                      | ``prune`` (\*): Delete destination directory before importing files.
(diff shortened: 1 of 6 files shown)


rhubert:
I just noticed that at least support for multiple references is missing...

jkloetzke:
We should certainly merge the scm-defaults PR first to get this untangled.

I think `references` could also be a list. I'm not yet sure about the `pattern`/`path` thing. Would the url always be the source or could the reference depend on other attributes too? And it should probably be a regex to make this more generic...

rhubert:
> We should certainly merge the scm-defaults PR first to get this untangled.

I updated this PR ;) 

> I think `references` could also be a list.

I agree. 

> I'm not yet sure about the `pattern`/`path` thing. Would the url always be the source or could the reference depend on other attributes too?

For my usecase there is a local path where a repo is mirrored using `git clone --mirror <url>/repo.git`.

With this there are two options to use this mirror: 
- `git clone --reference(-if-able)=<local-path>/repo.git <url>/repo.git [...]` 
-  manual providing the `.git/objects/info/alternates` file pointing to `<local-path>/repo.git/objects`.
I think it should also be tested if the local mirror is available. Adding garbage to the alternates files makes git somehow unhappy.

Anyway - ATM I do not see any other input than the URL which needs to be replaced partially.  

> And it should probably be a regex to make this more generic...

For the part of the ULR which needs to be replace? Should be possible.

jkloetzke:
> > I'm not yet sure about the `pattern`/`path` thing. Would the url always be the source or could the reference depend on other attributes too?
> 
> For my usecase there is a local path where a repo is mirrored using `git clone --mirror <url>/repo.git`.
> 
> With this there are two options to use this mirror:
> 
> * `git clone --reference(-if-able)=<local-path>/repo.git <url>/repo.git [...]`
> * manual providing the `.git/objects/info/alternates` file pointing to `<local-path>/repo.git/objects`.
>       I think it should also be tested if the local mirror is available. Adding garbage to the alternates files makes git somehow unhappy.

Yes. As we don't use `git clone` setting up the `alternates` file is certainly ok. I have the feeling we should also both `--reference` and `--reference-if-able` options somehow. It really depends on the use case if the checkout should abort if the alternate repository is not available. The default should probably be to fail if the reference repo is not available.

> 
> Anyway - ATM I do not see any other input than the URL which needs to be replaced partially.

Me neither. I was just pondering about the possibility that the current solution might be too specific. OTOH the local mirror will somehow always relate to the URL. Let's start with the current approach. We can add still something if the use case arises.

> > And it should probably be a regex to make this more generic...
> 
> For the part of the ULR which needs to be replace? Should be possible.

Exactly. It would allow to rewrite the URL in any way, not just the prefix.

Oh, and there should be some support for the Jenkins git plugin too...

jkloetzke:
Thanks :+1:

rhubert:
Thanks :+1: any plans for a new release?

jkloetzke:
Looks like the plans got updated. :wink: I can prepare a -rc1 in the next days....

PR #445, opened 2021-11-01
jfrog artifactory backend server for bob artifacts

mahaase:
* all credits to @rhubert #240 
* re-tested twice :)
* synced with current master
* removed the replace of "/" by "_" in upload/download url
* i guess we should bring this into the master (instead loosing it)

### DIFF
--- a/doc/manual/configuration.rst
+++ b/doc/manual/configuration.rst
@@ -1515,6 +1515,12 @@ the following table for supported backends and their configuration.
 Backend     Description
 =========== ===================================================================
 none        Do not use a binary repository (default).
+artifactory JFrog Artifactory backend. Use the ``url`` keyword to provide the
+            repository url. Use optional keys ``username`` to specify the user
+            and ``key`` for the API-key or password. Without ``username`` and
+            ``key`` either no authentication or the global artifactory
+            configuration file is used. See dohq-artifactory documentation for
+            details.
 azure       Microsoft Azure Blob storage backend. The account must be specified
             in the ``account`` key. Either a ``key`` or a ``sasToken`` may
             be set to authenticate, otherwise an anonymous access is used.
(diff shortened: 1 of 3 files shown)


rhubert:
There are some open issues in #240. Do you plan to solve them? Otherwise it might be better to reopen #240 and resume working on this? 
How do you delete artifacts from artifactory? Are you using `bob archive` on the server?

mahaase:
I can also open a MR to your branch, so u don't have to sync with master again.
Then we can reopen the #240.

> How do you delete artifacts from artifactory? Are you using bob archive on the server?

Currently we do not delete artifacts at all. But yeah, I have no idea, how it could be possible to clean up the artifactory. 
I don't know how the artifactory is managed in background and if `bob archive` would work. 
Good point, we should check this here.

rhubert:
I can just use your branch - this won't make a difference ;) I'm just asking because ATM I'm using http upload to artifatory (no need for additional properties) and this works without any problems. I don't have a good use case for additional artifactory integration - which properties are needed,... ? I'd need some input from your side.

```quote
 I don't know how the artifactory is managed in background and if bob archive would work.
```
I've a `bob archive` with support for artifactory cleanup, but this is currently only in a 'works-somehow' state...

mahaase:
the option to get additional properties would be nice. e.g. for images that will be released to customer. we could adapt the recipe like this, that a user of artifactory get all necessary information: versions, dependencies, variants, ....

> I've a bob archive with support for artifactory cleanup, but this is currently only in a 'works-somehow' state...

sounds cool, for a full bob-artifactory support, this is important.

mahaase:
we switched to gitea registry. opened a new pull-request for this new backend. so we don't need jfrog anymore. i will close it.

PR #453, opened 2022-01-22
input: add checkoutDep dependency flag

jkloetzke:
Normally dependencies are only available to the buildScript. But
sometimes it is necessary to access dependencies in the checkoutScript
already. Currently this requires to copy the sources in the
buildScript where access to all dependencies is possible.

The newly added checkoutDeps keyword takes a list of dependencies that
shall be available to the checkoutScript. The names must be given
exactly and they must be a valid dependency. Provided dependencies of
such named packages are *not* implicitly made available. The number of
checkout dependencies should be kept as small as possible to prevent
accidental creation of checkout variants which are usually confusing to
users.

### DIFF
--- a/doc/manual/configuration.rst
+++ b/doc/manual/configuration.rst
@@ -528,6 +528,16 @@ a script for the same step are joined in front of this script in the order the
 inheritance is specified. The inheritance graph is traversed depth first and
 every class is included exactly once.
 
+Dependencies of the recipe are by default only available to the
+``buildScript``. The path to the previous step (checkout workspace for
+``buildScript``, build workspace for ``packageScript``) is always passed in
+``$1``. Other dependencies are available in the order in which they were
+declared at the :ref:`configuration-recipes-depends` section of the recipe. If
+a dependencies ``checkoutDep`` flag is set to ``True`` it will also be
+available to the ``checkoutScript``. This should be used carefully as it makes
+the checkout of the recipe sources dependent on the result of another
+dependency.
+
 During execution of the script only the environment variables SHELL, USER,
 TERM, HOME and anything that was declared via {checkout,build,package}Vars
 are set. The PATH is reset to "/usr/local/bin:/bin:/usr/bin" or whatever was declared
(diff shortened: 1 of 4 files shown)


Ferruck:
Works with my test case (`libpciexpress` with its checkout dependency to `xorg-macros`). One question, though: Why not add the availability at the checkout step as a *property* to the initial declaration of dependencies. That way, all information regarding the recipe's (or rather package's) dependencies would stay in one place rather than being scattered across the recipe. Something like

```yaml
depends:
    - libs::zlib-dev
    - name: xorg-macros
      availableAtCheckout: True
    - use: []
      depends:
        - libs::zlib-tgt
```
(with `availableAtCheckout` obviously defaulting to `False`) would be more apparent when reading a recipe, IMHO...

jkloetzke:
> One question, though: Why not add the availability at the checkout step as a property to the initial declaration of dependencies. That way, all information regarding the recipe's (or rather package's) dependencies would stay in one place rather than being scattered across the recipe.

I also thought about this but arrived at the current solution. By doing it in the "depends" list you could not add indirect (provided) deps to the checkout dependencies. I also found `checkoutDeps` more natural because the `checkout*` stuff is usually grouped in the recipe and IMHO having a dependency in the checkout step is much like a tool or variable. It's another ressource available to the checkout step.

Note that the dependency is _always_ available to the build step. It's just that it is _also_ available to the checkout step. Though I'm not sure if this is a good idea or not...

Ferruck:
I'm fine with either solution, but

> By doing it in the "depends" list you could not add indirect (provided) deps to the checkout dependencies.

stuff like this makes recipes harder to reason about, in my personal opinion. If a package *has* a specific dependency it should list it, not require some (maybe even recursive) dependency incidentally providing it. So, in my world view that's not really a feature.. :wink: But again, this is just my personal opinion and your changes fix my problem at hand.

jkloetzke:
> I'm fine with either solution, but
> 
> > By doing it in the "depends" list you could not add indirect (provided) deps to the checkout dependencies.
> 
> stuff like this makes recipes harder to reason about, in my personal opinion. If a package _has_ a specific dependency it should list it, not require some (maybe even recursive) dependency incidentally providing it. So, in my world view that's not really a feature.. :wink: But again, this is just my personal opinion and your changes fix my problem at hand.

No, you have a point. For things like dependencies that need to be available in the checkout step we could require to name them explicitly. I think I'll rework the PR...

jkloetzke:
Ok. I think this should be better...

Ferruck:
I just tested your changes and they work as expected for my X11 recipes. Thanks!

PR #456, opened 2022-02-07
Git misc

rhubert:


pym/bob/scm/git.py
```diff
@@ -253,6 +253,15 @@ async def __checkoutTag(self, invoker, fetchCmd, switch):
...
                             "did not download the requested commit and the explicit fetch failed!",
                             returncode=ok)
+
+            # do not perform inline switch if we're in a unclean detaced
+            # head state
+            status = await invoker.checkOutputCommand(["git",  "status", "--porcelain"],
+                    cwd=self.__dir)
+            if status:
+                await invoker.checkCommand(["git", "symbolic-ref",
+                    "-q", "HEAD"], cwd=self.__dir)
```
jkloetzke, line 266: What about a clean detached head state? I thought the original issue was that somebody made commits without being on a branch..?

I would not make any distinction about clean/unclean because git handles this if modified files would be overwirtten. Unrelated files can still be left as-is. There is already a dedicated switch to ensure clean checkouts.

I think this is unrelated to checkouts of tags/commits. The user could have checked out a tag/commit himself. How about the following: If we ever switch away from a detached-head state then move to attic. The only exception is that the old checkout (`oldSpec`) was on the commit or tag that still matches the current HEAD. This would retain switching from tag/commits to other states unless the user checked out some other commit or made commits on detached heads. If the user created a branch we would also do an inline switch which is IMHO ok. What do you think?

doc/manual/policies.rst
```diff
@@ -542,3 +542,24 @@ New behaviour
    The ``prune`` property defaults to ``True``. The user must edit the files at
    the import source location because the destination in the workspace is
    overwritten and obsolete files are deleted.
+
+.. _policies-gitBranchAndCommit:
+
+gitBranchAndCommit
+~~~~~~~~~~~~~~~~~~
+
+Introduced in: 0.20
```
jkloetzke, line 551: Should be 0.21. 0.20 was already released.

pym/bob/scm/git.py
```diff
@@ -106,6 +107,7 @@ def __init__(self, spec, overrides=[], secureSSL=None, stripUser=None):
         self.__references = spec.get('references')
         self.__dissociate = spec.get('dissociate', False)
         self.__resolvedReferences = []
+        self.__useBranchAndCommit = spec.get('useBranchAndCommit', useBranchAndCommit)
```
jkloetzke, line 110: Why would you ever want to override this on a per-SCM basis? I would only add this if there is a valid use case.

pym/bob/scm/git.py
```diff
@@ -262,8 +265,22 @@ async def __checkoutTag(self, invoker, fetchCmd, switch):
...
+                    "--contains", self.__commit], cwd=self.__dir)
+                if not re.search( r".*/origin/"+re.escape(self.__branch), branches, re.MULTILINE):
+                    raise BuildError ("Commit: " + self.__commit + " is not on branch "
+                        + self.__branch)
+                # if we have both checkout the branch and move to the
+                # commit
+                await self.__checkoutBranch(invoker, fetchCmd, switch)
+
+                await invoker.checkCommand(["git", "reset", "--keep",
+                    "--no-recurse-submodules", self.__commit], cwd=self.__dir)
```
jkloetzke, line 280: I think this is dangerous if some sub-module was added between the commit and the current HEAD of the branch...

pym/bob/scm/git.py
rhubert: > What about a clean detached head state? I thought the original issue was that somebody made commits without being on a branch..?
> 
> I would not make any distinction about clean/unclean because git handles this if modified files would be overwirtten. Unrelated files can still be left as-is. There is already a dedicated switch to ensure clean checkouts.
> 
A clean detached state is the state we have with a `commit` checked out. I don't want to move to `attic` every time the commit ID is updated without a reason.

> I think this is unrelated to checkouts of tags/commits. The user could have checked out a tag/commit himself. How about the following: If we ever switch away from a detached-head state then move to attic. The only exception is that the old checkout (`oldSpec`) was on the commit or tag that still matches the current HEAD. This would retain switching from tag/commits to other states unless the user checked out some other commit or made commits on detached heads. If the user created a branch we would also do an inline switch which is IMHO ok. What do you think?

:+1: sounds good.

pym/bob/scm/git.py
rhubert: AFAICR I put it into the spec to have it set to a non `None` value in `__checkoutTag`.

pym/bob/scm/git.py
```diff
@@ -266,8 +269,21 @@ async def __checkoutTag(self, invoker, fetchCmd, switch, oldSpec):
...
+                branches = await invoker.checkOutputCommand(["git", "branch", "-a",
+                    "--contains", self.__commit], cwd=self.__dir)
+                if not re.search( r".*/origin/"+re.escape(self.__branch), branches, re.MULTILINE):
+                    raise BuildError ("Commit: " + self.__commit + " is not on branch "
+                        + self.__branch)
+                # if we have both checkout the branch and move to the commit
+                await self.__checkoutBranch(invoker, fetchCmd, switch)
+
+                await invoker.checkCommand(["git", "reset", "--keep",
+                    "--recurse-submodules", self.__commit], cwd=self.__dir)
```
jkloetzke, line 283: This will always move the branch to the current head and then back to the configured commit. If sub-modules were added in between, this will leave behind a wrecked workspace because even git cannot cope with vanishing submodules. Instead I would suggest that `__checkoutBranch()` somehow needs to check out the configured commit directly.

pym/bob/scm/git.py
rhubert: No, I don't want this to be set able per SCM. But I only have access to the policy when the scm is constructed from the input.py Scm-class. When the inkoker calls `getScm` I don't have access to the policy - or am I overlooking something?

jkloetzke: > When the inkoker calls getScm I don't have access to the policy - or am I overlooking something?

No, but then there is no need to also look in the spec. So the following should be sufficient:

```suggestion
        self.__useBranchAndCommit = useBranchAndCommit
```

rhubert: No it's not. Adding a debug print after `self.__useBranchAndCommit = useBranchAndCommit` shows:

```
useBranchAndCommit: True
useBranchAndCommit: None
```

within a `bob dev..` run. In `__checkoutTag` it's always `None`.

jkloetzke: You're absolutely right. Sorry for the noise. :see_no_evil: I totally forgot that the Scm instances are serialized and de-serialized before being executed.

pym/bob/scm/git.py
jkloetzke: Thinking more about it I think this new behaviour should not be folded into the existing code paths but be an entire new method, e.g. `__checkoutTagOnBranch`. Because effectively its a mixture of `__checkoutTag` and `__checkoutBranch`. Then we don't have to change the existing behaviour...

jkloetzke:
I took myself the freedom to push my "review" as a commit. I hope you don't mind. It still needs some cleanup and slicing into the different commits. There were quite a number of issues:

* The logic to move a "detached HEAD" state repository to the attic needs to be independent of what is being checked out. E.g. the user might move to a branch but the repository is currently in a detached HEAD state with some additional commits. Even then we want to move to the attic.
* The initial implementation did still checkout the head of the branch and then rewind to the given commit. This would have brought in submodules. The test just did not trigger this edge case because the first checkout was done without submodules.
* `__checkoutTagOnBranch` would call `git reset` every time, potentially throwing away commits from the user
* Calling `git rev-list <branch>` to check if a commit is on the branch has a huge overhead for big repositories like the Linux kernel

There were some other minor things that I forgot. Could you test your use case with the current state?

rhubert:
:+1: Thanks. I tested your rewrite today locally as well as on our CI without any issues.

jkloetzke:
Thanks. :+1:

PR #461, opened 2022-04-14
builder: check earlier for checkout-only flag

ThomasFeher:
It is not necessary to update the buildDigest in case `--checkout-only`
is set, obviously. This can avoid rebuilds in case an update of the
sources was done using `--checkout-only` without `--sandbox` and then
rebuilding with `--sandbox` for example.


jkloetzke:
Nice catch. :+1:  Thanks!

PR #463, opened 2022-04-19
mirrors for url scm

daxcore:
mirrors will be tried from top to bottom

examples:
```
checkoutSCM:
     scm: url
     url:
       - http://...
       - https://...
       - ftp://...
       - $(mirrors,$(list,http://...,https://...),/somewhere/archive.tar.gz)
```

or:
```
checkoutSCM:
     scm: url
     url: $(mirrors,$(list,http://...,https://...),/somewhere/archive.tar.gz)
```

it is possible to use a environment variable for the list:
```
checkoutSCM:
     scm: url
     url: $(mirrors,${MIRRORS},/somewhere/archive.tar.gz)
```

the variable can be defined e.g. in your default.yaml:
```
environment:
    MIRRORS: "$(list,http://...,https://...)"
```

### DIFF
--- a/pym/bob/scm/url.py
+++ b/pym/bob/scm/url.py
@@ -125,7 +125,7 @@ class UrlScm(Scm):
 
     __SCHEMA = {
         'scm' : 'url',
-        'url' : str,
+        'url' : schema.Or(str, [str]),
         schema.Optional('dir') : str,
         schema.Optional('if') : schema.Or(str, IfExpression),
         schema.Optional('digestSHA1') : str,
@@ -190,7 +190,7 @@ class UrlScm(Scm):
         self.__dir = spec.get("dir", ".")
         self.__fn = spec.get("fileName")
         if not self.__fn:
-            url = self.__url
+            url = self.__url[0] if isinstance(self.__url, list) else self.__url
             if isWindows():
                 # On Windows we're allowed to provide native paths with
                 # backslashes.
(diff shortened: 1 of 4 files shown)


daxcore:
feedback welcome. it's not the generic way. it's more the "minimal changes" for a new feature way...
i checked the generic way, this one will be much harder, because of the url parameter is used in so many locations!

pym/bob/scm/url.py
```diff
@@ -291,31 +291,47 @@ async def invoke(self, invoker):
 
         # Download only if necessary
         if not self.isDeterministic() or not os.path.isfile(destination):
-            try:
```
daxcore, line 294: this diff looks much more than it is, because of the indenting.

pym/bob/scm/url.py
```diff
@@ -291,31 +291,47 @@ async def invoke(self, invoker):
...
+                    # Local files: copy only if newer (u), target never is a directory (T)
+                    if isYounger(url.path, destination):
+                        if os.path.isdir(destination):
+                            invoker.fail("Destination", destination, "is an existing directory!")
+                        invoker.trace("<cp>", url.path, workspaceFile)
+                        shutil.copy(url.path, destination)
+                elif url.scheme in ["http", "https", "ftp"]:
+                    invoker.trace("<wget>", self.__url, ">", workspaceFile)
+                    try:
+                        err = await invoker.runInExecutor(UrlScm._download, self, destination)
```
daxcore, line 316: i tried to give that method the url, instead of self, but this hasn't worked for me for unknown reasons.
i had to use the self.__url. (ref. line 297 (new))

jkloetzke:
I'm afraid this will not work in general. String functions were intended to work only on strings. Now this actually only works because of the current implementation does not check the return type of string functions. But I'm sure you could create crashes when using the `mirrors` function in other contexts.

Generally speaking I like the "url may be a list" approach. I'm just not sure it will work out in the end.

Also mirrors can only work reliably if the SCM is deterministic. If no digest was given, the URL is used to detect changes to the SCM (and force a clean rebuild). What will happen if another URL is added. How do we handle the case when the URLs resolve to something different content wise and no digest was given? I have the vague feeling that the impact is much higher as it looks at the first glance. Unfortunately I don't have a good idea at the moment how things should really looks like.

PR #474, opened 2022-05-25
input: fixed typos for MSYS2 environment variables in whitelist

### LINKED ISSUE #473: wrong windows/MSYS whitelisted environment variable names
https://github.com/BobBuildTool/bob/blob/056cf294baee7e10d1e3d23140c1ef63aa0855ac/pym/bob/input.py#L3095-L3099

win | msys
CommonProgramFiles | COMMONPROGRAMFILES
CommonProgramFiles(x86) | CommonProgramFiles(x86)
ProgramFiles | PROGRAMFILES
ProgramFiles(x86) | ProgramFiles(x86)
Path | PATH
ProgramData | ProgramData
SystemDrive | SYSTEMDRIVE
SystemRoot | SYSTEMROOT
windir | WINDIR

The win-variant is not case-sensitive in native windows, but the MSYS-variant is case sensitive!!
Becuase the names are not consistent, we should add always both variants of each.
"windir" is all lower-case.. funny mhh? :D

would we add this one too?
PROCESSOR_ARCHITECTURE necessary for some libraries like boost >= 1.76.0

mahaase:
There are mismatching names in whitelist.
MSYS2 is handling the names case-sensitive, but not native windows.

.FIXES #473

### DIFF
--- a/pym/bob/input.py
+++ b/pym/bob/input.py
@@ -3512,6 +3512,8 @@ class RecipeSet:
         self.__projectRoot = recipesRoot or os.getcwd()
 
     def __parse(self, envOverrides, platform, recipesRoot=""):
+        if platform not in ('cygwin', 'darwin', 'linux', 'msys', 'win32'):
+            raise ParseError("Invalid platform: " + platform)
         self.__platform = platform
         self.__whiteList = set()
         if platform in ('cygwin', 'msys', 'win32'):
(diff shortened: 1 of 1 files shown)


jkloetzke:
I think this breaks native Windows which unfortunately is not visible in the CI because we do not have Windows tests (yet).

On native Windows, the Python interpreter will convert all variables to upper case. So the white list needs to stay in upper case. Maybe the non upper case variants should be used solely on MSYS? Are we sure that the case always stays the same? We once had a similar issue in #432 where it turned out that there are even variables that only differ in case, e.g. "tmp" and "TMP". Shouldn't we white list both?

jkloetzke:
I took myself the freedom to update the PR. Now the case sensitive variables are only used on MSYS. Native Windows is not changed.

I would also rather not add any more variables to the default whitelist. If other packages like boost need additionaly variables then it could easily be added in the layer/project default.yaml.

PR #484, opened 2022-07-14
Unify backends

jkloetzke:
Get rid of the special Jenkins build logic. Make it fit for native Windows and adapt the tests accordingly...


rhubert:
:+1: I did some testing with this branch locally as well as on our CI and couldn't find any issues.

PR #485, opened 2022-07-29
RFC: archive: add plugin interface

### LINKED ISSUE #340: bob-archive should be able to act remotely
> > That's because bob archive only works locally where the artifacts are stored. It's poorly documented in the manpage.
> 
> Well, yes, indeed. Is this just because nobody cared to implement a remote `bob archive` or are there technical reasons?

There was no need yet. Until now it was sufficient to have a cron-job on the artifact server that works locally. Technically it is possible to implement that but it will require the HTTP endpoint to be a WebDAV resource (needs `PROPFIND` to list the directories). The `shell` backend might need a `ls` command too.

_Originally posted by @jkloetzke in https://github.com/BobBuildTool/bob/pull/336#issuecomment-635184140_

rhubert:
Extend the archive command to clean up a non local archive using a
custom plugin. Therefore the plugin manifest must contain
'archiveAccessors'. Each entry should point to a class inheriting from
'BaseArchiveAccess' providing the appropriate methods to access the
custom archive.

Fixes #340.

### DIFF
--- a/contrib/plugins/artifactoryArchiveAccess.py
+++ b/contrib/plugins/artifactoryArchiveAccess.py
@@ -1,0 +1,73 @@
+from bob.archive_access import BaseArchiveAccess
+
+from artifactory import ArtifactoryPath
+import os
+import tempfile
+import datetime
+import calendar
+import struct
+
+import urllib3
+urllib3.disable_warnings(urllib3.exceptions.InsecureRequestWarning)
+
+class Artifactory(BaseArchiveAccess):
+    def __init__(self):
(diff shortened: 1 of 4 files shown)


jkloetzke:
I'm afraid a plug-in is not the right approach. Why not extend the archive classes to handle the garbage collection too and let the `bob archive` command work on the define archive entries from default.yaml?

Regarding Artifactory, I hope that it does properly support WebDAV. That's what the http backend should implement...

rhubert:
OK - I'll have a look @ WebDAV and come back...

PR #493, opened 2022-10-31
Run selected checkoutScripts in build-only mode

jkloetzke:
By default no checkout scripts are run when building with `--build-only`.
Some use cases practically require the `checkoutScript` to be always run,
through. A typical example are code generators that generate sources
from some high level description. These generators must be run every
time when the user has changed the input. A recipe or class can
explicitly opt in to run their `checkoutScript` also in build-only mode to
cover such a use case. This is done by either setting `checkoutUpdateIf`
to `True` or by a boolean expression that is evaluated to `True`. Otherwise
the `checkoutScript` is ignored even if some other class enables its
script. The `checkoutUpdateIf` property thus only applies to the
corresponding `checkoutScript` in the same recipe/class.

### DIFF
--- a/doc/manpages/bob-build-dev.rst
+++ b/doc/manpages/bob-build-dev.rst
@@ -181,7 +181,9 @@ Options
    defined by Bob itself (e.g. ``meta.bob``) cannot be redifined!
 
 ``-b, --build-only``
-    Don't checkout, just build and package
+    Don't checkout, just build and package. Checkout scripts whose
+    :ref:`configuration-recipes-checkoutUpdateIf` property was evaluated as
+    true will still be run.
 
     If the sources of a package that needs to be built are missing then Bob
     will still check them out. This option just prevents updates of existing
--- a/doc/manual/configuration.rst
+++ b/doc/manual/configuration.rst
@@ -966,7 +966,7 @@ git
    The ``rev`` property of the ``git`` SCM unifies the specification of the
    desired branch/tag/commit into one single property. If present it will be
    evaluated first. Any other ``branch``, ``tag`` or ``commit`` property is
-   evalued after it and may override a precious setting made by ``rev``. The
+   evaluated after it and may override a previous setting made by ``rev``. The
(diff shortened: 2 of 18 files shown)


jkloetzke:
@mahaase @rhubert It would be good to have your input on this. It should be rather flexible and cover the known use cases without introducing a new kind of checkoutScript...

rhubert:
I did some simple tests and it looks good to me. Just to get it right - atm we have something like:

```yaml
checkoutSCM:
 scm: url
 url: foo
 extract: false

checkoutScript: |
  extract foo

buildScript: |
  run_gen $1/foo
  build..
```

Now with the `checkoutUpdateIf` this could be changed to:

```yaml
checkoutSCM:
 scm: url
 url: foo
 extract: false

checkoutUpdateIf: True
checkoutScript: |
  if  [ foo_has_changed ]; then
     extract foo
  fi
  run_gen .

buildScript: |
  build ..
```

Is there any possibility to avoid the need of a custom `foo_has_changed`? Maybe a `BOB_BUILD_ONLY_MODE`  env var?

jkloetzke:
> Is there any possibility to avoid the need of a custom `foo_has_changed`? Maybe a `BOB_BUILD_ONLY_MODE` env var?

Actually I have already thought about something along these lines. The thing is, it is usually not needed to know. The checkout script can be re-run at any time. The user could add `-f` when building, a tool that used by the checkout script could change, you could manually call `checkout.sh`... In all these cases the `checkoutScript` should only do the `extract foo` if `foo` has actually changed. That's how the automatic extraction is implemented in Bob. So looking at a "build mode" environment variable would not be sufficient. Or do I overlook something?

rhubert:
Hmm...I think you're right. If `extract foo` needs to be rerun did not only depend on `-b`. What if `extract` is a `checkoutTool`,... 🤔  
Looks like special cases need special treatment....

rhubert (approved):

PR #494, opened 2022-11-04
git: fix status output for commits only on tags

rhubert:
If a commit is only on a tag but not on a branch the tag was shown as unpushed.

### DIFF
--- a/pym/bob/scm/git.py
+++ b/pym/bob/scm/git.py
@@ -680,7 +680,6 @@ class GitScm(Scm):
         status = ScmStatus()
         try:
             onCorrectBranch = False
-            onTag = False
             output = self.callGit(workspacePath, 'ls-remote' ,'--get-url')
             if output != self.__url:
                 status.add(ScmTaint.switched,
@@ -702,7 +701,6 @@ class GitScm(Scm):
                 # log" command at the end will trip.
                 try:
                     self.callGit(workspacePath, 'rev-parse', 'tags/'+self.__tag)
-                    onTag = True
                 except BuildError:
                     pass
             elif self.__branch:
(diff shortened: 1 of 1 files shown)


jkloetzke:
I'm not sure that this is a good idea. Git does not track where tags are coming from. So with this change, if I do some commits and create a tag, the commits would not be counted as unpushed. That was the reason why the checked out tag is selectively added to the exclude list because it will most likely have come from the remote. (see two lines below)

BTW, having commits that are not on branches is not a good idea. Usually you don't get them with a plain `git fetch`. This can cause a number of surprises...

rhubert:
We have a many of repos with many tags pointing to commits that are not on branches. This makes `bob status -v` unusable.

e.g. systemd has a `v250.5` tag which doesn't belong to a branch: https://github.com/systemd/systemd/tree/v250.5

This happens for many opensource repos (also because we don't delete tags in our mirror if there were deleted upstream - which is the case in the busybox-repo,...)

rhubert:
Another option might be to show `u` only if the verbosity is increased to `-vv`?

jkloetzke:
I see. Looks like this ship has sailed and the world is actually doing "tags not on a branch". :roll_eyes: 

So I think the change is ok. Could you please remove the special handling of the current tag from two lines below? It's now a no-op because `--tags` will include it already. An please adjust the comment above and add a rationale why we want to exclude all tags from the unpushed commits... :wink:

rhubert:
> So with this change, if I do some commits and create a tag, the commits would not be counted as unpushed.

I added some code to compare remote tags and local tags and if there is a local-only-tag this is shown as unpushed tag now.

jkloetzke:
> > So with this change, if I do some commits and create a tag, the commits would not be counted as unpushed.
> 
> I added some code to compare remote tags and local tags and if there is a local-only-tag this is shown as unpushed tag now.

I like the idea but I'm not sure if this working always well in practice. It very much depends on the git server. So far the status command worked without network connectivity. For example the GitHub server for the bob repository takes almost 2s to respond. I'm a bit worried about the slowdown...

rhubert:
Hmm..even for the linux kernel repo this takes only 400ms using my setup.

But I see your point. I think we could either add a new scm-value to the gitSCM (`statusAll` ?), which could be set using `scmDefaults` or skip this test completely?

jkloetzke:
> Hmm..even for the linux kernel repo this takes only 400ms using my setup.

It will highly depend on the responsiveness of the server. My fear is that the command becomes unusable if you have multiple servers and they are under high load. Previously the status command of git SCMs was working completely offline.

> But I see your point. I think we could either add a new scm-value to the gitSCM (`statusAll` ?), which could be set using `scmDefaults` or skip this test completely?

I would skip the test completely. The possibility that a user created a tag and forgot to push it is very unlikely. The `u` flag is also just information and never considered a real data loss. I don't think it's worth covering this corner case at the expense of the reduced speed.

jkloetzke:
Oh, I just saw that we have another test for unpushed commits in `__statusSubmodule()` too. Please adapt it there accordingly...

jkloetzke:
Thanks!

PR #500, opened 2022-12-13
doc: Fix typos

georgeto:


doc/tutorial/compile.rst
```diff
@@ -1,7 +1,7 @@
 Build a demo project
 ********************
 
-Prequisites
+Prerequisites
 ===========
```
jkloetzke, line 5: I'm afraid rst requires that the underlining must be at least as long as the text...

```suggestion
=============
```

jkloetzke (changes requested):
Thanks for fixing the typos.

doc/tutorial/compile.rst
georgeto: You never stop learning ;)

jkloetzke:
Thanks!

PR #503, opened 2023-01-20
add missing schema option attic

MarcKe:
attic was missing in this schema, so it was not possible to use it in user configurations


jkloetzke:
Thanks!

PR #504, opened 2023-01-23
tty: make the MassiveParallelTUI threshold settable

rhubert:
Make the threshold customizable as some people were confused about the new MassiveParallelTUI. They wished to get the old TUI back without being limited to 16 jobs.


jkloetzke:
Thanks. Makes sense. :smile:

PR #507, opened 2023-01-30
generators: fix ParseError reporting

rhubert:
If a invalid regular expression is provided a parse error should be shown. Unfortunately the error reporting was broken and raised another exception:

    raise ParseError("Invalid regular expression '{}': {}".format(e.pattern), e)
 NameError: name 'ParseError' is not defined

    raise ParseError("Invalid regular expression '{}': {}".format(e.pattern), e)
 IndexError: Replacement index 1 out of range for positional args tuple


jkloetzke:
Thanks!

PR #510, opened 2023-02-17
input: support for variable depends

rhubert:


pym/bob/input.py
```diff
@@ -2355,11 +2355,21 @@ def prepare(self, inputEnv, sandboxEnabled, inputStates, inputSandbox=None,
         depDiffSandbox = diffSandbox
         depDiffTools = diffTools.copy()
         thisDeps = {}
+        depsEnv = env.derive()
+        depVars = set()
 
-        for dep in self.__deps:
+        __provideDeps = [ env.substitute(p, "providedDep::"+p)
```
jkloetzke, line 2361: This is also unexpected. For example `provideVars` uses the final environment of the package after all dependencies were traversed. I would like to have this for `provideDeps` too unless there is a good reason against it.

pym/bob/input.py
```diff
@@ -2545,7 +2555,7 @@ def prepare(self, inputEnv, sandboxEnabled, inputStates, inputSandbox=None,
         # optional checkout step
         if self.__checkout != (None, None, None) or self.__checkoutSCMs or self.__checkoutAsserts:
             checkoutDigestEnv = env.prune(self.__checkoutVars)
-            checkoutEnv = ( env.prune(self.__checkoutVars | self.__checkoutVarsWeak)
+            checkoutEnv = ( env.prune(self.__checkoutVars | self.__checkoutVarsWeak | depVars)
```
jkloetzke, line 2558: This can't be right. Why are the substituted variables implicitly added to the environment variables? If the user wants to access these variables then he can add easily add them to `checkoutVars`...

jkloetzke (commented):
I think there is a fundamental problem with the handling of `provideDeps`. With the current approach the variable expansion will be done *after* the pattern expansion. This is exactly the opposite order in which shells are doing it. Think about patterns like `*-${FOO}-dev`. This won't ever match anything. I think the order has to be reversed to be useful...

pym/bob/input.py
rhubert: Hmm...the reason is most likely my misunderstanding of the code. I guess it must be added to `[checkout|build|package]DigestEnv` only to get different `[checkout|build|package]Ids`?

But for the `checkoutId` this is only necessary if `checkoutDep` is set, isn't it?

rhubert:
At least `providedDeps` are still broken... I'll look into this later...

pym/bob/input.py
jkloetzke: > Hmm...the reason is most likely my misunderstanding of the code. I guess it must be added to `[checkout|build|package]DigestEnv` only to get different `[checkout|build|package]Ids`?

No, this is not needed. You can safely remove this and it should work as expected.

rhubert: I'm lost. If I remove the `| depVars` changes the unit `test_input_recipeset` is failing Because the package is always the same regardless of `$A` setting... 🤔

rhubert:
I think the `provideDeps` issue is now solved. I squashed all the commits - please comment the latest version ;)

pym/bob/input.py
jkloetzke: This points to some yet not understood problem. I don't have an idea at the moment but I'm looking into it...

jkloetzke: Actually this is nothing new. The package graph is aggressively consolidated based on the following premise: if a particular recipe is fed with different environment+tools but yields the same result (as in Variant-Id) then the package graph is collapsed to a single variant. Usually you don't notice this but in your testcase the `b-foo` and `c-foo` recipes are identical so this optimization kicks in. You can get the same effect if you change `a.yaml` to:

```yaml
depends:
    - name: b-foo
      if: "$(eq,$A,b)"
    - name: c-foo
      if: "$(eq,$A,c)"
buildScript: "true"
packageScript: "true"
provideDeps: [ "*" ]
provideVars:
    D: "e"
```

So it's better to fix the test case by making the scripts in  `b-foo` and `c-foo` different.

rhubert: Hmm...so whenever we use a variable only in the depends section we'll get potentially wrong results? I wonder why I never fall about this.
My usecase is to provide different configs to busybox:

```yaml
depends
 - "${BUSYBOX_CONFIG:-core::busybox-config}"

buildScript: |
  ...
  if  [[ -e $2/busybox.config ]]; then  
     cp $2/busybox.config .
  fi
```

-> There is no obvious need to add `BUSYBOX_CONFIG` as `buildVar`. I think this should be at least documented somewhere...

jkloetzke: > Hmm...so whenever we use a variable only in the depends section we'll get potentially wrong results?

No, Bob will do the right thing **content wise**. If your `$BUSYBOX_CONFIG` points to different recipes that provide *different* configurations (technically: the variant-id of the packages must differ) then everything will be as expected. The only unexpected thing that happens is when `$BUSYBOX_CONFIG` points to different recipes but these recipes are identical (read: same variant-id). Then Bob will still build the right thing but in `bob ls` you will see only one of the `$BUSYBOX_CONFIG` dependencies and the other one is silently dropped.

> -> There is no obvious need to add `BUSYBOX_CONFIG` as `buildVar`. I think this should be at least documented somewhere...

There is no need to add `BUSYBOX_CONFIG` as `buildVar`. What may be documented is that identical dependencies with different names might "vanish". But I consider this an outlier because I have never encountered different recipes that were identical. That's why I have never added code to handle this case that I though will never happen in the first place.

jkloetzke:
I think the code works as intended. However, I took myself the freedom to optimize it a bit so that the performance overhead is not noticeable. Previously the recipe parsing time was ~8% longer, even if the feature was not used. Now it's somewhere at 0.3%.

Once the documentation is updated it's ready for merging I guess...

rhubert:
Thanks!!! 

I did some testing this morning and it's still working for my use-cases ;) 

Added a few words to the documentation + squashed my commits.

jkloetzke:
Thanks!

PR #514, opened 2023-04-12
project-generators: enable use of query language

rhubert:
While many of the generators do not need it there are use cases where a generator needs to handle a pathspec rather than a package-path.

### DIFF
--- a/pym/bob/cmds/build/project.py
+++ b/pym/bob/cmds/build/project.py
@@ -107,7 +107,11 @@ def doProject(argv, bobRoot):
             parser.error("--jobs argument must be greater than zero!")
         extra.extend(['-j', str(args.jobs)])
 
-    package = packages.walkPackagePath(args.package)
+    if generator.get('query'):
+        package = packages.queryPackagePath(args.package)
+    else:
+        package = packages.walkPackagePath(args.package)
+        print(">>", colorize("/".join(package.getStack()), "32;1"))
 
     # execute a bob dev with the extra arguments to build all executables.
     # This makes it possible for the plugin to collect them and generate some runTargets.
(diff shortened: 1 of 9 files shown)


pym/bob/cmds/build/project.py
```diff
@@ -107,7 +107,11 @@ def _downloadArg(arg):
             parser.error("--jobs argument must be greater than zero!")
         extra.extend(['-j', str(args.jobs)])
 
-    package = packages.walkPackagePath(args.package)
+    if generator.get('query'):
+        package = packages.queryPackagePath(args.package)
```
jkloetzke, line 111: I guess a simple test for such a new generator should be added.

rhubert: > I'm wondering how much effort it would be to support queries in the existing generators...

I'm not sure what the expected result is for the existing generators? ATM we generate one project in `projects` for the given package. If a query matches multiple packages we would generate one project for each match?

rhubert:
> I'm wondering how much effort it would be to support queries in the existing generators...

I'm not sure what the expected result is for the existing generators? ATM we generate one project in `projects` for the given package. If a query matches multiple packages we would generate one project for each match?

jkloetzke:
> > I'm wondering how much effort it would be to support queries in the existing generators...
> 
> I'm not sure what the expected result is for the existing generators? ATM we generate one project in `projects` for the given package. If a query matches multiple packages we would generate one project for each match?

You're right. Didn't think about this point. I guess we leave them as they are...

rhubert:
Hmm...What's wrong with the CI? Why is the bob-version `0.21+unknown`??
```
Parse error: Your Bob is too old. Plugin 'plugins/g2.py' requires at least version 0.22! Bob-Version: 0.21+unknown
```

rhubert:
...the test failed because shallow cloning was enabled on the CI since b06dfa4ddfa80a1a59e05d4309d3ccc864049c98 🤔

jkloetzke:
> ...the test failed because shallow cloning was enabled on the CI since [b06dfa4](https://github.com/BobBuildTool/bob/commit/b06dfa4ddfa80a1a59e05d4309d3ccc864049c98) 🤔

By bad. I forgot to bump the fallback version. I've fixed it and rebased your PR. Thanks!

Issue #331, opened 2020-05-12
Implement retries for URL SCM downloads


mahaase:
Hey,

this problem I have since the very beginning: 

```
+ '[' -e .state/sandbox_ubuntu-x86_64_src_1-f301-98ca.canary ']'
+ echo f63302148dced4ef2ec67d869b958dd99da61e1b
+ bob _invoke sandbox_ubuntu-x86_64_src_1-f301-98ca.spec.json -vv
+ <wget> http://fdt-c-pcs-0004.fdtech.intern/fileserver/ubuntu-bionic-amd64-rootfs.tgz > ./ubuntu-bionic-amd64-rootfs.tgz
Error: Response too short: 2147479552 < 2797978679 (bytes)
Error: Recipe sandbox::ubuntu-x86_64 in checkoutSCM: dir:., url:http://fdt-c-pcs-0004.fdtech.intern/fileserver/ubuntu-bionic-amd64-rootfs.tgz in checkoutSCM: dir:., url:http://fdt-c-pcs-0004.fdtech.intern/fileserver/ubuntu-bionic-amd64-rootfs.tgz failed
Build step 'Execute shell' marked build as failure
Archiving artifacts
Finished: FAILURE
```

I tried to configure the nginx fileserver with some stuff, but no success at all.
The archives, which **sometimes** fails, are >1,5 GB.

If I restart the job, in most cases it will work than..

calling 
`curl -sSg --fail -o sandbox_ubuntu-x86_64_dist_1-68ac-e703.tgz http://fdt-c-pcs-0004.fdtech.intern/artifact/f1/bb/21909256750d2d6038bbf148e3c7ce8fd9de-1.tgz`
 manually will work everytime...

Any ideas?

BR.

jkloetzke:
Any chance that the file is downloaded or served from a 32-bit Linux installation? The file length suspiciously looks like the 2 GiB barrier: exactly 2GiB - 4kiB!

What do you mean by "since the very beginning"? That code was refactored from calling `curl` to using Python. Does the problem exist since the rewrite or was it already there when `curl` was still used?

That put aside I'll look into how we could gracefully handle the error...

daxcore:
No there are no 32-bit machines in company.
"since the very beginning" means, since I use large sandboxes, also before with curl!

Maybe the nginx webserver is not the best solution for that, maybe I should try another one, or switch to different protocol like ftp? could bob handle ftp?

jkloetzke:
[...]

PR #517, opened 2023-06-10
Fetch retry

rhubert:


pym/bob/scm/git.py
```diff
@@ -294,7 +294,7 @@ async def __checkoutTag(self, invoker, fetchCmd, switch):
         head = await invoker.callCommand(["git", "rev-parse", "--verify", "-q", "HEAD"],
             stdout=False, cwd=self.__dir)
         if head or switch:
-            await invoker.checkCommand(fetchCmd, cwd=self.__dir)
+            await invoker.checkCommand(fetchCmd, retries=10, cwd=self.__dir)
```
jkloetzke, line 297: I find 10 a little bit over the top. Maybe 3? And what about the other `git fetch` invocations? I guess there we should also retry...

rhubert: Yesterday it failed with 10 retries... :( Maybe this needs to be a scm-property. I also added a sleep(5) after each retry but without success. 🤷

jkloetzke: Hmm, my assumption is that if it did not work with three retries then it most probably won't work with 10 retries either. BTW, the retries are probably also needed for `ls-remote` calls.

Making the retries configurable sounds like a good idea. I think a global retry count is probably sufficient. If there is ever a need to make it configurable per-SCM, we can still introduce a property...

jkloetzke:
BTW, there is a related issue for URL SCMs: #331

pym/bob/utils.py
```diff
@@ -741,7 +741,8 @@ def __exit__(self, exc_type, exc_value, traceback):
         asyncio.set_event_loop(None)
 
 
-async def run(args, universal_newlines=False, errors=None, check=False, shell=False, **kwargs):
```
jkloetzke, line 744: This isn't used, is it?

pym/bob/invoker.py
```diff
@@ -509,14 +510,26 @@ async def runCommand(self, args, cwd=None, **kwargs):
...
 
-    async def checkCommand(self, args, cwd=None, **kwargs):
+    async def checkCommand(self, args, cwd=None, retries=0, **kwargs):
         cwd = os.path.join(self.__cwd, cwd) if cwd else self.__cwd
-        await self.__runCommand(args, cwd, check=True, **kwargs)
+        while retries != 0:
+            ret = await self.__runCommand(args, cwd, **kwargs)
+            if ret.returncode == 0:
+                return
+            time.sleep(3)
```
jkloetzke, line 529: This is an async function. You need to use 
```suggestion
            await asyncio.sleep(3)
```

pym/bob/invoker.py
```diff
@@ -509,14 +510,26 @@ async def runCommand(self, args, cwd=None, **kwargs):
...
 
-    async def callCommand(self, args, cwd=None, **kwargs):
+    async def callCommand(self, args, cwd=None, retries=0, success=0, **kwargs):
         cwd = os.path.join(self.__cwd, cwd) if cwd else self.__cwd
-        ret = await self.__runCommand(args, cwd, **kwargs)
+        while True:
+            ret = await self.__runCommand(args, cwd, **kwargs)
+            if (ret == success) or (retries == 0):
+                break
+            time.sleep(3)
```
jkloetzke, line 519: This is an async function. You need to use 
```suggestion
            await asyncio.sleep(3)
```

jkloetzke (commented):
Now that it's an SCM property I guess we can also add it the git and url schemas.

It would be good to have some tests. For the url-SCM there is a unit test with a HTTP server mock that can be adapted. But for git I have no good idea at the moment. If don't have one either, I'm fine with merging even without git retry test...

pym/bob/scm/git.py
```diff
@@ -197,7 +200,8 @@ async def invoke(self, invoker, switch=False):
                 realUrl = await invoker.checkOutputCommand(
                     ["git", "ls-remote", "--get-url", remote], cwd=self.__dir)
                 if cfgUrl != realUrl:
-                    await invoker.checkCommand(["git", "remote", "set-url", remote, cfgUrl], cwd=self.__dir)
+                    await invoker.checkCommand(["git", "remote", "set-url", remote, cfgUrl],
```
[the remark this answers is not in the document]
rhubert: hmm...not sure why ... I guess I wanted to add the retry the `ls-remote` one line above ... 😕

pym/bob/scm/git.py
```diff
@@ -659,28 +663,34 @@ def hasJenkinsPlugin(self):
...
+            time.sleep(3)
+            retries -= 1
+        return ""
 
     def status(self, workspacePath):
         status = ScmStatus()
         try:
             onCorrectBranch = False
-            output = self.callGit(workspacePath, 'ls-remote' ,'--get-url')
+            output = self.callGit(workspacePath, 'ls-remote'
```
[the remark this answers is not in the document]
rhubert: > BTW, the retries are probably also needed for `ls-remote` calls.

🤔  I have never seed a `ls-remote` failing. It always failed on `fetch`... I'll remove the retries from any other commands..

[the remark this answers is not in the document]
jkloetzke: Well, the `ls-remote` in `predictLiveBuildId()` could fail because it will talk to the server. I would keep the retries there.

pym/bob/utils.py
rhubert: It seams to be used by `predictLiveBuildId` ->  `await check_output`   -> `await check_output` -> `run`

jkloetzke: Aaahhh, of course. Never mind.

jkloetzke:
Thanks!

PR #518, opened 2023-06-16
git: filter useCommitOnBranch in canSwitch

rhubert:
It's still possible to do inline switching if the useCommitOnBranch property changes.


jkloetzke:
Thanks!

PR #519, opened 2023-06-21
Use internal extractors on win32

jkloetzke:
Turns out that at least tarfile is not usable on POSIX systems for us due to the `chmod` handling. Play safe and use it only on native Windows where this doesn't make a difference. Given that, we can also prefer the Python versions there.


jkloetzke:
@sbixl Could you give this a try?

sbixl:
Sorry for the delay. I will test it next week and give you feedback soon.

sbixl:
Did some testing today and it works like expected. Thanks for that improvement!

jkloetzke:
Thanks for testing!

PR #521, opened 2023-06-25
Run changed checkoutScript in build-only mode

jkloetzke:
If a recipe utilizes `checkoutUpdateIf`, relax the checks so that also modified `checkoutScript`s are run in build-only mode.

@rhubert  Could you please give it a try?

:warning: This change bumps the project state version. Once used in a build tree, there is no going back to an older Bob version! So maybe better test it in a new build tree.


rhubert:
:+1: Did some testing today -> works as expected now.

jkloetzke:
Thanks for testing!

PR #525, opened 2023-08-10
generators: fix filter option and error reporting

rhubert:
Fix for exceptions occurring when using the filter option:
```
  File "pym/bob/generators/common.py", line 309, in configure
    additionalFiles = re.compile(args.filter) if self.args.filter else None
  NameError: name 'args' is not defined

  File "pym/bob/generators/common.py", line 311, in configure
    raise ParseError("Invalid regular expression '{}': {}".format(e.pattern), e)
  NameError: name 'ParseError' is not defined

  File "pym/bob/generators/common.py", line 311, in configure
    raise ParseError("Invalid regular expression '{}': {}".format(e.pattern), e)
  IndexError: tuple index out of range
```

### DIFF
--- a/pym/bob/generators/common.py
+++ b/pym/bob/generators/common.py
@@ -3,7 +3,7 @@
 #
 # SPDX-License-Identifier: GPL-3.0-or-later
 
-from ..errors import BuildError
+from ..errors import BuildError, ParseError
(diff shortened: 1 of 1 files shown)


jkloetzke:
Thanks!

PR #526, opened 2023-08-11
msys: fixed version handling

mahaase:
there is no CYGWIN_VERSION_DLL_EPOCH anymore. epoch and major has to be calculated. define CYGWIN_VERSION_DLL_MAJOR 3004 means version 3.4.x.

### DIFF
--- a/pym/bob/languages.py
+++ b/pym/bob/languages.py
@@ -45,8 +45,13 @@ int main(void)
 {
 #if defined(CYGWIN_VERSION_DLL_IDENTIFIER)
     printf("%s/%d.%d.%d\n", CYGWIN_VERSION_DLL_IDENTIFIER,
+#ifdef CYGWIN_VERSION_DLL_EPOCH
         CYGWIN_VERSION_DLL_EPOCH,
         CYGWIN_VERSION_DLL_MAJOR,
+#else
+        CYGWIN_VERSION_DLL_MAJOR / 1000,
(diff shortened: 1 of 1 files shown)


jkloetzke:
It looks like `CYGWIN_VERSION_DLL_EPOCH` was simply removed with no apparent reasoning given. :thinking:  I would propose to use `-1` in case it's undefined to keep compatibility with older installations:

```c
#ifndef CYGWIN_VERSION_DLL_EPOCH
#define CYGWIN_VERSION_DLL_EPOCH -1
#endif
```

Otherwise the fingerprint would not be backwards compatible...

mahaase:
maybe like this?
```
printf("%s/%d.%d.%d\n", CYGWIN_VERSION_DLL_IDENTIFIER,
#ifdef CYGWIN_VERSION_DLL_EPOCH
        CYGWIN_VERSION_DLL_EPOCH,
        CYGWIN_VERSION_DLL_MAJOR,
#else
        CYGWIN_VERSION_DLL_MAJOR / 1000,
        CYGWIN_VERSION_DLL_MAJOR % 1000,
#endif
        CYGWIN_VERSION_DLL_MINOR);
```

jkloetzke:
Sure, that's also possible...

mahaase:
updated. 👍

jkloetzke:
Thanks!

PR #530, opened 2023-08-23
build: add no-force-branch switch

rhubert:
A common workflow is to:
- write some code in a components repo
- push it to a feature branch
- write the new commitID of the component into the recipe
- use bob dev ...

With activated gitCommitOnBranch policy this is no longer possible if there is also a branch configured in the recipe as the build bails out with something like:
```
  Branch XXX does not contain abcdef.
```

With `--no-force-branch`  the branch is ignored and the commit is checked out (results in detached HEAD) making it possible to resume work without a scmOverride.


rhubert:
I've another idea how this workflow could be improved... closing this...

PR #531, opened 2023-08-23
git: fix scm switch when going back to older revision

rhubert:
If gitCommitOnBranch is active and commit and branch are configured in the recipe changing the commit back to a older one did not update the sources back to the configured version. Instead the remained on the current version.

### DIFF
--- a/pym/bob/scm/git.py
+++ b/pym/bob/scm/git.py
@@ -295,14 +295,30 @@ class GitScm(Scm):
         else:
             # We're switching the ref and the branch exists already. Be extra
             # careful: the user might have committed to this branch, some other
-            # branch might be checked out currently or both of that. To keep
-            # things simple, assume a fast-forward of the commit on the branch.
-            # It will catch user changes and is usually safe wrt. submodules.
+            # branch might be checked out currently or both of that.
             await invoker.checkCommand(["git", "checkout", "--no-recurse-submodules",
                 self.__branch], cwd=self.__dir)
             preUpdate = await self.__updateSubmodulesPre(invoker)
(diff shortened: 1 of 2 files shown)


jkloetzke:
I think the goal should be to remove `git merge --ff-only` and replace it by `git reset --mixed $commit/$tag`. The problem is that we must make sure that no commit of the user is lost. Whenever *any* remote branch holds the current commit (after switching to the branch) then we're safe. I *think* calling `git branch --remotes --contains HEAD` (before `git reset`) and verify that it produced some output before should be sufficient...

rhubert:
I think `git reset --mixed` is wrong as it would leave all changes uncommitted in the workspace. 

If there are 3 commits:
```
- A - B - C  (HEAD)
```
`git reset --mixed B` would leave `C`s-changes in the workspace (uncommitted).

I think we need to use `git reset --hard` if `git status --porcelain` returns nothing and `C` is on any other local or a remote branch. Otherwise move to attic.

jkloetzke:
Right, my bad. Obviously I was not reading the manpage carefully enough. :see_no_evil: But `git reset --hard` does feel too dangerous. What about `git reset --keep`? It moves the branch forward and backwards while aborting the operation if a file has changes that needs to be updated. In this case Bob will move the repository into the attic as expected. OTOH unrelated files do not prohibit updates. Sounds safe to me.

But as already said, we have to make sure that the old commit is not lost. Maybe `git branch --remotes --contains HEAD` is a bit too pessimistic. The user might have another local branch. My current idea is to run `git branch -a --format='%(refname:lstrip=2)' --contains HEAD` and remove the current branch name from the list. Thoughts?

jkloetzke:
Thanks for bearing with me. :)

PR #532, opened 2023-09-01
git: relax the gitCommitOnBranch check

rhubert:
If a repo was moved to a new commit and the commitID was updated in the recipe to the same commit, the previous behavior was to switch to the new commit running the checkout again including all checks.

In pratice there are cases where the new commit hasn't been pushed to the configured branch when running bob again. The commitOnBranch check noticed that the commit did not match the branch and the build fails. To recover from this a developer needs to either move it's stuff back from the attic, write a scmOverride, ... which is annoying and not really necessary as the workspace is already on the correct commit.

To make this more user friendly relax the "Branch ... did not contain .." error if the workspace is on the configured commit and leave the workspace untouched.

### DIFF
--- a/pym/bob/scm/git.py
+++ b/pym/bob/scm/git.py
@@ -247,10 +247,23 @@ class GitScm(Scm):
     async def __checkoutTagOnBranch(self, invoker, fetchCmd, switch):
         # Only do something if nothing is checked out yet or a forceful switch
         # is requested.
-        headValid = (await invoker.callCommand(["git", "rev-parse", "--verify",
-            "-q", "HEAD"], stdout=False, cwd=self.__dir)) == 0
-        if headValid and not switch:
-            return
+        rev_parse = await invoker.runCommand(["git", "rev-parse", "--verify",
+            "-q", "HEAD"], stdout=True, cwd=self.__dir)
+        headValid = rev_parse.returncode == 0
+        if headValid:
+            if not switch:
+                return
+            # if the workspace is already on the correct commit / tag do nothing
+            if self.__commit:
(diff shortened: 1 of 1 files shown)


pym/bob/scm/git.py
```diff
@@ -262,6 +263,10 @@ async def __checkoutTagOnBranch(self, invoker, fetchCmd, switch):
         if await invoker.callCommand(["git", "merge-base", "--is-ancestor",
                                      commit, "remotes/origin/"+self.__branch],
                                      cwd=self.__dir):
+            # already on the configured commit -> do nothing
+            if self.__commit and rev_parse.stdout.rstrip() == self.__commit:
+                invoker.warn("Not updating", self.__dir, ". Already on the configured commit.")
+                return
```
jkloetzke, line 269: I think this could be moved even before the `fetchCmd` above. It doesn't even need a warning because the desired commit is already checked out. Could you also add the case for a tag, e.g. something along these lines?

```python
if self.__commit ...
    ...
else:
    # Convert tag to commit. Beware of annotated commits!
    tag = await invoker.runCommand(["git", "rev-parse", "tags/"+self.__tag+"^0"], stdout=True, cwd=self.__dir)
    if tag.returncode == 0 and tag.stdout == rev_parse.stdout:
        return
```

I'm wondering if this should all be done only if `headValid == True`..?

rhubert: Thanks. I modified the code following you're suggestions. My assumption was that the output of `git rev-parse` did not match the commit/tag if the HEAD is not valid but this might be wrong. Moving behind the `headValid` should be safe.

jkloetzke:
Thanks!

PR #537, opened 2023-11-13
git: inline switch when reference changes

rhubert:
There is no need to move the workspace to attic if the reference or dissociate are changed.


jkloetzke:
Thanks!

PR #541, opened 2023-11-23
intermediate: support colon seperated tool paths

rhubert:
Make it possible to define tools with more than one directory containing executables by handling colons in path as separator.

### DIFF
--- a/doc/manual/configuration.rst
+++ b/doc/manual/configuration.rst
@@ -558,7 +558,8 @@ automatically:
   ``${BOB_DEP_PATHS[libfoo-dev]}``) instead of the position (e.g. ``$2``).
 * ``BOB_TOOL_PATHS``: An associative array that holds the execution paths to
   consumed tools indexed by the package name. All these paths are in ``$PATH``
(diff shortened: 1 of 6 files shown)


rhubert:
I think this basically works. 

The issue I faced was a package with multiple tools installed in different directories. Symlinking them was not possible as these tools tried to open files relative to the symlink...

Anyway I found another solution to solve my problem. Not sure if this is needed. What do you think?

jkloetzke:
How about to just declare multiple tools? Something like this:

```yaml
provideTools:
    foo: path/a
    bar: path/b
```

Then a consuming recipe can choose if it needs `foo`, `bar` or both. If there are multiple tools then they should have their own name. I would even introduce different names if multiple tools are in exactly the same directory. This makes refactoring in the future much easier.

I don't think that the proposed solution is viable. The original design is to have one path per tool name. Attaching an incrementing suffix looks kind of arbitrary. What should happen if this creates a name collision? The numbering will change too if some additional path is added in the middle. So all in all I would rather not go this route.

PR #550, opened 2024-01-15
Fix audit execution

### LINKED ISSUE #539: Build steps can get finished with large delay in multi job mode
If Bob runs in multi job mode (with option `-j`), task scheduling is active, which currently (Bob v0.23) has some unexpected side effects.

When the script of a build step is finished, this is indicated on the TTY interface by shifting the corresponding line
```
[    ] BUILD ...
```
from the stage of running jobs to the stage of finished jobs. However, this only shows that the task for running the build script has finished, but not the complete build step. To finish the build step, the audit trail has to be generated at the end (unless Bob is invoked with `--no-audit`). This is an independent task, indicated on the TTY interface with a separate line
```
[    ] AUDIT ...
```

Currently the tasks for audit trail generation are first scheduled after all build scripts that can be executed independently. In a configuration with many independent and long running build steps, this can be very late. Moreover this means that each of these independent build steps only finishes after all the other build scripts have finished and the tasks for audit trail generation have run.

This has at least the following observed side effects:

1. When the build process is cancelled and resumed with option `--resume`, the complete build step is restarted, if the audit trail generation task did not run yet, even when the build scripts were already finished. This can be very annoying in large builds with long running build scripts.
2. The progress indicator, which shows the number and percentage of done tasks is updated too late. It is first updated, when the build step is finished and not the build script.

jkloetzke:
Don't delay the execution of audits to finish jobs as soon as possible when the build/package script has returned. Also include the audit into a visible action so that the user knows the build/package step has not finished yet.

Fixes #539

### DIFF
--- a/pym/bob/builder.py
+++ b/pym/bob/builder.py
@@ -1753,14 +1753,14 @@ cd {ROOT}
                                       step, False, self.__buildIdTasks, False)
                 for step in steps
             ]
-            ret = await self.__yieldJobWhile(gatherTasks(tasks), True)
+            ret = await self.__yieldJobWhile(gatherTasks(tasks))
         else:
             tasks = []
             for step in steps:
                 task = self.__createCookTask(lambda s=step: self.__getBuildIdTask(s, depth),
                                              step, False, self.__buildIdTasks, False)
                 tasks.append(task)
-                await self.__yieldJobWhile(asyncio.wait({task}), True)
+                await self.__yieldJobWhile(asyncio.wait({task}))
             # retrieve results as last step to --keep-going
             ret = [ t.result() for t in tasks ]
         return ret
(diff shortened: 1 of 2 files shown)


pym/bob/builder.py
```diff
@@ -1265,7 +1261,7 @@ async def _cookCheckoutStep(self, checkoutStep, depth):
         # Generate audit trail before setResultHash() to force re-generation if
         # the audit trail fails.
         if checkoutHash != oldCheckoutHash or self.__force:
-            await self._generateAudit(checkoutStep, depth, checkoutHash, checkoutExecuted)
+            await self._generateAudit(checkoutStep, depth, checkoutHash, checkoutHash, checkoutExecuted)
```
gruberchr, line 1264: The implicit identity of `checkoutHash` with the build ID of the checkout step is quite confusing, if one compares the invocation of `_generateAudit()` within the three functions `_cookCheckoutStep()`, `_cookBuildStep()` and `_cookPackageStep()`. From my point of view this should be made explicit by introducing a variable `checkoutBuildId` consistently with `buildBuildId` and `packageBuildId`, which is set to `checkoutHash`. This makes the code more readable. Maybe this identity could be commented in the code.

jkloetzke: Good suggestion! Adapted accordingly...

PR #556, opened 2024-03-05
builder: pass process error code to BuildError

KevinSlijepcevic:
This change shall enable Bob to return a process error code in an invoked process to the caller of a Bob command, e.g. `bob dev test`.

Afterwards, it shall be possible to evaluate the returned error code with `echo $?`, which currently always returns `1`.


jkloetzke:
Thanks!

PR #559, opened 2024-03-26
Let sandbox supports custom user information settings

yuanxingyang:
This patch is used to address the issue of needing to use a specific user ID during the compilation phase to connect to the server and validate the compiler license legality, e.g. QNX project qcc compiler need to check license via online server

### DIFF
--- a/test/black-box/sandbox/recipes/root.yaml
+++ b/test/black-box/sandbox/recipes/root.yaml
@@ -9,7 +9,8 @@ checkoutScript: |
     echo ok
 buildScript: |
     echo ok
-packageVars: [FOO]
+packageVars: [FOO, EXPECT_UID]
 packageScript: |
     test $FOO = bar
+    test ${EXPECT_UID:-65534} = $UID
     echo ok
--- a/doc/manual/configuration.rst
+++ b/doc/manual/configuration.rst
@@ -1750,6 +1750,13 @@ variables. See :ref:`configuration-principle-subst` for the available
 substations. The mount paths are also subject to an additional variable
 expansion when a step using the sandbox *is actually executed*. This can be
 useful e.g. to expand variables that are only available on the build server.
+
+By default, the user ID inside the sandbox is ``nobody``. The optional ``user``
+key allows to use two other identities: ``root`` or ``$USER``.  Note that using
+``root`` does not provide any more privileges. It merely maps the current user
+ID to the root user ID inside the sandbox. The ``$USER`` option keeps the
+current user ID when entering the sandbox. No other values are allowed.
+
 Example::
 
     provideSandbox:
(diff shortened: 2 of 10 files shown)


yuanxingyang:
The sandbox recipes will be like the following:

`provideSandbox:
   paths: [ "/bin", "/usr/bin", "/sbin","usr/include"]
   mount:
       - [ "\\$HOME/.ssh", "\\$HOME/.ssh"]
       - [ "\\$HOME/.qnx", "\\$HOME/.qnx",[rw]]
       - [ "\\$HOME/.flexlmrc", "\\$HOME/.flexlmrc"]
       - [ "\\$HOME/qnx", "\\$HOME/qnx"]
   user: 
     uid: "\\$UID"
     gid: "\\$UID" `

jkloetzke:
Thanks for the PR. Interesting that the license check is somehow influenced by the user/group id. Anyway, I think we shall support this.

I don't think we need a fully generic solution here, do we? Currently I see three sensible behaviours:

* Run as user "nobody" in the sandbox. This is currently the default.
* Run with the same user/group-Id. This seems to be you use case.
* Run as fake root in the sandbox. Already supported by `namespace-sandbox` but not exposed in Bob.

I would propose a new sandbox property that allows just the three cases above:

```yaml
provideSandbox:
    user: "nobody" / "current" / "root"
```

Would you mind to update the PR in this direction? A couple of additional thoughts:

* It should be possible to adjust the behaviour in `default.yaml` too (`sandbox` key)
* The changes to the namespace-sandbox should be separate commits
* Some words in the documentation describing the new option should be added too
* Ideally some smoke test would be nice (maybe extend `test/black-box/sandbox/...` a bit?)

yuanxingyang:
I don't mind and I'm very happy if you can make direct modifications on this PR.

 I have a few ideas, see if you can adopt them：
1.  Use '\\$USER' instead of 'current', this way it's more intuitive and easier to understand
provideSandbox:
    user: "nobody" / "\\$USER" / "root"

2.  When obtaining information about the current user, avoid accessing the /etc/passwd file. In some organizations, accounts are managed through the Active Directory mechanism, so there is no user information in the /etc/passwd file

jkloetzke:
Please have a look at the current state of the PR. Let me know if it fits your needs...

test/black-box/sandbox/recipes/root.yaml
```diff
@@ -9,7 +9,8 @@ checkoutScript: |
     echo ok
 buildScript: |
     echo ok
-packageVars: [FOO]
+packageVars: [FOO, EXPECT_UID]
```
yuanxingyang, line 12: Is it more reasonable to use packageVarsWeak: [EXPECT_UID] instead of packageVars: [EXPECT_UID], because the UID is just for permission validation and does not affect the result?

yuanxingyang:
> Please have a look at the current state of the PR. Let me know if it fits your needs...

This PR perfectly implements my requirements, and I have validated it locally without any issues.

test/black-box/sandbox/recipes/root.yaml
jkloetzke: In principle you're right but in this particular test it is part of the result. If the variable would be weak, Bob would not rebuild the package if just the expected uid changes.

PR #561, opened 2024-04-02
bob managed layers

rhubert:
This is a very rough first draft of support for layers handled by bob to get some early comments about this.

The idea is to add a scm-spec to the `layers` section of the `config.yaml`:

```yaml
layers:
  - foo:
     scm: git
     url: git@foo:/foo.git
     branch: master
     commit: ....
```

A layer specified like this also changes the behavior when using nested layers where multiple layers depend on the same layer as the layer structure is flattened and each layer is checked out only once. This make it possible to have the same recipes either used as root-recipes repo or as layer from another recipes package. (ATM this is done by matching the name of the layer only, maybe the url should be used instead or a additional `uuid`,... ??)

ATM a internal `.bob_layers` root package is generated for this with dependencies to a package for each layer. I don't know if this is a good idea but it enables reusing most of the code. ;)
 
If this is a viable approach it would need some filtering for the other commands like `bob ls`  to avoid the `.bob_layers` package is shown there as well. Maybe this filtering can be done by conditionally adding the `.bob_layers` to the `virtualRoot` and providing the define by the layers-command? But this would result in longer parsing times when switching from `bob layers` to any other bob command.

### DIFF
--- a/pym/bob/input.py
+++ b/pym/bob/input.py
@@ -2991,6 +2991,7 @@ class RecipeSet:
         self.__commandConfig = {}
         self.__uiConfig = {}
         self.__shareConfig = {}
+        self.__layers = []
         self.__policies = {
             'noUndefinedTools' : (
                 "0.17.3.dev57",
@@ -3431,7 +3432,7 @@ class RecipeSet:
                 os.path.join(os.path.expanduser("~"), '.config')), 'bob', 'default.yaml'))
 
         # Begin with root layer
-        self.__parseLayer([], "9999", recipesRoot)
+        self.__parseLayer("", "9999", recipesRoot)
 
         # Out-of-tree builds may have a dedicated default.yaml
         if recipesRoot:
@@ -3475,9 +3476,13 @@ class RecipeSet:
         self.__addRecipe(self.__rootRecipe)
 
     def __parseLayer(self, layer, maxVer, recipesRoot):
-        rootDir = os.path.join(recipesRoot, *(os.path.join("layers", l) for l in layer))
+        if layer in self.__layers:
+            return
+        self.__layers.append(layer)
+
+        rootDir = os.path.join(recipesRoot, os.path.join("layers", layer) if layer != "" else "")
         if not os.path.isdir(rootDir or "."):
-            raise ParseError("Layer '{}' does not exist!".format("/".join(layer)))
+            raise ParseError(f"Layer '{layer}' does not exist!")
 
         configYaml = os.path.join(rootDir, "config.yaml")
         def preValidate(data):
@@ -3495,7 +3500,7 @@ class RecipeSet:
             preValidate=preValidate)
         minVer = config.get("bobMinimumVersion", "0.16")
         if compareVersion(maxVer, minVer) < 0:
-            raise ParseError("Layer '{}' reqires a higher Bob version than root project!"
+            raise ParseError("Layer '{}' requires a higher Bob version than root project!"
                                 .format("/".join(layer)))
         if compareVersion(minVer, "0.16") < 0:
             raise ParseError("Projects before bobMinimumVersion 0.16 are not supported!")
@@ -3518,7 +3523,7 @@ class RecipeSet:
         # First parse any sub-layers. Their settings have a lower precedence
         # and may be overwritten by higher layers.
         for l in config.get("layers", []):
-            self.__parseLayer(layer + [l], maxVer, recipesRoot)
+            self.__parseLayer(l, maxVer, recipesRoot)
 
         # Load plugins and re-create schemas as new keys may have been added
         self.__loadPlugins(rootDir, layer, config.get("plugins", []))
--- a/pym/bob/input.py
+++ b/pym/bob/input.py
@@ -1718,6 +1718,36 @@ class ScmValidator:
                 None)
         return data
 
+class LayerSpec:
+    def __init__(self, name, scm=None):
+        self.__name = name;
+        self.__scm = scm
+
+    def getName(self):
+        return self.__name
+
+    def getScm(self):
+        return self.__scm
+
+class LayerValidator:
+    def __init__(self):
+        self.__scmValidator = ScmValidator({
+            'git' : GitScm.SCHEMA,
+            'svn' : SvnScm.SCHEMA,
+            'cvs' : CvsScm.SCHEMA,
+            'url' : UrlScm.SCHEMA})
+
+    def validate(self, data):
+        if isinstance(data,str):
+            return LayerSpec(data)
+        if 'name' not in data:
+            raise schema.SchemaMissingKeyError("Missing 'name' key in {}".format(data), None)
+        _data = data.copy();
+        name = _data.get('name')
+        del _data['name']
+
+        return LayerSpec (name, self.__scmValidator.validate(_data)[0])
+
 class VarDefineValidator:
     def __init__(self, keyword):
         self.__varName = re.compile(r'^[A-Za-z_][A-Za-z0-9_]*$')
@@ -1947,7 +1977,7 @@ class Recipe(object):
             "buildScript" : "true",
             "packageScript" : "true"
         }
-        ret = Recipe(recipeSet, recipe, [], "", ".", "", "", properties)
+        ret = Recipe(recipeSet, recipe, "", "", ".", "", "", properties)
         ret.resolveClasses(Env())
         return ret
 
@@ -2914,7 +2944,30 @@ class RecipeSet:
             schema.Optional('max_depth') : int,
         })
 
-    STATIC_CONFIG_SCHEMA = schema.Schema({
+    SCM_SCHEMA = ScmValidator({
+        'git' : GitScm.SCHEMA,
+        'svn' : SvnScm.SCHEMA,
+        'cvs' : CvsScm.SCHEMA,
+        'url' : UrlScm.SCHEMA,
+        'import' : ImportScm.SCHEMA,
+    })
+
+    STATIC_CONFIG_LAYER_SPEC = {
+        schema.Optional('layersWhitelist') : [str],
+        schema.Optional('layersScmOverrides') : schema.Schema([{
+            schema.Optional('match') : schema.Schema({ str: object }),
+            schema.Optional('del') : [str],
+            schema.Optional('set') : schema.Schema({ str: object }),
+            schema.Optional('replace') : schema.Schema({
+                str : schema.Schema({
+                    'pattern' : str,
+                    'replacement' : str
+                })
+            })
+        }])
+    }
+
+    STATIC_CONFIG_SCHEMA_SPEC = {
         schema.Optional('bobMinimumVersion') : str, # validated separately in preValidate
         schema.Optional('plugins') : [str],
         schema.Optional('policies') : schema.Schema(
@@ -2939,23 +2992,48 @@ class RecipeSet:
             },
             error="Invalid policy specified! Are you using an appropriate version of Bob?"
         ),
-        schema.Optional('layers') : [str],
+        schema.Optional('layers') : [LayerValidator()],
         schema.Optional('scriptLanguage',
                         default=ScriptLanguage.BASH) : schema.And(schema.Or("bash", "PowerShell"),
                                                                   schema.Use(ScriptLanguage)),
-    })
-
-    SCM_SCHEMA = ScmValidator({
-        'git' : GitScm.SCHEMA,
-        'svn' : SvnScm.SCHEMA,
-        'cvs' : CvsScm.SCHEMA,
-        'url' : UrlScm.SCHEMA,
-        'import' : ImportScm.SCHEMA,
-    })
+    }
 
     MIRRORS_SCHEMA = ScmValidator({
         'url' : UrlScm.MIRRORS_SCHEMA,
     })
+
+    POLICIES = {
+        'noUndefinedTools' : (
+            "0.17.3.dev57",
+            InfoOnce("noUndefinedTools policy not set. Included but undefined tools are not detected at parsing time.",
+                help="See http://bob-build-tool.readthedocs.io/en/latest/manual/policies.html#noundefinedtools for more information.")
+        ),
+        'scmIgnoreUser' : (
+            "0.17.3.dev97",
+            InfoOnce("scmIgnoreUser policy not set. Authentication part URL is tainting binary artifacts.",
+                help="See http://bob-build-tool.readthedocs.io/en/latest/manual/policies.html#scmignoreuser for more information.")
+        ),
+        'pruneImportScm' : (
+            "0.17.3.dev102",
+            InfoOnce("pruneImportScm policy not set. Incremental builds of 'import' SCM may lead to wrong results.",
+                help="See http://bob-build-tool.readthedocs.io/en/latest/manual/policies.html#pruneimportscm for more information.")
+        ),
+        'gitCommitOnBranch' : (
+            "0.21.0.dev5",
+            InfoOnce("gitCommitOnBranch policy not set. Will not check if commit / tag is on configured branch.",
+                help="See http://bob-build-tool.readthedocs.io/en/latest/manual/policies.html#gitcommitonbranch for more information.")
+        ),
+        'fixImportScmVariant' : (
+            "0.22.1.dev34",
+            InfoOnce("fixImportScmVariant policy not set. Recipe variant calculation w/ import SCM is boguous.",
+                help="See http://bob-build-tool.readthedocs.io/en/latest/manual/policies.html#fiximportscmvariant for more information.")
+        ),
+        "defaultFileMode": (
+            "0.24rc1",
+            InfoOnce("defaultFileMode policy not set. File mode of URL SCMs not set for locally copied files.",
+                help="See http://bob-build-tool.readthedocs.io/en/latest/manual/policies.html#defaultfilemode for more information.")
+        ),
+    }
 
     _ignoreCmdConfig = False
     @classmethod
@@ -2992,38 +3070,7 @@ class RecipeSet:
         self.__uiConfig = {}
         self.__shareConfig = {}
         self.__layers = []
-        self.__policies = {
-            'noUndefinedTools' : (
-                "0.17.3.dev57",
-                InfoOnce("noUndefinedTools policy not set. Included but undefined tools are not detected at parsing time.",
-                    help="See http://bob-build-tool.readthedocs.io/en/latest/manual/policies.html#noundefinedtools for more information.")
-            ),
-            'scmIgnoreUser' : (
-                "0.17.3.dev97",
-                InfoOnce("scmIgnoreUser policy not set. Authentication part URL is tainting binary artifacts.",
-                    help="See http://bob-build-tool.readthedocs.io/en/latest/manual/policies.html#scmignoreuser for more information.")
-            ),
-            'pruneImportScm' : (
-                "0.17.3.dev102",
-                InfoOnce("pruneImportScm policy not set. Incremental builds of 'import' SCM may lead to wrong results.",
-                    help="See http://bob-build-tool.readthedocs.io/en/latest/manual/policies.html#pruneimportscm for more information.")
-            ),
-            'gitCommitOnBranch' : (
-                "0.21.0.dev5",
-                InfoOnce("gitCommitOnBranch policy not set. Will not check if commit / tag is on configured branch.",
-                    help="See http://bob-build-tool.readthedocs.io/en/latest/manual/policies.html#gitcommitonbranch for more information.")
-            ),
-            'fixImportScmVariant' : (
-                "0.22.1.dev34",
-                InfoOnce("fixImportScmVariant policy not set. Recipe variant calculation w/ import SCM is boguous.",
-                    help="See http://bob-build-tool.readthedocs.io/en/latest/manual/policies.html#fiximportscmvariant for more information.")
-            ),
-            "defaultFileMode": (
-                "0.24rc1",
-                InfoOnce("defaultFileMode policy not set. File mode of URL SCMs not set for locally copied files.",
-                    help="See http://bob-build-tool.readthedocs.io/en/latest/manual/policies.html#defaultfilemode for more information.")
-            ),
-        }
+        self.__policies = RecipeSet.POLICIES.copy()
         self.__buildHooks = {}
         self.__sandboxOpts = {}
         self.__scmDefaults = {}
@@ -3383,7 +3430,8 @@ class RecipeSet:
         else:
             return schema[0].validate(default)
 
-    def parse(self, envOverrides={}, platform=getPlatformString(), recipesRoot=""):
+    def parse(self, envOverrides={}, platform=getPlatformString(), recipesRoot="",
+              noLayers=False):
         if not recipesRoot and os.path.isfile(".bob-project"):
             try:
                 with open(".bob-project") as f:
@@ -3393,17 +3441,19 @@ class RecipeSet:
         recipesDir = os.path.join(recipesRoot, "recipes")
         if not os.path.isdir(recipesDir):
             raise ParseError("No recipes directory found in " + recipesDir)
+        self.__projectRoot = recipesRoot or os.getcwd()
         self.__cache.open()
         try:
-            self.__parse(envOverrides, platform, recipesRoot)
+            self.__parse(envOverrides, platform, recipesRoot, noLayers)
         finally:
             self.__cache.close()
-        self.__projectRoot = recipesRoot or os.getcwd()
 
-    def __parse(self, envOverrides, platform, recipesRoot=""):
+    def __parse(self, envOverrides, platform, recipesRoot="", noLayers=False):
         if platform not in ('cygwin', 'darwin', 'linux', 'msys', 'win32'):
             raise ParseError("Invalid platform: " + platform)
         self.__platform = platform
+        self.__layers = []
+        self.__policies = RecipeSet.POLICIES.copy()
         self.__whiteList = set()
         if platform == 'win32':
             self.__whiteList |= set(["ALLUSERSPROFILE", "APPDATA",
@@ -3432,7 +3482,10 @@ class RecipeSet:
                 os.path.join(os.path.expanduser("~"), '.config')), 'bob', 'default.yaml'))
 
         # Begin with root layer
-        self.__parseLayer("", "9999", recipesRoot)
+        self.__parseLayer(LayerSpec(""), "9999", recipesRoot, noLayers)
+
+        if noLayers:
+            return
 
         # Out-of-tree builds may have a dedicated default.yaml
         if recipesRoot:
@@ -3475,14 +3528,17 @@ class RecipeSet:
         self.__rootRecipe = Recipe.createVirtualRoot(self, sorted(filteredRoots), self.__properties)
         self.__addRecipe(self.__rootRecipe)
 
-    def __parseLayer(self, layer, maxVer, recipesRoot):
+    def __parseLayer(self, layerSpec, maxVer, recipesRoot, noLayers=False):
+        layer = layerSpec.getName()
+
         if layer in self.__layers:
             return
         self.__layers.append(layer)
 
         rootDir = os.path.join(recipesRoot, os.path.join("layers", layer) if layer != "" else "")
         if not os.path.isdir(rootDir or "."):
-            raise ParseError(f"Layer '{layer}' does not exist!")
+            raise ParseError(f"Layer '{layer}' does not exist!",
+                                     help="You probably want to run 'bob layers update' to fetch missing layers.")
 
         configYaml = os.path.join(rootDir, "config.yaml")
         def preValidate(data):
@@ -3496,8 +3552,15 @@ class RecipeSet:
             if compareVersion(BOB_VERSION, minVer) < 0:
                 raise ParseError("Your Bob is too old. At least version "+minVer+" is required!")
 
-        config = self.loadYaml(configYaml, (RecipeSet.STATIC_CONFIG_SCHEMA, b''),
+        config_spec = {**RecipeSet.STATIC_CONFIG_SCHEMA_SPEC, **RecipeSet.STATIC_CONFIG_LAYER_SPEC}
+        config = self.loadYaml(configYaml, (schema.Schema(config_spec), b''),
             preValidate=preValidate)
+
+        # merge settings
+        for (name, value) in sorted([s for s in config.items() if s[0] in self.__settings],
+                                    key=lambda i: self.__settings[i[0]].priority):
+            self.__settings[name].merge(value)
+
         minVer = config.get("bobMinimumVersion", "0.16")
         if compareVersion(maxVer, minVer) < 0:
             raise ParseError("Layer '{}' requires a higher Bob version than root project!"
@@ -3505,6 +3568,7 @@ class RecipeSet:
         if compareVersion(minVer, "0.16") < 0:
             raise ParseError("Projects before bobMinimumVersion 0.16 are not supported!")
         maxVer = minVer # sub-layers must not have a higher bobMinimumVersion
+
 
         # Determine policies. The root layer determines the default settings
         # implicitly by bobMinimumVersion or explicitly via 'policies'. All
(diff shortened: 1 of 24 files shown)


jkloetzke:
Taking a step back: what is the actual motivation of having Bob managing the layers? What is the benefit of adding this feature compared to git submodules? I think we should have a compelling benefit to add the complexity. Could you shed some more light on this?

If we go this route, the following things come to my mind:

* Collapsing the layers hierarchy is probably OK. I don't think that anybody has ever used nested layers.
* Internally treating the layers with the same logic will most probably not fly.
  * Nested layers require some loop to discover and parse the next level. The current logic can handle only exactly one level.
  * Error messages and other output will probably be misleading.
* I could imagine that one may want to *optionally* automatically update layers unless building with `--build-only`.

rhubert:
> Taking a step back: what is the actual motivation of having Bob managing the layers? What is the benefit of adding this feature compared to git submodules? I think we should have a compelling benefit to add the complexity. Could you shed some more light on this?

There are two reasons:

My current layer structure would look like this if it would be supported:

```
- recipes a
   - layer b
     - layer c
   - layer c
   - layer d 
     - layer b
       - layer c
     - layer c
```

`layer d` also contains recipes for a stand alone product and it would be nice to be able to build then separate to `recipes a`. As of today this is not possible without changing the `config.yaml`  of `d` and to add `c` and `d` to the layers.
The remaining dependencies are just invisible and one would need to know them when reusing one of the other layers.

IMO this can't be solved with the current layers except by splitting the recipes much more. Or do you have any suggestions how this could be solved?


The other reason is "because some developers don't like submodules." We have (at least) one project where the layer content has been copied into the recipes just to avoid the need of using submodules. (I'm not sure if using bob managed layers would help them...)


> 
> If we go this route, the following things come to my mind:
> 
>     * Collapsing the layers hierarchy is probably OK. I don't think that anybody has ever used nested layers.
> 

I guess we can also add a policy for this?

>     * Internally treating the layers with the same logic will most probably not fly.
>       
>       * Nested layers require some loop to discover and parse the next level. The current logic can handle only exactly one level.

Yes, the loop is missing.

>       * Error messages and other output will probably be misleading.
> 
>     * I could imagine that one may want to _optionally_ automatically update layers unless building with `--build-only`.

I also thought about this as it would avoid the need of adding a new command. Not sure what would be more convenient.

jkloetzke:
> > Taking a step back: what is the actual motivation of having Bob managing the layers? What is the benefit of adding this feature compared to git submodules? I think we should have a compelling benefit to add the complexity. Could you shed some more light on this?
> 
> There are two reasons:
> ...
> `layer d` also contains recipes for a stand alone product and it would be nice to be able to build then separate to `recipes a`. As of today this is not possible without changing the `config.yaml` of `d` and to add `c` and `d` to the layers. The remaining dependencies are just invisible and one would need to know them when reusing one of the other layers.
> 
> IMO this can't be solved with the current layers except by splitting the recipes much more. Or do you have any suggestions how this could be solved?

Right. This makes sense and I also cannot see how it could be solved with git submodules either. So let's do it.

> The other reason is "because some developers don't like submodules." We have (at least) one project where the layer content has been copied into the recipes just to avoid the need of using submodules. (I'm not sure if using bob managed layers would help them...)

I'm not sure either. But having an "automatic" layers update that behaves like the SCMs in regular recipes might be more attractive.

> > If we go this route, the following things come to my mind:
> > ```
> > * Collapsing the layers hierarchy is probably OK. I don't think that anybody has ever used nested layers.
> > ```
> 
> I guess we can also add a policy for this?

No, I wouldn't unless somebody speaks up. I have never heard of anybody using nested layers. We can still add such a policy if somebody comes around.

> > * I could imagine that one may want to _optionally_ automatically update layers unless building with `--build-only`.
> 
> I also thought about this as it would avoid the need of adding a new command. Not sure what would be more convenient.

I think we need both. The layers command makes perfectly sense. I was thinking about the case where somebody clones a project and expects it to "just work". So

```
git clone ...
cd project
bob dev ...
```

should just work, even when layers are used. On the other hand, you might not always want the automatic layer pulling. So some overridable option to bob build/dev makes sense.

sbixl:
We have similar use cases as reported by @rhubert. I often get the question why the layers can't be managed better via bob (in analogy to the SCMs in the recipes). 

The current situation is that when a layer is updated, the whole team must be notified about this to update their local gitsubmodule(s). Sometimes it is forgotten and the colleague work on an older version. It would be nice if this update could be done via `bob dev/build` in the background. 

However, if the main recipes have a breaking dependency on one or more layers, these would also have to be updated. In this use-case, updating the layer alone would lead to problems and you would have to explicitly make a pull on the recipes (if you know it, which is mostly not the case).

The use of nested layers also makes sense (and we are now at the point where we absolutely need it). I have already experimented with nested layers but encountered similar problems to those reported by @rhubert.

As mentioned:

- Having an "automatic" layers update that behaves like the SCMs in regular recipes might be more attractive, would be a great improvement.
- Executing `git clone ...` without the knowledge of a gitsubmodule (`--recurse-submodules`) behind would improve the use-ability much more.
- Updating the main recipes in addition to the layers would be nice too. This could be done via a query that has to be confirmed by the user. This way the user knows that something has changed and an update is being carried out. This query could also be made configurable and set to `auto-update` so that the system does not wait for input e.g. in case of a CI build.

rhubert:
> The current situation is that when a layer is updated, the whole team must be notified about this to update their local gitsubmodule(s). Sometimes it is forgotten and the colleague work on an older version. It would be nice if this update could be done via `bob dev/build` in the background.
> 
> However, if the main recipes have a breaking dependency on one or more layers, these would also have to be updated. In this use-case, updating the layer alone would lead to problems and you would have to explicitly make a pull on the recipes (if you know it, which is mostly not the case).

I think the last point would become valid if `floating`layers are used, which - with submodules - is not possible(?). And even if layers managed by bob this is probably not a good idea. Otherwise the layers are bound to a version of the recipes and can't become incompatible to the recipes? Or do I miss something?

sbixl:
I had the following scenario in mind (assuming the layers are no longer managed via gitsubmodules):

```yaml
layers:
  - foo:
     scm: git
     url: git@foo:/foo.git
     branch: master # <-- layer is floating
```

Changes are made to the layer that also affect the main recipes (e.g. incompatible changes). Developer 1 adjusts the main recipes accordingly and push the changes to the remote. Developer 2 or a whole team knows nothing about this and would automatically update the layer with the next `bob dev/build` because it is floating. The system then stops work building because nobody knows that the recipes also need to be updated. This quickly leads to frustration, especially among colleagues who are not so familiar with recipes and their interrelationships. So if `floating` layers are an option there should be an anditional check if the main recipes must be updated too. 

But if I think more about it for a while, it makes no sense to work with floating layers. If several layers are nested inside each other, this increases the level of complexity considerably, becomes confusing and is error prone. I think it would be better to always hard link the layers to a tag/commit like the gitsubmodules already do it. 

So I agree that this was not the best idea. ;-)

Nevertheless, it would be a nice feature when the main recipes are on a branch and there are new changes on it you can you can get a notification from bob about this. In our team, at least, this is a recurring use case where you always have to make sure that everyone is really synchronized. I think this can be solved more elegant (it should be relatively easy for Bob to provide something like this?)

rhubert:
> Nevertheless, it would be a nice feature when the main recipes are on a branch and there are new changes on it you can you can get a notification from bob about this. In our team, at least, this is a recurring use case where you always have to make sure that everyone is really synchronized. I think this can be solved more elegant (it should be relatively easy for Bob to provide something like this?)

I think this can't be easily done as bob has no knowledge about where the recipes are coming from. Could be `git`, `svn`, `cvs`, ... so it might be hard to fetch them.

As of today you can add a `preBuildHook` to check if your recipes are up-to-date, otherwise you can output a warning and exit with a non zero exit code.

https://bob-build-tool.readthedocs.io/en/latest/manual/configuration.html#hooks

Note: you can not simply `git pull` your recipes in the hook and exit with 0 as the hook is executed after the recipes parsing has been done and the build would still use the old recipes.

rhubert:
I pushed some updates to the code to avoid the internal package.. It's still not ready for review...

Anyway - after some internal discussions we think that merging the (sub-) layers of different layers by their name/url/whatever might not work very well. Instead we want to filter the layers provided by a sublayer when depending on it:

Given a layer foo with
```yaml
# foo - config.yaml
layers:
 - bar:
   [...]
 - baz:
   [...]
``` 

The user of this layer can avoid the use of `bar` from `foo` by adding only `baz` to foo's `useSubLayers` , e.g. because `foo` also has a direct dependency on `bar`

```yaml
layers:
 - foo:
    checkoutSCM:
      scm: git
      url: ....
      commit: ...
    useSubLayers: [baz]
 - bar:
   [...]
```

Not adding `useSubLayers` will simply use all.

jkloetzke:
> Anyway - after some internal discussions we think that merging the (sub-) layers of different layers by their name/url/whatever might not work very well.

I agree that merging them on the basis of the URL is certainly not viable. But why not by name? Projects/layers that include other layers should know the naming below. Only if two layers refer to some common third layer by a different name things go south. But is this really likely to happen?

One could imagine to encourage people to use reverse polish notation for the layers (e.g. `dev.bobbuildtool.basement`). That would make the names more predictable.

> Instead we want to filter the layers provided by a sublayer when depending on it:

I'm not sure that this is better. The behaviour can get very subtle I guess. Also, how it it supposed to work transitively (control the sub-sub-layers)?

I don't have any better idea at the moment but if *feels* its more like hack...

rhubert:
The idea was to have more control about the layers. e.g. if a common sublayer is used by 2 different layers one might need to use the newer layer spec. And if there are two common sublayers by 2 different layers they might cross reference newer versions of their sublayers:

```
- recipes
 - layer foo
   - layer b (1.0)
   - layer c (1.1)
  - layer bar
   - layer b (1.1)
   - layer c (1.0)
```

To make this work we'd need to select `c` from `foo` and `b` from `bar`. Or add `b (1.1)` and `c (1.1)` to the recipes?

jkloetzke:
My current line of thinking goes like this:

We would always flatten the layer dependency tree by name. This happens even for layers that are not SCM-managed but just referenced in `config.yaml`. So whenever layers have the same name, only the first one of them is used. Which one it is, is defined by the order.

The project has the highest precedence. So if a layer "b" is defined in the project, it would override any deeper layer dependencies (layer `foo` and `bar` in your case). If the dependencies to some deeper layer are on the same level (`b` and `c` in your case), the order is important. It's already defined that layers in ``config.yaml`` are named from highest to lowest precedence. So in your case layer `b (1.0)` and layer `c (1.1)` would be used because layer `foo` has a higher precedence than `bar`. If you want `b (1.1)` and `c (1.1)`, they must be added to the project (at least `b (1.1)`).

While this can still be very subtle what is actually used, it at least is consistent with the current layer precedence. For SCM managed layers, Bob should certainly not even checkout layers that were already satisfied by higher precedence layers. So it will be more or less obvious what is going to be used.

Does that make sense?

rhubert:
I think this is now in a state where it might make sense to start reviewing process... 

One note to the last commit [XXXXXX: DevelopDirOracle: make a singleton + fix recalculation](https://github.com/BobBuildTool/bob/pull/561/commits/59043e221689915d2a15725f8271b7e0f88f830e). This tries to fix the failing `generator` blackbox test before I noticed that I missed

```python
   62     recipes.setUpdateLayers(not args.execute_buildonly)
```
in `pym/bob/cmds/build/project.py`. With this line it's not needed and I'm unsure if it's a good idea to drop and rebuild the cache if the key changes within a single bob-run? Maybe it would be better to raise a internal exception in that case?

jkloetzke:
Thanks. I'll have a look in the next days.

jkloetzke:
I'm not entirely sure if all of the following makes sense but these are my first thoughts after skimming though the stack for an hour or so.

* We should probably only create one `EventLoopWrapper()` and use it for both the layers update and the build instead of using two distinct instances.
* Instead of calling `setUpdateLayers()` and doing the update magically in `parse()`, it would be better to call an `updateLayers(eventLoop)` to do the update explicitly.
* Maybe we should exclude the `import` SCM? It doesn't look useful and we could spare some special cases.
* Using `setDirectoryState()`/`getDirectoryState()` is probably not a good idea because these directories are assumed to be workspaces by the `clean` and `status` commands. The layers state should be stored in a separate category.

The layer schema looks more complicated that it needs to be. How about the following?

```yaml
layers:
  - name: foo
    scm: git
    url: ...
  - name: bar
    ...
```

I'm not sure about the `DevelopDirOracle` change yet. My gut feeling is that it points to some underlying problem and making it a singleton just papers over it. But I have to admit I did not think this through.

I was hoping that we could integrate the layers update in the regular build TUI. Especially to use multiple jobs in parallel. But that may be something for a later step.

rhubert:
Thanks for the comments :) 
 
>     * We should probably only create one `EventLoopWrapper()` and use it for both the layers update and the build instead of using two distinct instances.

The layers update is also needed for non build commands like `bob ls` where we do not have a `EventLoopWrapper()` (yet?).  It could be created in the `RecipeSet` and provided to the build-commands using a getter?

>     * Instead of calling `setUpdateLayers()` and doing the update magically in `parse()`, it would be better to call an `updateLayers(eventLoop)` to do the update explicitly.

:+1: will do this.

>     * Maybe we should exclude the `import` SCM? It doesn't look useful and we could spare some special cases.

:+1:

>     * Using `setDirectoryState()`/`getDirectoryState()` is probably not a good idea because these directories are assumed to be workspaces by the `clean` and `status` commands. The layers state should be stored in a separate category.

Thanks - I wasn't aware of this ...

> The layer schema looks more complicated that it needs to be. How about the following?
> 
> ```yaml
> layers:
>   - name: foo
>     scm: git
>     url: ...
>   - name: bar
>     ...
> ```
I think this is a reasonable simplification.  We'd loos some flexibility but I can't think of a use-case for having multiple git-scms for a single layer or something like this..

> I was hoping that we could integrate the layers update in the regular build TUI. Especially to use multiple jobs in parallel. But that may be something for a later step.

The layers update is usually fast (compared to build + parse times). Not sure if there is a benefit of doing this in parallel, but I'll keep it in mind.

jkloetzke:
> The layers update is also needed for non build commands like `bob ls` where we do not have a `EventLoopWrapper()` (yet?). It could be created in the `RecipeSet` and provided to the build-commands using a getter?

I'm not sure that this is a good idea. Especially `bob ls` is not a state-changing command. I would be rather surprised when any layers would be updated in this case. IMHO there is only `bob build`, `dev` and the `layers` command that are expected to update the layers.

rhubert:
I did some rework and changed the code following your suggestions. Take your time for review I'm N/A the next weeks, so I there will be no more update from my side before ~~mid~~ end of July ;)

pym/bob/input.py
```diff
@@ -1737,6 +1737,21 @@ def validate(self, data):
...
+class LayerValidator:
+    def __init__(self):
+        self.__scmValidator = ScmValidator({
+            'git' : GitScm.SCHEMA,
+            'svn' : SvnScm.SCHEMA,
+            'cvs' : CvsScm.SCHEMA,
+            'url' : UrlScm.SCHEMA})
+
+    def validate(self, data):
+        if 'name' not in data:
```
jkloetzke, line 1749: How about adding the following here:

```python
if isinstance(data, str):
    return { 'scmSpec' : None, 'name' : data }
```

Then the layer could be treated always the same way, regardless of if it's a simple name or an SCM backed layer. Or even better, make a small `LayerSpec` class with two members (`name`, `scm`).

pym/bob/input.py
```diff
@@ -1737,6 +1737,21 @@ def validate(self, data):
...
+            raise schema.SchemaMissingKeyError("Missing 'name' key in {}".format(data), None)
+        name = data.get('name')
+        del data['name']
```
jkloetzke, line 1752: You should copy `data` before. Otherwise you're changing the parsed result which could lead to side effects.

pym/bob/input.py
```diff
@@ -1852,6 +1867,12 @@ def getProvideDepsResolver(pattern):
     else:
         return VerbatimProvideDepsResolver(pattern)
 
+def getLayerName(layerSpec):
+    if isinstance(layerSpec, str):
+        return layerSpec
+    else:
+        return layerSpec.get('name')
+
```
jkloetzke, line 1875: That should be unnecessary once the `LayerValidator` always creates `LayerSpec` objects.

pym/bob/input.py
```diff
@@ -2940,20 +2969,12 @@ class RecipeSet:
             },
             error="Invalid policy specified! Are you using an appropriate version of Bob?"
         ),
-        schema.Optional('layers') : [str],
+        schema.Optional('layers') : [schema.Or(str, LayerValidator())],
```
jkloetzke, line 2972: Could be `schema.Optional('layers') : [LayerValidator()]` then...

pym/bob/input.py
```diff
@@ -3447,15 +3478,15 @@ def __parse(self, envOverrides, platform, recipesRoot=""):
             self.__parseUserConfig(c)
 
         # calculate start environment
-        osEnv = Env(os.environ)
-        osEnv.setFuns(self.__stringFunctions)
```
jkloetzke, line 3450: That line should be retained. The layers can add something to `self.__stringFunctions` and that *could* be lost. I don't think it is (because we pass a dict and store it) but it is misleading, nonetheless.

pym/bob/input.py
```diff
@@ -3447,15 +3478,15 @@ def __parse(self, envOverrides, platform, recipesRoot=""):
...
         env = Env({ k : osEnv.substitute(v, k) for (k, v) in
             self.__defaultEnv.items() })
         env.setFuns(self.__stringFunctions)
         env.update(envOverrides)
-        env["BOB_HOST_PLATFORM"] = platform
```
jkloetzke, line 3455: That line must be retained.

pym/bob/input.py
```diff
@@ -3476,7 +3507,18 @@ def __parse(self, envOverrides, platform, recipesRoot=""):
...
-    def __parseLayer(self, layer, maxVer, recipesRoot):
+    def updateLayers(self, loop, defines, verbose):
+        from .layers import fetchLayers
+        try:
+            self.parse(defines, dryRun=True)
+        except ParseError:
+            pass
+        self.__layers = []
+        fetchLayers(self, self.__cache, loop, verbose)
+
```
jkloetzke, line 3518: That should be moved into the layers module. The input module should only deal with parsing the recipes.

pym/bob/layers.py
```diff
@@ -0,0 +1,228 @@
...
+    def status(self, verbose):
+        for level in self.__layers:
+            for layer in self.__layers[level]:
+                layer.status(verbose)
+
+def fetchLayers(recipes, yamlCache, loop, verbose):
+    layers = Layers(recipes, loop, yamlCache)
+    layers.collect(True, verbose)
+
+def doLayers(argv, bobRoot):
```
jkloetzke, line 200: The layers command handling should probably be moved into `pym/bob/cmds/layers.py` next to the other commands.

pym/bob/state.py
```diff
@@ -608,6 +625,12 @@ def resetWorkspaceState(self, path, dirState):
             else:
                 self.__dirStates[path] = dirState
             needSave = True
+        if self.__layerStates.get(path) != dirState:
+            if dirState is None:
+                del self.__layerStates[path]
+            else:
+                self.__layerStates[path] = dirState
+            needSave = True
```
jkloetzke, line 633: Is this necessary? I think `resetWorkspaceState` should never be called for a layers directory, shouldn't it?

jkloetzke (commented):
I think this is heading in the right direction. I still miss the loop to handle arbitrarily nested layers. Did I overlook it?

rhubert:
> I think this is heading in the right direction. I still miss the loop to handle arbitrarily nested layers. Did I overlook it?

`Layers::__collect` recursively calls itself...

jkloetzke:
I'm looking into the PR in detail but it will take some time. I'll tinker with it and probably push some fixups if that's ok...

jkloetzke:
I'm not really comfortable with the default.yaml handling during the layers update so far. It's also why the tests are failing. We try to parse the user configuration files but paper over any problem. This will certainly lead to surprising results. Maybe we should not even try that? Or do we need variable substitution in the layers SCMs?

rhubert:
I think having `scmOverrides` for the layer scm's is somehow important. As user I'd except they behave like normal `checkoutSCM`. And - in our build environment - we're using mirrors for all upstream repos. If we want to use `upstream-layer-foo` which has a configured to use `upstream-layer-b` with a checkout to `https://github...` we'd need to override this to point to our server. That's why I think parsing at least the `default.yaml` in the recipesRoot would be nice. Also to get the `scmDefaults` for the layers checkout,.... I pushed a fixup to do it like this.

If this won't work I think no support for Overrides+defaults in the layer scm's would be also acceptable. In this case one could still name `upstream-layer-b` in the recipes-root `config.yaml` ...

jkloetzke:
I see. In principle I fully support to have scmOverrides and environment substitution. The problem is that this creates us a circular dependency.

To fetch and/or update the layers we have to naturally go in top-to-bottom order. Even though it may not evident, the parsing of `default.yaml` of layers is done in bottom-up order. The rationale is so that upper layers can rely on layers beneath it and can also override settings from lower layers. For example, you could have a lower layer providing some string functions through a plugin and a higher layer uses that in it's `environment` section in `default.yaml`.

I'm not sure how we can solve this elegantly (or at all)...

rhubert:
Thanks for the explanation...sometimes it takes longer but now I got the point...I guess....

I removed the parsing of the `default.yaml` while updating the layers. Should be sufficient for now to have no scmOverrides  for the layers - let's see how long this holds...
If needed some light+static parsing of the scm-relevant settings of the root-`default.yaml` without support for (additional) string functions and all the stuff which could be added by lower layers might be a solution...

pym/bob/layers.py
```diff
@@ -0,0 +1,193 @@
...
+
+    def parse(self):
+        configYaml = os.path.join(self.__layerDir, "config.yaml")
+        config = self.loadYaml(configYaml, (RecipeSet.STATIC_CONFIG_SCHEMA, b''))
+        for l in config.get('layers', []):
+            scmSpec = l.getScm()
+            if scmSpec is None: continue
+            scmSpec.update({'recipe':configYaml})
+            layerScms = Scm(scmSpec,
+                            self.__recipes.getRootEnv(),
```
jkloetzke, line 128: Better pass an empty environment?

pym/bob/layers.py
```diff
@@ -0,0 +1,193 @@
...
+                                 recipeSet=self.__recipes)):
+                ret = await invoker.executeScmSwitch(self.__scm, oldState["prop"])
+                log("SWITCH: Layer '{}' .. ok".format(self.getName()), EXECUTED, INFO)
+
+                if ret == 0:
+                    BobState().setLayerState(layerSrcPath, newState)
+                    return
+        ret = os.path.exists(layerSrcPath)
+
+        if os.path.exists(layerSrcPath) and not created:
```
jkloetzke, line 84: This looks a bit dangerous. IIUC this might move away an existing directory that was placed there before the checkout. In `_cookCheckoutStep` we make sure that we only touch directories that we know of and don't obstruct any unknown directories. We should do the same here unless we have a good reason to deviate.

pym/bob/layers.py
```diff
@@ -0,0 +1,193 @@
...
+        self.__name = name
+        self.__root = root
+        self.__recipes = recipes
+        self.__yamlCache = yamlCache
+        self.__subLayers = []
+        self.__scm = scm
+        self.__layerDir = os.path.join(self.__root, "layers", self.__name) if len(self.__name) else self.__root
+
+    async def __checkoutTask(self, verbose):
+        if self.__scm is None:
```
jkloetzke, line 37: So we don't move obsolete layers into the attic? This is different compared to normal checkouts in recipes...

pym/bob/input.py
```diff
@@ -3431,8 +3464,19 @@ def __parse(self, envOverrides, platform, recipesRoot=""):
...
+        osEnv = Env(os.environ)
+        osEnv.setFuns(self.__stringFunctions)
+        env = Env({ k : osEnv.substitute(v, k) for (k, v) in
+            self.__defaultEnv.items() })
+        env.setFuns(self.__stringFunctions)
+        env.update(envOverrides)
+        env["BOB_HOST_PLATFORM"] = platform
+        self.__rootEnv = env
+
+        if noLayers:
```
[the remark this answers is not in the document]
rhubert: :thinking: you are right. OTOH this would only effect policies set in the recipes-root not in any sublayer since the layers are not parsed again after a checkout of a new layer...

pym/bob/layers.py
rhubert: I don't think this works. We need some OS-env (e.g `SSH_AGENT_*`) for the layer checkout. Use `Env(os.evironment)` might work?

pym/bob/layers.py
rhubert: I think this test is obsolete and can be removed. In `Layers::parse` a layer without a scm is already skipped.

Anyway - I think this is a edge case and I'm not sure how tho handle cases where a manager-layer is tured back into a external-layer. This not easily possible and would require multiple steps anyway. A simple `git pull --recures-submodules` would fail as the submodule-directoy already exists? So the user would need to update the recipes, run `bob layers update` to move the layer into attic and than update the submodules to make this work? I think in this case some communication to the team is needed in any case and than you can also advise them to move the existing layer manually?

pym/bob/input.py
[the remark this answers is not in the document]
jkloetzke: Luckily, this is not an issue. The policies are determined by the root layer. It's only checked that sub-layers do not contradict them.

pym/bob/layers.py
jkloetzke: The usual checkout logic has the following rules:

* A new checkout checks that there is not already a file or directory at the checkout location.
* Only checkouts are moved to the attic when their update fails.
* Obsolete checkouts are always moved to the attic.

If possible, I would like to have the same behaviour for managed layers.

If I understand your example correctly you have a project with submodules and now, after a git-pull the submodules is gone and the layer(s) is/are managed by Bob itself? I'm afraid this requires some communication. Git is not handling vanishing sub-layers well (and neither does Bob). Git will just abandon the submodule and the managed layers logic will bail out because the new checkout will collide with the abandoned submodule.

I think that letting the user sort that out is reasonable. Bob tries to checkout a layer and finds a directory there that is unknown. The only sensible default is to bail out. I guess this won't happen often.

pym/bob/layers.py
rhubert: `whitelist` (and `whitelistRemove`?) should be easy. 

But for the environment we'd need make restrictions with regards to the sting-functions use as plugins are not available during layers checkout,...
I'm also not sure if `environment` is really needed. The layer-Spec is not substituted. What could be a usecase?

pym/bob/layers.py
rhubert:
```quote
Obsolete checkouts are always moved to the attic.
```

Question is how `obsolete` is defined here. If I do no longer depend on a package the checkout for this is not moved to attic. The checkout for this is no longer run so the sources were kept in the workspace.

So if a layer is removed from the `config.yaml` would it become obsolete? As of today it's fine to have more layers in the `layers` directory than referenced by the `config.yaml`.

Or if the 'managed'-foo layer
```yaml
layers:
 - name: foo
   scm: git
   url: ....
```
becomes unmanaged:
```yaml
layers:
 - foo
```
is it now obsolete? It's still used... :shrug:

pym/bob/layers.py
jkloetzke: Sure, if `whitelist` is sufficient then we certainly don't add `environment` for no good reason. I was under the impression that this might be needed. Anyway, I'm fine with just `whitelist`.

We cannot add `whitelistRemove` because that can't be implemented in a top-down parsing fashion. This key exists so that higher layers (higher precedence) can remove `whitelist` entries of lower layers...

pym/bob/layers.py
jkloetzke: I would argue that, if a layer was checked out by Bob and the new version of `config.yaml` does not have a checkout, it should be moved to the attic. I see the project with the SCM layers like the source workspace of a recipe. When its updated, it should reflect the state of the recipe. In case of layers, its the state of `config.yaml` but the same logic should apply. Its the principle of consistency and least surprise.

Keeping the layer around might be confusing because the recipes are there but simply unused. Sure, it's the current behaviour for unmanaged layers but if Bob brings the layer into existence, it should probably also be moved away when being removed from the configuration. Additionally, what should happen if the next `config.yaml`, after the previous layer was abandoned, checks out a completely different layer `foo`? Maybe even by a different SCM? That must fail because there is already something at the checkout location.

So in your example, it would be moved to the attic. But that's interesting because I guess the `git pull` of the project will probably have failed. Then the user will take care of `foo` himself (delete/move away) and do the git pull again. Now, given the above logic, Bob will move the `foo` directory into the attic. This is clearly not what we want. So we need to handle the managed->unmanaged transition specially.

In summary, these may be the coarse rules:

1. New managed layers are not checked out into existing directories. Like normal SCMs, the directory must not exist on the initial checkout.
2. The transition of a layer from unmanaged to managed uses the same rules. That is, the user is responsible to remove the directory at the place of the newly managed layer first.
3. As always, we try to update in-place and if this fails move to the attic. I guess we want to give `bob layers` some switches to be able to tune the behaviour.
4. If a managed layer is removed, the layer is moved into the attic.
5. A layer that is transitioning from managed to unmanaged is abandoned. We don't want to move something into the attic that has been placed there by the user. This is the only deviation from the usual source workspace update rules.

pym/bob/layers.py
rhubert: I added the `whitlist` to `config.yaml`. Passing a empty `Env` breaks the unittest as substition of the `scm`  properties is failing...
I think it's possible to rewrite the test in order to avoid the need for passing additional variables, but maybe we can also pass the command line defines to the scm?

pym/bob/layers.py
rhubert: Thanks! I updated the code accordingly and added some tests for the different situations.

pym/bob/layers.py
rhubert: I did some testing on our CI. Passing variables + substitution on the scm-properties is necessary to get this to work. Locally the layer is checked out using a `git@gitserver.foo:repo.git` url. But on the gitlab-ci we need to use `https://gitlab-ci-token:$CI_JOB_TOKEN@gitserver.foo/repo.git` instead [1]. Since there are no `scmOverrides` for the layer-scms I think the only possible solution is to use
```yaml
 url: ${GITLAB_SERVER:-git@gitserver.foo:}/repo.git
```
and use `bob layers update -DGITLAB_SERVER="https://gitlab-ci-token:$CI_JOB_TOKEN@gitserver.foo/"` on the CI.

[1] https://docs.gitlab.com/ee/ci/jobs/ci_job_token.html#use-a-job-token

jkloetzke: This really bothers me. Your use case basically shows that something like `scmOverrides` are necessary. Approximating this with variables may be possible but this won't be a good solution when integrating other peoples layers. If we stick just to variable substitution, it basically means that you have to patch all layers so that the URL can be overridden in a way that works for you. But inventing special `scmOverrides` for layers does not feel right either... Feels like we're running in circles... :confused:

rhubert: Maybe we have to separate the layers from the `config/default.yaml` and put them into something like `layers_default`. This special config would be parsed top-down without using additional information from the other config-files. For the CI-Usecase it would be necessary to add more layers_config's via command-line or to support `include`s as well.

jkloetzke: I don't think we need another type of configuration file. We just have to bite the bullet and admit that some dedicated layers SCM override is necessary. I'm not sure how to handle the details, though. In any case, how about something like the following in `config.yaml`:

```yaml
layersScmOverrides:
  - match: ...
```

which applies solely to layer SCMs? Such setting applies for the current layer and all sub-layers.

We could also have `scmOverrides` in `config.yaml`. They would apply to layer SCMs and also regular SCMs in _all_ recipes. While possible, it might make things confusing because `scmOverrides` from `default.yaml` and other configuration file (`-c ...`) won't apply for layers, even though the would have the same name. That's why I think having a separate name for layer SCM overrides might be a better idea.

In the same vein, we might better add `layersWhitelist` instead of plain `whitelist` to `config.yaml`. It will only apply to layer SCMs and would prevent any confusion about the precedence of `default.yaml` `whitelist` entries.

What we still have to add is a way how to pass additional `config.yaml` style configuration file to Bob commands. Such files could only have `layersScmOverrides` and `layersWhitelist` keys, though. Anything else would not make sense IMHO.

jkloetzke:
Thanks. I think this still needs some polishing IMHO. But this can be done on top of it. The basic functionality is most probably ok.

PR #565, opened 2024-05-22
input: add 'inherit' dependency property

rhubert:
This property controls the inheritance of environment, tools and sandbox
for a dependency. It defaults to true so everything is inherited.

If set to `false` environment, tools and sandbox are dropped. The
dependency starts with the default-environment, no tools and no sandbox.

This becomes particularly useful when an existing root-package should
become a dependency of another root-package, e.g. for building an
installer.

### DIFF
--- a/doc/manual/configuration.rst
+++ b/doc/manual/configuration.rst
@@ -1291,6 +1291,16 @@ The following settings are supported:
 |             |                 | At the dependency both names will refer to the same |
 |             |                 | tool.                                               |
 +-------------+-----------------+-----------------------------------------------------+
+| inherit     | Boolean         | Inherit current environment, tools and sandbox to   |
+|             |                 | this dependency. When set to ``false``, all         |
+|             |                 | environment variables are reset to their default    |
+|             |                 | and no tools or sandbox are passed down to the      |
+|             |                 | dependency. This is mostly useful to make an        |
+|             |                 | existing root-package become a dependency of        |
+|             |                 | another (root) package.                             |
+|             |                 |                                                     |
+|             |                 | Default: ``true``                                   |
++-------------+-----------------+-----------------------------------------------------+
 
 .. _configuration-recipes-env:
 
--- a/pym/bob/input.py
+++ b/pym/bob/input.py
@@ -1855,10 +1855,11 @@ class Recipe(object):
     """
 
     class Dependency(object):
-        def __init__(self, recipe, env, fwd, use, cond, tools, checkoutDep):
+        def __init__(self, recipe, env, fwd, use, cond, tools, checkoutDep, inherit):
             self.recipe = recipe
             self.envOverride = env
             self.provideGlobal = fwd
+            self.inherit = inherit
             self.use = use
             self.useEnv = "environment" in self.use
             self.useTools = "tools" in self.use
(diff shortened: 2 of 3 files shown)


doc/manual/configuration.rst
```diff
@@ -1291,6 +1291,11 @@ The following settings are supported:
 |             |                 | At the dependency both names will refer to the same |
 |             |                 | tool.                                               |
 +-------------+-----------------+-----------------------------------------------------+
+| clean       | Boolean         | Drop all the environment and tools collected so far |
```
jkloetzke, line 1294: Could we invert the attribute to `inherit`? That would normally be `true`, and the user can set it to false. The idea is that this could actually also be a list some day, e.g. `[environment, tools, sandbox]`. I wouldn't implement this right now but when we name it `inherit` it's natural that we can enumerate in the future what exactly is inherited.

pym/bob/input.py
```diff
@@ -2317,24 +2319,29 @@ def prepare(self, inputEnv, sandboxEnabled, inputStates, inputSandbox=None,
...
-                    thisDepTools = depTools.derive({
-                        k : depTools[v] for k,v in dep.toolOverride.items() })
-                except KeyError as e:
-                    raise ParseError("Cannot remap unkown tool '{}' for dependency '{}'!"
-                        .format(e.args[0], recipe))
-                thisDepDiffTools = depDiffTools.copy()
-                thisDepDiffTools.update({
-                    k : depDiffTools.get(v, v)
-                    for k, v in dep.toolOverride.items() })
+            if dep.clean:
```
jkloetzke, line 2322: I would rather arrange it in a way where, even though `inherit` is set to `false`, some parts can still be inherited explicitly. Consider the following example:

```yaml
- name: some::root::recipe
  inherit: false
  environment:
    FOO: "$FOO"
  tools:
    host-toolchain: host-toolchain
```

This would only inherit the `FOO` environment variable and the `host-toolchain` tool. Right now the `environment` and `tools` keys would be completely ignored!

pym/bob/input.py
```diff
@@ -2317,24 +2319,29 @@ def prepare(self, inputEnv, sandboxEnabled, inputStates, inputSandbox=None,
...
+                thisDepEnv = self.getRecipeSet().getRootEnv()
+                thisDepTools = Env()
+                thisDepDiffTools = Env()
```
jkloetzke, line 2325: Hmm, what about the sandbox? Shouldn't that be cleared too?

doc/manual/configuration.rst
rhubert: :+1: `inherit` sounds good to me. We also had this list idea but we didn't find a usecase for cleaning only one of these.

pym/bob/input.py
rhubert: Thanks - I changed this and it seam to work now. not sure if I got this `thisDepDiffTools` thing right...

jkloetzke: I'm afraid it was not fully correct. I've took the freedom to just push an update. Please check if this still works for you...

rhubert:
Thanks for fixing this! I did some testing this morning and it works like expected.

jkloetzke:
Thanks.

PR #566, opened 2024-05-28
builder: consider all execution paths

jkloetzke:
The builds step implicitly depends on the execution path of all dependencies, even for tools.


rhubert:
The commit message looks somehow incomplete?

jkloetzke:
:facepalm: Thanks for noticing. Apparently I was a bit distracted yesterday evening. :laughing:

PR #570, opened 2024-06-18
show: fix crash with checkoutAssert

rhubert:
Fix a regression of 0c40313b:

     File "pym/bob/cmds/show.py", line 54, in <listcomp>
       { k:v for k,v in a.getProperties().items() if not k.startswith("__") }
                        ^^^^^^^^^^^^^^^
   AttributeError: 'dict' object has no attribute 'getProperties'

and add a test for this.


jkloetzke:
Thanks.

PR #576, opened 2024-07-12
Conditional tools

jkloetzke:
Add support for conditional tool usage...

### DIFF
--- a/doc/manual/configuration.rst
+++ b/doc/manual/configuration.rst
@@ -613,7 +613,7 @@ Other than the above differences setup scripts are identical to
 {checkout,build,package}Tools
 ~~~~~~~~~~~~~~~~~~~~~~~~~~~~~
 
-Type: List of strings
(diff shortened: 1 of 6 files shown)


doc/manual/configuration.rst
```diff
@@ -623,6 +623,23 @@ added to ``$LD_LIBRARY_PATH``.  The order of tools in ``$PATH`` and
 separate set of executables so that the order of their inclusion does not
 matter.
 
+In the simple form, a tool is only specified as simple string. This will use
+the tool unconditionally::
+
+    checkoutTools: [foo, bar]
+
+In necessary, a tool can also be used conditionally. In this case the tool
```
rhubert, line 631:
```suggestion
If necessary, a tool can also be used conditionally. In this case the tool
```

jkloetzke: Thanks for catching this.

PR #577, opened 2024-07-17
input: apply env substitution to metaEnvironment

### LINKED ISSUE #573: metaEnvironment does not seem to support var substitution
Something like this is not possible right now:
```
metaEnvironment:
       PKG_VERSION: "$LINUX_VERSION"
```
If we have generic recipes which e.g. get the version number to be used passed in, it would be nice if this is also reflected in the PKG_VERSION. I guess this would be the main place someone would like to look at for this kind of information.

rhubert:
Fixes #573.


jkloetzke:
Thanks for the fix!

PR #584, opened 2024-09-05
git: do not move detached HEAD to attic if same commit

rhubert:
If a git-workspace is in a detached HEAD state and the recipe has been updated to point to the same commit there is no need to move this into attic.


jkloetzke:
Thanks!

PR #585, opened 2024-09-17
build: add bundle build

rhubert:
A bob-bundle collects all scm's neccessary to build a package in a tar-file. Next to the tar-file a scmOverrides file is generated to be able to rebuild the package only with the bundle.

Such a bundle can be used to build on a air-gapped system or to archive all sources of a build.

ATM only bundling of git and url-scm's is supported.

### DIFF
--- a/contrib/bash-completion/bob
+++ b/contrib/bash-completion/bob
@@ -118,7 +118,7 @@ __bob_clean()
 
 __bob_cook()
 {
-   if [[ "$prev" = "--destination" ]] ; then
+   if [[ "$prev" = "--destination" || "$prev" == "--bundle" ]] ; then
       __bob_complete_dir "$cur"
    elif [[ "$prev" = "--download" ]] ; then
          __bob_complete_words "yes no deps forced forced-deps forced-fallback"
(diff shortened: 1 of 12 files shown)


rhubert:
TODO / RFC:

- [ ] black-box test
- [ ] add recipes to bundle?
- [ ] add some kind of unbundle command / script / ... ??

jkloetzke:
Why closed? I did not look into the details but it at the first glance it sounded like a desirable feature...

rhubert:
Just to not waste your time. :) 

I was caught up by reality shortly after pushing this. The initial idea to bundle only the `checkoutScm`s while still running the `checkoutScript`s doesn't work for all `checkoutScript` doing more fetches. (`cargo`, `alire`, `pip`, `meson subprojects download` ... ).  OTOH there are `checkoutScript`s that might be still needed on the `unbundled` machine (`autoreconf`, ...) so this is (as usual) much more complex...

I'm working on an update and will reopen this once I think it might hold a little longer...

jkloetzke:
Thanks. :)

I think the only reasonable option is to pack the whole source workspace after the checkout step is finished. Very much like the binary artifacts are treated now for package steps. Of course, this will only work for deterministic checkout steps...

PR #587, opened 2024-09-30
Polish layers

jkloetzke:
Some minor things that popped up while testing the layers SCM features....


jkloetzke:
@rhubert, could you have look?

rhubert:
LGTM :+1:

PR #593, opened 2024-10-22
layers: fix move to attic

rhubert:
When the layers where updated from bob build/dev move to attic was disabled by default.
As the unittest only tests the attic move from `bob layers update` this has been overlooked as attic is True in this case while it's None for build/dev. Adapt the default for the layers update command to match the default from build/dev.


jkloetzke:
Thanks.

PR #594, opened 2024-10-24
input: add substituteMetaEnv policy

rhubert:
Using bob 0.25 on older recipes with unfortunate metaEnvironment strings raises a parse error. Add a policy to issue a warning only.

### DIFF
--- a/doc/manual/configuration.rst
+++ b/doc/manual/configuration.rst
@@ -1567,6 +1567,10 @@ This predestines metaEnvironment variables to add the license type or version of
 
 The :ref:`manpage-query-meta` command can be used to retrieve metaEnvironment variables.
 
+All metaEnvironment variables are subject to :ref:`string substitution
+<configuration-principle-subst>`, unless the :ref:`policies-substituteMetaEnv`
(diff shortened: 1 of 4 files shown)


jkloetzke:
Damn, I was hoping that we don't have to roll a policy on this. But here we are. :roll_eyes: 

Actually, I wouldn't hide the error but apply the whole substitution only if the policy is on the new behaviour. Hiding the error just papers over the most obvious error. But suppose the substitution *does* work. Then older versions will show a different behaviour than now. So I think the policy should be named `substituteMetaEnv` and just skip the whole substitution if its on the old behaviour...

rhubert:
I was of the opinion that we're already using this in basement and I didn't want to break it there. But that doesn't seam to be the case...

In my particular case it fails on unclosed single quotes. Do you have another idea how to solve this without a policy?

jkloetzke:
> In my particular case it fails on unclosed single quotes. Do you have another idea how to solve this without a policy?

No, I don't think this can be solved without a policy. Actually, I should have added the policy in the first place.

jkloetzke:
Thanks

PR #604, opened 2024-12-29
fix unit tests on arch linux

rhubert:
A series of small fixes and adjustments to get the unit tests working on arch linux based distributions using the latest versions of git, different `hostname`, latest java, ...

### DIFF
--- a/pym/bob/scm/git.py
+++ b/pym/bob/scm/git.py
@@ -184,6 +184,12 @@ class GitScm(Scm):
             properties.update({GitScm.REMOTE_PREFIX+key : val})
         return properties
 
+    def _getGitConfigOptions(self):
+        config = [ "-c", "submodule.recurse=0" ]
+        if self.__url.startswith("file:") or self.__url.startswith("/"):
+            config += [ "-c", "protocol.file.allow=always" ]
+        return config
+
     async def invoke(self, invoker, switch=False):
         alternatesFile = invoker.joinPath(self.__dir, ".git/objects/info/alternates")
 
@@ -240,7 +246,7 @@ class GitScm(Scm):
         # refspec is kept in the git config.
 
         # Base fetch command with shallow support
-        fetchCmd = ["git", "-c", "submodule.recurse=0", "fetch", "-p"]
+        fetchCmd = ["git", *self._getGitConfigOptions(), "fetch", "-p"]
         if isinstance(self.__shallow, int):
             fetchCmd.append("--depth={}".format(self.__shallow))
         elif isinstance(self.__shallow, str):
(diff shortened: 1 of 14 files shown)


src/namespace-sandbox/namespace-sandbox.c
```diff
@@ -586,7 +586,7 @@ static void SetupDirectories(struct Options *opt, uid_t uid) {
     } while (entry != NULL && entry->pw_uid != uid);
     fclose(passwd);
 
-    if (entry == NULL) {
+    if ((entry == NULL) || strcmp(entry->pw_dir, "/") == 0) {
```
jkloetzke, line 589: Why do we have to fall back to `$HOME` exactly? Obviously, `/` is a valid directory. I think we should rather fix whatever prevents using `/`...

pym/bob/scm/git.py
```diff
@@ -174,12 +174,22 @@ def getProperties(self, isJenkins, pretty=False):
...
+        return config
+
     async def invoke(self, invoker, switch=False):
         alternatesFile = invoker.joinPath(self.__dir, ".git/objects/info/alternates")
 
         # make sure the git directory exists
         if not os.path.isdir(invoker.joinPath(self.__dir, ".git")):
             await invoker.checkCommand(["git", "init", self.__dir])
+            if self.__url.startswith("file:"):
+                await invoker.checkCommand(["git", "config", "--local",
```
jkloetzke, line 190: This looks odd. First, isn't the `self.__url.startswith("/")` test missing? Then, why is `_getGitConfigOptions` needed if the option is already set for the repository?

src/namespace-sandbox/namespace-sandbox.c
rhubert: Thanks. I fixed the creation of `/`.

pym/bob/scm/git.py
rhubert: Oh...This was the first attempt to get the tests back working on my system. Sorry - I overlooked it during cleanup. :see_no_evil:

jkloetzke: np

test/black-box/buildid-change/run.sh
```diff
@@ -15,6 +15,7 @@ popd
 D="$(mktemp -d)"
 pushd "$D"
 git init .
+git checkout -b master
```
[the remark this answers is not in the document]
rhubert: That works and I adapted the change accordingly.
Just a small note: the initial-branch option is available since git 2.28. Ubuntu 20.04 seams to use git 2.25, so it won't be available there. Same for debian oldoldstable..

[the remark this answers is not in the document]
jkloetzke: Thanks. I guess this is OK as it only affects executing the unit tests.

jkloetzke:
Thanks!

PR #606, opened 2025-01-01
Place url-scm download files in .bob-download

rhubert:
Separate the compressed input files from the extracted files in the workspace in preparation of bundling the sources.


jkloetzke:
Uhh, changing the `stable-variant-ids` test is a hinting that it breaks backwards compatibility. Moving the downloaded file into a subdirectory could break existing recipes. Even if the file is extracted, the recipe can still rely on the file being present today. So at least we must have a policy to guard this.

I would propose to have the downloaded file not in a magic hidden directory but instead completely outside of `workspace`. We could have a `downloads` next to `workspace` that stores the downloaded file and the canary. That would make it completely transparent. No need to exclude the directory from hashing and the workspace appears clean...

rhubert:
> I would propose to have the downloaded file not in a magic hidden directory but instead completely outside of `workspace`.

I also thought about this. The main reason for placing it inside was the attic logic for workspaces. I'm unsure how to thread the download-folder files if the recipe changes. Simply delete it? Or do we need some attic logic there as well?

jkloetzke:
> > I would propose to have the downloaded file not in a magic hidden directory but instead completely outside of `workspace`.
> 
> I also thought about this. The main reason for placing it inside was the attic logic for workspaces. I'm unsure how to thread the download-folder files if the recipe changes.

Good point. Indeed, this will require some extra logic to keep the `downloads` directory from collecting garbage. But this will probably enable further optimizations in the future. If the downloaded file is already available locally (e.g. when using a `preMirror`), we can directly extract such archives in the future without even copying the file in the workspace.

> Simply delete it? Or do we need some attic logic there as well?

I think we can simply delete it. The "downloads" directory probably needs to have the same structure as the "workspace". This is required anyway because there might be multiple url SCMs in the same workspace. I guess adding a `postAttic` method to SCMs which is called in the attic code path can simply delete the downloaded file.

pym/bob/input.py
```diff
@@ -3083,6 +3084,11 @@ class RecipeSet:
             InfoOnce("managedLayers policy is not set. Only unmanaged layers are supported.",
                 help="See http://bob-build-tool.readthedocs.io/en/latest/manual/policies.html#managedlayers for more information.")
         ),
+        "urlScmSeparateDownload": (
+            "0.25.1.dev27",
+            InfoOnce("urlScmSeparateDownload policy is not set. Urlscm downloads resist in the workspace.",
```
jkloetzke, line 3089: As this only affects downloaded files that are extracted, maybe the following would be more appropriate?

```
"urlScmSeparateDownload policy is not set. Extracted archives of the 'url' SCM are retained in the workspace."
```

doc/manual/policies.rst
```diff
@@ -344,6 +344,22 @@ New behavior
 
     Unmanaged layers are expected in the same directory.
 
+urlScmSeparateDownload
+~~~~~~~~~~~~~~~~~~~~~~
+
+Introduced in: 0.25.1
```
jkloetzke, line 350: The next version will be `1.0`.

pym/bob/scm/url.py
```diff
@@ -596,6 +628,9 @@ async def _put(self, invoker, workspaceFile, source, url):
             invoker.fail("Upload not supported for URL scheme: " + url.scheme)
 
     def canSwitch(self, oldScm):
+        if self.__separateDownload and not self.__version.get("valid"):
```
jkloetzke, line 631: I think the proper comparison is

```python
if self.__separateDownload != oldScm.__separateDownload:
    return False
```

Basically, this should make whole `VERSION` mechanism obsolete, does it?

test/black-box/extractors/recipes/extract_test.yaml
```diff
@@ -25,6 +25,9 @@ checkoutSCM:
     dir: zip
     digestSHA256: "8e2612b6ab1bacb5c0b342a81925ba6a0349ca8f66db4fc3dd22ddd53ff4f430"
 
+depends:
+  - nodir
```
jkloetzke, line 29: I think this `notdir` case can already be squashed into the first commit.

pym/bob/scm/url.py
```diff
@@ -788,6 +788,11 @@ def __getExtractor(self):
             raise ParseError("Invalid extract mode: " + self.__extract)
         return extractor
 
+    def postAttic(self, workspace):
+        if self.__separateDownload:
+            downloadDestination = os.path.abspath(os.path.join(workspace, "..", "download", self.__dir))
```
jkloetzke, line 793: Is the `os.path.abspath` really necessary? If true, then I would appreciate a comment why. It's somewhat unexpected.

jkloetzke (commented):
:+1: Just some minor nits...

pym/bob/scm/url.py
rhubert: the reason for `abspath` was, that `os.path.exists` has some trouble if the path contains `..`:

```
>>> import os
>>> os.path.exists('foo/bar/../..')
False
``` 

however I think using `os.path.normpath` instead of `abspath` is the cleaner solution?

jkloetzke: I think the trouble comes from the fact that it fails if some of the path elements does not exist. Like for example if the user removed `workspace` from `dev/[...]/1/workspace/../download` then the path really does not exist. So I think you're right, `os.path.normpath` is more appropriate. Still, I would appreciate a comment because it's not obvious. :wink:

rhubert: This sounds like a good explanation :) I looked into the documentation but couldn't find anything for `os.path.exists`. But I found a [note](  https://docs.python.org/3/library/os.html#os.makedirs) for `os.makedirs` saying it's also confused by `..`. So I added `normalize` there as well. And a comment ;)

jkloetzke: :+1:

jkloetzke:
Thanks!

PR #608, opened 2025-01-08
doc: Fix incorrect option "disassociate" in "checkoutSCM" git section

kenlau46:
Bob return failure when use `disassociate` in `checkoutSCM` section 
```
Parse error: Error while validating recipes/bob.yaml: Key 'checkoutSCM' error:
Wrong key 'disassociate' in {'scm': 'git', 'url': 'git@github.com:BobBuildTool/bob.git', 'branch': 'master', 'singleBranch': True, 'shallow': 1, 'disassociate': True}
```

Browse through the source code and `dissociate` is defined and used instead. 

Also fix typo and add man page section that explains the option


jkloetzke:
Good catch. The documentation obviously misspelled the option. Thanks!

PR #609, opened 2025-01-19
input: add support for package aliases

jkloetzke:
Virtual packages transparently create an alias for a recipe. They are configured by yaml files in the newly defined `virtual` directory. For example, the file `virtual/libs/libegl.yaml`

```
multiPackage:
    "":  ${CONFIG_SELECT_LIBEGL:-libs::mesa3d}
    dev: ${CONFIG_SELECT_LIBEGL:-libs::mesa3d}-dev
    tgt: ${CONFIG_SELECT_LIBEGL:-libs::mesa3d}-tgt
```

would create packages `libs::libegl`, `libs::libegl-dev` and `libs::libegl-tgt` whose underlying recipe is selected by `CONFIG_SELECT_LIBEGL`.

If there is only one package, the yaml file can be a simple string containing just the target name.

### DIFF
--- a/doc/manual/configuration.rst
+++ b/doc/manual/configuration.rst
@@ -19,6 +19,12 @@ be included from recipes and other classes to factor out common stuff. Files
 that do not have the '.yaml' extension are ignored when parsing the recipes and
 classes directories.
 
+In case there needs to be a choice between multiple recipes that provide
+interchangeable packages, aliases can be defined that resolve to other recipes.
+The optional ``aliases`` directory holds YAML files that declare alias names.
+The naming structure is the same as in the ``recipes`` and ``classes``
+directory. This is usually known as virtual packages in other build systems.
+
 There are two additional configuration files: ``config.yaml`` and
 ``default.yaml``. The former contains static configuration options while the
 latter holds some options that can have a default value and might be overridden
@@ -28,6 +34,8 @@ the following:
 .. code-block:: none
 
     .
+    ├── aliases
+    │   └── kernel.yaml
     ├── classes
     │   └── make.yaml
     ├── config.yaml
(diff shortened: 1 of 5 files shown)


jkloetzke:
@sixtyfourktec @rhubert It would be nice to get you feedback here. Given the "virtual" packages in basement-gnu-linux, I figured some native support should be provided. Especially since it's not possible to virtualize tools at all at the moment.

This PR is the bare minimum feature set that I could think of. Is this enough from your POV or do we need something more for proper virtual packages..?

rhubert:
:+1:  Look good + sufficient to me.

Only a quick idea: What about virtual classes? I'm thinking of the the integration `clang`, where something like a `virtual::clang` class could solve the issue of using clang or not without touching the recipes...maybe... not thought to the end...

jkloetzke:
Thanks for looking into it. I'm currently leaning to name this `alias` instead of `virtual`. Because it's creating an alias for an existing package. There is also #314 which requires the same underlying mechanism to create multiple dependencies to the same recipe by giving them distinct names:

```
depends:
    - name: bar
      alias: bar-one
      environment:
        CONFIG_BAR: "one"
    - name: bar
      alias: bar-two
      environment:
        CONFIG_BAR: "two"
```

As for classes I'm not sure if this will work. The aliases are created after all classes have been resolved already. But I have to think about this too...

jkloetzke:
Ok, so I renamed everything consistently to "alias". There are now global aliases that are configured in the `aliases` directory as YAML files. And then there are local aliases where each dependency in the `depends` list can get a different (but still locally unique) name.

@daxcore you might want to try out the local aliases. Any feedback if this fits your needs is highly appreciated. :)

More tests and the documentation are still missing, though...

PR #614, opened 2025-02-05
sandbox: mount dev/shm

sixtyfourktec:
Some tools use this while building. E.g. the python multiprocessing library uses this for its locking mechanism.


jkloetzke:
Thanks

PR #615, opened 2025-02-10
url: fix canary name of extracted archives

jkloetzke:
With the introduction of separate download directories in 2dffbccc, the extraction canary had changed its name. This change affects the source hash in case the urlScmSeparateDownload is kept on the old behaviour, though. To keep backwards compatibility, name the canary exactly like before.


jkloetzke:
@rhubert, this fixes a regression, leading to many "Restart build due to wrongly predicted sources." problems. It will also affect uploaded artifacts in case the urlScmSeparateDownload policy is on the old behaviour!

rhubert:
:+1:  Thank's for the hint. We're not yet using this productive -> fine for me.

PR #616, opened 2025-02-12
doc: add multiple inheritance example to 'inherit'

icedieler:
This example makes it easier for non-native Python developers to understand how to inherit from multiple classes.


jkloetzke:
Thanks!

PR #617, opened 2025-02-21
Make bob archive work remotely

MarcKe:
relates to issue https://github.com/BobBuildTool/bob/issues/340

### DIFF
--- a/pym/bob/archive.py
+++ b/pym/bob/archive.py
@@ -365,6 +365,7 @@ class BaseArchive(TarHelper):
         self.__useLocal = "nolocal" not in flags
         self.__useJenkins = "nojenkins" not in flags
         self.__useCache = "cache" in flags
+        self.__managed = "managed" in flags
         self.__wantDownloadLocal = False
         self.__wantDownloadJenkins = False
         self.__wantUploadLocal = False
@@ -401,6 +402,9 @@ class BaseArchive(TarHelper):
         raise ArtifactNotFoundError()
 
     def canManage(self):
+        return self.__managed and self._canManage()
+
+    def _canManage(self):
         return False
 
     async def downloadPackage(self, step, buildId, audit, content, caches=[],
@@ -754,7 +758,7 @@ class LocalArchive(BaseArchive):
         self.__fileMode = spec.get("fileMode")
         self.__dirMode = spec.get("directoryMode")
 
-    def canManage(self):
+    def _canManage(self):
         return True
 
     def _getPath(self, buildId, suffix):
@@ -879,7 +883,7 @@ class HttpArchive(BaseArchive):
                 if not retry: return (False, e)
                 retry = False
 
-    def canManage(self):
+    def _canManage(self):
         return True
 
     def _resetConnection(self):
--- a/pym/bob/archive.py
+++ b/pym/bob/archive.py
@@ -63,6 +63,8 @@ def writeFileOrHandle(name, fileobj, content):
     with open(name, "wb") as f:
         f.write(content)
 
+def namedErrorString(name, err):
+    return "Archive '{}': {}".format(name, err)
 
 class DummyArchive:
     """Archive that does nothing"""
@@ -109,6 +111,9 @@ class DummyArchive:
 
     async def downloadLocalFingerprint(self, step, key, executor=None):
         return None
+
+    def getArchiveName(self):
+       return "Dummy"
 
 class ArtifactNotFoundError(Exception):
     pass
@@ -195,6 +200,7 @@ class JenkinsArchive(TarHelper):
 
     def __init__(self, spec):
         self.__xferArtifacts = spec.get("xfer", False)
+        self.__name = spec.get("name", "unknown name")
 
     def wantDownloadLocal(self, enable):
         pass
@@ -228,7 +234,7 @@ class JenkinsArchive(TarHelper):
             with open(self.buildIdName(step), "wb") as f:
                 f.write(buildId)
         except OSError as e:
-            raise BuildError("Cannot store artifact: " + str(e))
+            raise BuildError(namedErrorString(self.__name, "Cannot store artifact: " + str(e)))
 
         if self.__xferArtifacts:
             loop = asyncio.get_event_loop()
@@ -243,7 +249,7 @@ class JenkinsArchive(TarHelper):
                             audit, content)
                         a.setResult(msg, kind)
                 except (concurrent.futures.CancelledError, concurrent.futures.process.BrokenProcessPool):
-                    raise BuildError("Packing of package interrupted.")
+                    raise BuildError(namedErrorString(self.__name, "Packing of package interrupted."))
 
     def _uploadPackage(self, name, buildId, audit, content):
         # Set default signal handler so that KeyboardInterrupt is raised.
@@ -252,7 +258,7 @@ class JenkinsArchive(TarHelper):
         try:
             self._pack(name, None, audit, content)
         except (tarfile.TarError, OSError) as e:
-            raise BuildError("Cannot pack artifact: " + str(e))
+            raise BuildError(namedErrorString(self.__name, "Cannot pack artifact: " + str(e)))
         finally:
             # Restore signals to default so that Ctrl+C kills process. Needed
             # to prevent ugly backtraces when user presses ctrl+c.
@@ -270,7 +276,7 @@ class JenkinsArchive(TarHelper):
                 if not ret: a.fail(msg, WARNING)
                 return ret
             except (concurrent.futures.CancelledError, concurrent.futures.process.BrokenProcessPool):
-                raise BuildError("Extraction of package interrupted.")
+                raise BuildError(namedErrorString(self.__name, "Extraction of package interrupted."))
 
     def _downloadPackage(self, tgzName, buildIdName, buildId, audit, content):
         # Set default signal handler so that KeyboardInterrupt is raised.
@@ -287,7 +293,7 @@ class JenkinsArchive(TarHelper):
                 self._extract(f, audit, content)
             return (True, None)
         except (OSError, tarfile.TarError) as e:
-            raise BuildError("Error extracting binary artifact: " + str(e))
+            raise BuildError(namedErrorString(self.__name, "Error extracting binary artifact: " + str(e)))
         finally:
             # Restore signals to default so that Ctrl+C kills process. Needed
             # to prevent ugly backtraces when user presses ctrl+c.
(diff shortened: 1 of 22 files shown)


MarcKe:
I want to get some first feedback from your side.

jkloetzke:
Sure. First of all, this feature is very much appreciated. It's been on the TODO list for a long time. :+1: 

Before reviewing the code in detail, a couple of things that I noticed:

* The archive command should still support working locally. That is, it should be possible to execute the command for a local directory. This is useful to run `bob archive` as a cron-job on the storage server.
* I guess it should be possible to work on selected archive backends if there is more than one. I would even argue that it's better to require an explicit selection of the user on the desired servers and/or local directories. By doing the operation on all backends, it might lead to undesired data loss because the user was not aware it is affecting all at the same time.
* That probably requires something like a `name` attribute for a backend. This will be handy also in the DOWNLOAD/UPLOAD messages. It was always confusion if multiple backends are configured, which one exactly had a hit or miss when building. If no name is given, you simply cannot run bob-archive on it.
* Should we add a `managed` flag? You might not have the rights to enumerate or delete artifacts even though you can upload.

Code wise, I think the circular dependency between the webdav and archive modules needs to be properly resolved. IMHO, the webdav module should throw HTTP errors. That will require some adaptation in the archive module but this hack to resolve the circular dependency is, well, just a hack...

MarcKe:
> Sure. First of all, this feature is very much appreciated. It's been on the TODO list for a long time. 👍
> 
> Before reviewing the code in detail, a couple of things that I noticed:
> 
>     * The archive command should still support working locally. That is, it should be possible to execute the command for a local directory. This is useful to run `bob archive` as a cron-job on the storage server.
> 

Yes, sounds reasonable, with an option like `-l/--local`? Would that be ok?

>     * I guess it should be possible to work on selected archive backends if there is more than one. I would even argue that it's better to require an explicit selection of the user on the desired servers and/or local directories. By doing the operation on all backends, it might lead to undesired data loss because the user was not aware it is affecting all at the same time.
> 

Sounds good to me. There could be an argument like `-a/--all` to just do it like I have implemented it now.

>     * That probably requires something like a `name` attribute for a backend. This will be handy also in the DOWNLOAD/UPLOAD messages. It was always confusion if multiple backends are configured, which one exactly had a hit or miss when building. If no name is given, you simply cannot run bob-archive on it.
> 

I will have a look at this together with the explicit selection.

>     * Should we add a `managed` flag? You might not have the rights to enumerate or delete artifacts even though you can upload.
> 

I think this is also a good idea.

> 
> Code wise, I think the circular dependency between the webdav and archive modules needs to be properly resolved. IMHO, the webdav module should throw HTTP errors. That will require some adaptation in the archive module but this hack to resolve the circular dependency is, well, just a hack...

Good idea.

jkloetzke:
> Yes, sounds reasonable, with an option like `-l/--local`? Would that be ok?

Yes, sounds good.

test/test-lib.sh
```diff
@@ -154,6 +154,15 @@ expect_fail()
 	return 1
 }
 
+expect_success()
```
jkloetzke, line 157: All tests run with `set -e`. So implicitly every command is prefixed by `expect_success`.

test/unit/test_archive.py
```diff
@@ -581,6 +581,12 @@ def do_MKCOL(self):
                 self.send_response(403)
             self.end_headers()
 
+        def do_OPTIONS(self):
```
jkloetzke, line 584: That could go to a preparatory commit. Otherwise, the test adaptations, should be in the same commit that changes the code. The goal is that is must be possible to `git bisect` and run the tests at any stage...

pym/bob/cmds/archive.py
```diff
@@ -484,40 +484,52 @@ def doArchive(argv, bobRoot):
...
-            archiver = getSingleArchiver(recipes, archivespec[0])
-            if archiver.archiveCmdSupported():
-                archivers.append(archiver)
-            else:
-                raise BobError("Archiver does not support the archive command")
+    if args.local:
+        # provide local directory as backend
+        archivespec = [{"backend": "file", "path": "{}".format(os.getcwd())}]
+        archiver = getSingleArchiver(None, archivespec[0])
+        if archiver.archiveCmdSupported():
```
jkloetzke, line 507: Can this ever be false? Maybe `assert` instead...

test/black-box/archive/run.sh
```diff
@@ -18,89 +18,120 @@ function clean_output () {
...
+  popd
+  test "$found1" != "$found2"
+
+  # Make sure invalid audit meta keys are rejected
+  expect_fail run_bob build -M "!nv@l1d=key" root-alpha
+}
+# run tests with configuration with file backend
+run_tests
+# run tests with local directory
+run_tests "local"
```
jkloetzke: Missing newline at end of file.

pym/bob/input.py
```diff
@@ -2982,6 +2982,7 @@ def __init__(self):
         self.__validTypes = schema.Schema({'backend': schema.Or('none', 'file', 'http', 'shell', 'azure')},
             ignore_extra_keys=True)
         baseArchive = {
+            'name' : str,
```
jkloetzke, line 2985: The `name` attribute must be optional. Otherwise we'll break a lot of existing projects. That will also save you the test adaptions in large part I guess.

pym/bob/cmds/archive.py
```diff
@@ -486,6 +486,12 @@ def doArchive(argv, bobRoot):
 """.format(subHelp))
     parser.add_argument("-l", "--local", action='store_true',
                         help="ignore the archive configuration and work on the current directory")
+    parser.add_argument("-a", "--all", action='store_true',
+                        help="execute the command on all suitable archives in the user configuration")
+    parser.add_argument("-b", "--backends", type=str, metavar="names",
```
jkloetzke, line 491: There is an `action='append'` mode of argparse. With that, the user could just give `-b` multiple times, one for each archive.

jkloetzke (commented):
This turned out to be quite a lot of comments. I certainly did not catch everything yet but I hope you bear with me...

MarcKe:
> This turned out to be quite a lot of comments. I certainly did not catch everything yet but I hope you bear with me...

No, thank you, this is great.

pym/bob/webdav.py
```diff
@@ -0,0 +1,276 @@
+from . import BOB_VERSION
+from .errors import BuildError
+from .utils import sslNoVerifyContext
+import base64
+import http.client
+import os
+from urllib.parse import unquote, urlsplit, urlparse
+from xml.etree.ElementTree import fromstring
+
+class HttpDownloadError(http.client.HTTPException):
```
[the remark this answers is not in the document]
jkloetzke: My original idea was to just let the `http.client.HTTPException` bubble up and return the other codes. But your approach is also fine. But in this case the exceptions do not need to derive from `http.client.HTTPException`.

pym/bob/archive.py
```diff
@@ -777,6 +950,41 @@ def _openUploadFile(self, buildId, suffix, overwrite):
...
+        self.webdav.delete(filename)
+
+    def _stat(self, filename):
+        stats = self.webdav.stat(filename)
+        if not stats['cdate'] or not stats['mdate'] or not stats['etag'] or not stats['len']:
+            print("error")
+        from datetime import datetime
+        from email.utils import parsedate_to_datetime
+        import binascii
+        return struct.pack('=ddLL', datetime.fromisoformat(stats['cdate']).timestamp(), parsedate_to_datetime(stats['mdate']).timestamp(), binascii.crc32(stats['etag'].encode('utf8')), stats['len'])
```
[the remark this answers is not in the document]
jkloetzke: :face_with_spiral_eyes: As always, every server does it differently. :facepalm:  I would suggest to use the "getetag" if there is one that is not weak. Otherwise, the combination of length and last modification date looks like the best approach.

pym/bob/cmds/archive.py
```diff
@@ -23,15 +22,17 @@
...
     CUR_VERSION = 2
 
-    def __init__(self):
+    def __init__(self, archiver):
         self.__dirSchema = re.compile(r'[0-9a-zA-Z]{2}')
         self.__archiveSchema = re.compile(r'[0-9a-zA-Z]{36,}-1.tgz')
         self.__db = None
         self.__cleanup = False
+        self.__archiver = archiver
+        self.__dbName = ".bob-archive_{}.sqlite3".format(self.__archiver.getArchiveHash())
```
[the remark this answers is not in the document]
jkloetzke: I would prefer to have only one database file. In the end, the database is just a cache so it's not an issue if we change the schema. This is already anticipated with the version check...

pym/bob/archive.py
```diff
@@ -559,6 +582,119 @@ async def downloadLocalFingerprint(self, step, key, executor=None):
             except (concurrent.futures.CancelledError, concurrent.futures.process.BrokenProcessPool):
                 raise BuildError("Download of fingerprint interrupted.")
 
+    def deleteFile(self, filepath):
+        # Set default signal handler so that KeyboardInterrupt is raised.
```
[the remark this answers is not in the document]
jkloetzke: You mean the signal stuff? No, that is not necessary here.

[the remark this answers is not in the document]
MarcKe: Sorry, that comment was in pending for a long time. I was not aware that I needed to press "submit".

MarcKe:
I think I have managed to go through all your remarks.

jkloetzke:
Thanks. I'll have a look on the weekend (hopefully).

pym/bob/cmds/archive.py
```diff
@@ -23,15 +22,19 @@
 class ArchiveScanner:
     CUR_VERSION = 2
```
jkloetzke, line 23: We must bump the version because the schema changed.

pym/bob/cmds/archive.py
```diff
@@ -500,6 +500,12 @@ def doArchive(argv, bobRoot):
 """.format(subHelp))
     parser.add_argument("-l", "--local", action='store_true',
                         help="ignore the archive configuration and work on the current directory")
+    parser.add_argument("-a", "--all", action='store_true',
```
jkloetzke, line 503: I think you can use a `add_mutually_exclusive_group` for `-l`, `-a` and `-b`. That way, only one of them can be given and we do not have to define a precedence.

jkloetzke (commented):
That looks quite good already. Would it be possible to add some testing for the Webdav parts? Maybe (re)use the HTTP mock server from the unit tests to have some minimally responsive Webdav server? Otherwise this is basically all untested and I fear it will break unnoticed whenever we touch the archive stuff...

MarcKe:
> That looks quite good already. Would it be possible to add some testing for the Webdav parts? Maybe (re)use the HTTP mock server from the unit tests to have some minimally responsive Webdav server? Otherwise this is basically all untested and I fear it will break unnoticed whenever we touch the archive stuff...

Are those tests enough?
To compare the branches I used this:
https://github.com/BobBuildTool/bob/compare/ea743015fb0edea4bfce231ba21578158339fa16...7d3c0fb8df4f98166ba0a6d1ba694f0d8364e0df
Is there a better way to do that on github?

jkloetzke:
Thanks a lot!

PR #622, opened 2025-03-14
input: custom properties must not start with lower case

jkloetzke:
Just like for custom settings in default.yaml, plugin defined recipe properties and state trackers shall not start with a lower case letter. This namespace is reserved by Bob for future extensions.


rhubert:
This has been announced/discussed in https://github.com/BobBuildTool/bob/discussions/620 (where you can also find the reason for the change). Thomas confirmed you are not using custom properties... :thinking: :shrug:

PR #627, opened 2025-03-23
Improve audit performance

jkloetzke:


pym/bob/audit.py
```diff
@@ -113,178 +117,125 @@ def __getOsRelease(cls):
...
+    def reset(self, variantId, buildId, resultHash, date=None):
         u = platform.uname()
-        self.__build = {
-            'sysname'  : u.system,
-            'nodename' : u.node,
-            'release'  : u.release,
-            'version'  : u.version,
-            'machine'  : u.machine,
-            'date'     : datetime.now(timezone.utc).isoformat(),
+        # Versions ob Bob before 0.25 required some fields to be always
```
rhubert, line 132:
```suggestion
        # Versions of Bob before 0.25 required some fields to be always
```

jkloetzke: Thanks for noticing. :)

PR #628, opened 2025-03-26
make HttpArchive retry configurable

MarcKe:


test/unit/test_archive.py
```diff
@@ -152,13 +170,7 @@ def __getSingleArchiveInstance(self, spec):
...
 
     def setUp(self):
-        # create repo
-        self.repo = TemporaryDirectory()
-
-        # add artifacts
-        self.dummyFileName = self.__createArtifact(DOWNLOAD_ARITFACT)
-        self.__createArtifact(WRONG_VERSION_ARTIFACT, "0")
-        self.__createBuildId(DOWNLOAD_ARITFACT)
+        super().setUp()
```
jkloetzke, line 178: I guess you can remove the method if you just call the base class method. Same for `tearDown` below...

pym/bob/archive.py
```diff
@@ -885,6 +885,13 @@ def __retry(self, request):
                 if retries == 0: return (False, e)
                 retries -= 1
 
+    def _retry(self, request):
+        (ok, result) = self.__retry(request)
+        if ok:
+            return result
+        else:
+            raise result
+
```
jkloetzke: Why not just modify `__retry` accordingly? AFAICS, all call sites rely on throwing in case of an error.

Also, please separate the cleanup (getting rid of all the identical if/else constructs) from the actual change. Those are two independent changes.

pym/bob/archive.py
```diff
@@ -944,14 +939,13 @@ def _openUploadFile(self, buildId, suffix, overwrite):
...
 
     def _putUploadFile(self, path, tmp, overwrite):
-        (ok, result) = self.__retry(lambda: self._webdav.upload(path, tmp, overwrite))
-        if ok:
-            return result
-        else:
-            raise result
+        return self._retry(lambda: self._webdav.upload(path, tmp, overwrite))
+
+    def __listdir(self, path):
```
jkloetzke, line 944: Why add a new method? There is only one caller, isn't it. Same for `__delete` and `__stat`...

test/unit/test_archive.py
MarcKe: There is more code following here. I agree with the tearDown()

pym/bob/archive.py
MarcKe: That's a good question. I don't remember my thoughts :>

MarcKe:
> Thanks. Just some minor comments...
> 
> There seems to be an issue with the test execution on Python 3.12. Maybe one of the new test hangs sporadically?
> 
> (BTW, improving on the commit message style would be nice. Just take an inspiration on the existing commits. Usually, the commit message should give a rationale _why_ the change is made. Basically a message to our future selves to understand why things had to be changed. 😉 )

Im confused by python 3.12. The tests that I touched both completed...

MarcKe:
> > Thanks. Just some minor comments...
> > There seems to be an issue with the test execution on Python 3.12. Maybe one of the new test hangs sporadically?
> > (BTW, improving on the commit message style would be nice. Just take an inspiration on the existing commits. Usually, the commit message should give a rationale _why_ the change is made. Basically a message to our future selves to understand why things had to be changed. 😉 )
> 
> Im confused by python 3.12. The tests that I touched both completed...

It looks like the blackbox test "setup-scripts" did not run and the coverage is missing. I do not know why. I tested it with pyenv and 3.12.9. The tests just worked.

jkloetzke:
Strange. Not sure what changed. Seems like I'll have to investigate this. Anyway, that should not prevent us from merging the PR.

Thanks.

PR #630, opened 2025-03-28
webdav: do not consider 404 an error for delete

MarcKe:
deleting a file that is not there (anymore) should not throw an exception. this caused some issues in our company's interal CỊ where cache issues lead to delete attempts on files that were already gone

### DIFF
--- a/pym/bob/webdav.py
+++ b/pym/bob/webdav.py
@@ -196,7 +196,7 @@ class WebDav:
         connection.request("DELETE", filepath, headers=headers)
         response = connection.getresponse()
         response.read()
-        if response.status not in [200, 204]:
+        if response.status not in [200, 204, 404]:
             raise HttpDownloadError("DELETE {} {}".format(response.status, response.reason))
 
     def stat(self, file):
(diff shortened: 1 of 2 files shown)


MarcKe:
@jkloetzke I think your 3.12 environment is somehow broken. This should just run as before...

MarcKe:
It is definitely hanging in the `scripts` call in `setup-scripts`. I could replicate it locally (had parallel missing)
I also tried different version. 3.11.11 and 13.13.2 work. 3.12.8 and 3.12.9 do not. Sometimes the tests run through, so there seems to be some interference between tests...
3.12 also consistently has the slowest execution time.
Here are some example run times:
3.13.2:
real	3m26,217s
user	8m3,704s
sys	1m1,103s

3.12.9 (canceled when other tests finished):
real	4m51,279s
user	11m38,881s
sys	0m59,628s

3.11.11:
real	2m39,814s
user	5m43,922s
sys	1m1,341s

jkloetzke:
I took myself the freedom to reformat the commit message.

The [failure in 3.10](https://github.com/BobBuildTool/bob/actions/runs/14134811953/job/39603855101?pr=630#step:7:555) seems to be a recent regression. It's not related to this PR, though.

I still have no clue what causes 3.12 to fail consistently. :shrug:

Issue #645, opened 2025-05-19
Make dependencies available in package setp


jkloetzke:
**Is your feature request related to a problem? Please describe.**
Currently, the dependencies of a recipe are only available in the build step. In case the recipe just gathers the result of other recipes, this is wasteful because the aggregated result needs to be copied once again in the package step.

**Describe the solution you'd like**
Add a `packageDep` property to the `depends` entries, mirroring the `checkoutDep` property.

PR #647, opened 2025-05-20
Package step dependencies

jkloetzke:
Implements #645.

### DIFF
--- a/doc/manual/configuration.rst
+++ b/doc/manual/configuration.rst
@@ -548,7 +548,7 @@ a script for the same step are joined in front of this script in the order the
 inheritance is specified. The inheritance graph is traversed depth first and
 every class is included exactly once.
 
-Dependencies of the recipe are by default only available to the
+Dependencies of the recipe are generally only available to the
 ``buildScript``. The path to the previous step (checkout workspace for
 ``buildScript``, build workspace for ``packageScript``) is always passed in
 ``$1``. Other dependencies are available in the order in which they were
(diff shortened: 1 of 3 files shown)


jkloetzke:
@icedieler @rhubert @sbixl Any thoughts on this?

Basically, you can now write:

```yaml
depends:
    - ...

packageDepends: True
packageScript: |
    for i in "{@:2}" ; do
        ...
    done
```

I figured that my original idea with a fine-grained `packageDep` property in the `depends` section does not really fly. The problem are indirect dependencies that would then have to be handled in the fine grained fashion too. Also, I don't see a use case where only selected dependencies should be made available to the package step. But you guys may have a different idea...

rhubert:
I think having all dependencies in the package step set is sufficient.  I assume `$1` is still the build step? Can we also make the checkout Step available? Is `BOB_DEP_PATHS` set?

jkloetzke:
> I assume `$1` is still the build step?

Indeed, this is the case. Just like in the build step, the dependencies start at `$2`.

> Can we also make the checkout Step available?

You can have that even today. Even in the sandbox, the checkout step is accessible for the package step. The reason is that we used `make install` in the package step in the past. So the following will work:

```yaml
checkoutSCM:
    ...

buildScript: |
    ln -sf "$1" src

packageScript: |
    rsync "$1/src/" .
```

> Is `BOB_DEP_PATHS` set?

Yes, just like for the build step.

rhubert:
I know this make-a-symlink-pointing-to-the-sources way. I just thought it might be easier to read / less to write if we could get rid of this and have a direct access to the checkoutStep.

```yaml
checkoutSCM:
  ...

packageScript: |
  rsync $CHECKOUT_DIR/ .
```

jkloetzke:
Hmm, this is unfortunately not easily possible. The problem is that `BOB_DEP_PATHS` will be ambiguous then. The package name of the checkout step and the build step will obviously be identical. While we could argue that this is just a minor nuisance, I made the observation that such inconsistencies will bite us in the end. Therefore, I would not like to go this route.

rhubert:
Fine for me - was just an idea.

sbixl:
The only use case I can think of right now is the one Ralf already mentioned. Fine for me too.

jkloetzke:
Thanks for the feedback. :+1:

PR #653, opened 2025-07-21
bob-[dev, build] manpage: improve command syntax descriptions

icedieler:
Unify the description of the option syntax. Add the long option where one is available.

### DIFF
--- a/doc/manpages/bob-build-dev.rst
+++ b/doc/manpages/bob-build-dev.rst
@@ -170,6 +170,9 @@ Options
     forced=<layer regex>
       like 'yes' above, but fail if any download fails
 
+``-h, --help``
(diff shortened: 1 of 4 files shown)


jkloetzke (commented):
The synopsis of the manpages is a verbatim copy of the respective output of `bob <command> -h | cat`. I would like to keep it this way, even if we do not show the long variant of arguments with short options. Otherwise, keeping the synopsis in sync will get increasingly harder.

icedieler:
Thanks @jkloetzke for the helpful review. In the new pull request I made sure to have both the command help and the manpage in sync and at least don't worsen the situation.

jkloetzke:
Thanks!

PR #655, opened 2025-07-23
Urlscm: add support for custom headers

rhubert:
Some servers need additional informations provided by custom headers.
Add a `headers` attribute where the user can set these custom headers.


jkloetzke:
Thanks.

Issue #462, opened 2022-04-14
Artifact download fails with HTTP redirects


jkloetzke:
The built-in HTTP artifact downloader does not handle redirects such as "302 Found" but instead treats them as soft failure. At least for downloads we should follow redirects...

PR #663, opened 2025-09-14
Webdav urllib

rhubert:


pym/bob/webdav.py
```diff
@@ -82,34 +61,49 @@ def _getHeaders(self):
...
-            if response.status == 404:
+
+        req = urllib.request.Request (self._getURL(path),
+                                      headers=headers, method="GET")
+        try:
+            resp = urllib.request.urlopen (req, context=self.__context)
+            if resp.status in [200, 206]:
+                return resp
+            else:
+                raise HttpDownloadError("{} {}".format(resp.status,
```
jkloetzke, line 97: I think `resp.close()` must be done here. But thinking more about it, I guess the case cannot happen at all. The server replied with a 2xx code to a GET request. I cannot imagine any other response than 200 and 206. So I would remove this case here.

pym/bob/archive.py
```diff
@@ -991,9 +987,6 @@ def __init__(self, archiver, response):
     def __enter__(self):
         return (None, self.response)
     def __exit__(self, exc_type, exc_value, traceback):
-        # reset connection on abnormal termination
```
jkloetzke, line 994: I *think* we must call `self.response.close()` here unconditionally. The urllib library uses `Connection: close`. So if everything goes well, the socket will be closed by the server. But if we are interrupted in the middle of a response, the connection will hang. Even if everything is read, we should probably still close the connection from our side too.

jkloetzke (commented):
Thanks a lot for the conversion. :tada: Can you add a "Fixes #462" to the commit? That long standing bug will be fixed too by the conversion.

pym/bob/webdav.py
```diff
@@ -82,34 +61,49 @@ def _getHeaders(self):
...
-        else:
-            raise HttpUploadError("HEAD {} {}".format(response.status, response.reason))
+        req = urllib.request.Request (self._getURL(path),
+                                      headers=self._getHeaders(), method="HEAD")
+        try:
+            with urllib.request.urlopen (req, context=self.__context) as response:
+                if response.status == 200:
+                    return True
+        except urllib.error.HTTPError as e:
+            e.close()
```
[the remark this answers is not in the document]
rhubert: Without this there are errors like 
```
test_webdav              | Exception ignored in: <socket.socket [closed] fd=6, family=2, type=1, proto=6>
test_webdav              | ResourceWarning: unclosed <socket.socket [closed] fd=6, family=2, type=1, proto=6>
```

:shrug:

pym/bob/webdav.py
```diff
@@ -121,55 +109,66 @@ def upload(self, path, buf, overwrite):
...
             # for the best...
-            if response.status not in [201, 405]:
-                raise HttpUploadError("MKCOL {} {}".format(response.status, response.reason))
+            if status not in [201, 405]:
+                raise HttpUploadError("MKCOL {} {}".format(status, reason))
 
     def listdir(self, path):
         base_path = self.__url.path
-        # create a full path ending with trailing / (should prevent http 301 - moved permanently)
-        path = '/'.join([base_path, path.strip('/'), ''])
```
jkloetzke, line 162: Just to be sure: can we really remove these lines? I think they are crucial...

rhubert: Looks like you are right. I did some testing against jfrog-artifactory this morning resulting in a backtrace with `bob.webdav.HttpDownloadError: PROPFIND 302 Moved Temporarily`. I can't remember why I removed this :thinking:  I restored these lines and added catches for the HTTPException in stat and listdir.

There was also a problem in  `_getURL` producing invalid paths for artifactory, e.g. `http://foo/archive//.` instead of `http://foo/archive/.` causing a 404 -> empty listdir -> backtrace. I added fixes for both issues.

PR #664, opened 2025-09-17
Jenkins clean-after-build support

jkloetzke:


jkloetzke:
> Can this be added for the local build backend as well? ;)

Interesting idea. I'm not sure if this really works. On Jenkins, the result is either uploaded or stored directly in the Jenkins job. For local builds, that's instead everything below `dev/dist`. So it pretty much boils down to `rm -rf dev/{src,build}` after the build. Not sure if this is worth a new option... :thinking:

rhubert:
> So it pretty much boils down to rm -rf dev/{src,build} after the build. Not sure if this is worth a new option... 🤔

In our environment we have the local build backend running in a gitlab-ci pipeline...on a cluster based system with pre allocated resources... Since we do not know and can not request enough resources for all possible build scenarios the build sometimes fails due to exceeded resource limits - mostly filesystem usage. Such an option could help us there I guess.

jkloetzke:
I see. That makes sense. I'm afraid this will be tricky compared to the Jenkins backend. There is no internal tracking yet of which workspaces are still required during the build.

I guess in your case you only need the final package(s) and want to discard all dependencies? Or keep all dist workspaces?

rhubert:
In the end I only need the final packages. But keeping the dist packages would also be sufficient. But this is not a very important feature - using the cache and a couple of retries is a viable workaround. It just came to my mind reading this PR ;)

PR #700, opened 2026-06-08
audit: add getters for SCMs, Files and variantId

rhubert:
Needed for sbom from audit generation


jkloetzke:
The Audit classes are not part of the public plugin API. Actually, only the classes named in the [documentation](https://bob-build-tool.readthedocs.io/en/latest/manual/extending.html#class-documentation) should be used by plugins.

So if we expose the Audit class it would have to be added to the documentation at least. But I'm not sure we should really do this. It's only a thin wrapper around the JSON data. I guess there isn't too much value to use it, is it?

PR #705, opened 2026-07-30
archive: add gitea generic package registry backend

mahaase:
Add a `gitea` archive backend that stores binary artifacts in a Gitea generic package registry. The backend is configured with the server `url`, the registry `owner` and the generic `package` name. Credentials are given explicitly as `user`/`password` (or `token`) pair for HTTP basic authentication inside the `url`, so uploads work even inside the build sandbox. Optional `sslVerify` and `retries` keys tune the transport. Also managed feature is implemented, so `bob archive` can be used to collect and remove garbage artifacts.

```
archivePrepend:
    - backend: gitea
      name: gitea
      url: "https://<YOUR_USER>:<YOUR_GITEA_TOKEN>@gitea.example.com
      owner: "owner"           # owner/org the registry belongs to
      package: "package"       # name of the generic package
      sslVerify: false
      flags: [download, upload, managed]
```

depends on https://github.com/BobBuildTool/bob/pull/706 (archive: do not leak URL credentials in user visible messages)

### DIFF
--- a/doc/manual/configuration.rst
+++ b/doc/manual/configuration.rst
@@ -2560,7 +2560,7 @@ The default is ``[download, upload]``.
 ``managed``
     This archive is managed, meaning the files can be iterated and deleted.
     This is required for the archive command to work. Only supported with the
-    ``file`` and ``http`` backends so far.
+    ``file``, ``http`` and ``gitea`` backends so far.
 ``cache``
     Use this archive to cache downloaded artifacts from other archives. If a
     binary artifact was successfully downloaded from another archive it will
@@ -2602,6 +2602,13 @@ file        Use a local directory as binary artifact repository. The directory
             directory. The optional ``fileMode`` and ``directoryMode`` keys
             take the desired access modes as numeric value to override the
             default umask derived modes.
+gitea       Uses a `Gitea generic package registry`_ as binary artifact
+            repository. The base server URL is given in ``url``, the registry
+            owner (user or organization) in ``owner`` and the generic package
+            name in ``package``. As with the ``http`` backend the credentials
+            are part of the URL. The optional ``sslVerify`` boolean key
+            controls whether to verify the SSL certificate and ``retries`` sets
+            the number of retries on transient errors.
 http        Uses a HTTP server as binary artifact repository. The server has to
             support the HEAD, PUT and GET methods. The base URL is given in the
             ``url`` key. The optional ``sslVerify`` boolean key controls
@@ -2679,6 +2686,29 @@ the anonymous access to the container can be used like this::
 
 The ``flags: [download]`` makes sure that Bob does not try to upload artifacts
 in case other backends are configured too.
+
+The ``gitea`` backend stores the artifacts in a Gitea generic package registry.
+The artifacts are put below ``{url}/api/packages/{owner}/generic/{package}/``
+with one package version per artifact::
+
+    archive:
+        -
+            backend: gitea
+            url: "https://user:passw%40rd@gitea.example.com"
+            owner: "bob-artifacts"
+            package: "myproject"
+            flags: [download, upload, managed]
+
+As for the ``http`` backend the credentials for the HTTP basic authentication
+are taken from the URL and have to be percent encoded. Gitea accepts a personal
+access token instead of the password. Uploading requires the ``write:package``
+scope, the ``managed`` flag additionally needs ``read:package``.
+
+.. warning::
+   The password will be part of the Jenkins job configuration. Anybody who can
+   read the jobs ``config.xml`` will be able to retrieve the password!
+
+.. _Gitea generic package registry: https://docs.gitea.com/usage/packages/generic
 
 .. _configuration-config-archive-prepend-append:
 
--- a/pym/bob/archive.py
+++ b/pym/bob/archive.py
@@ -862,6 +862,16 @@ class LocalArchiveUploader:
         return False
 
 
+def retryWebdavRequest(request, retries):
+    """Run a WebDav request, retrying transient transport errors."""
+    while True:
+        try:
+            return request()
+        except (WebdavError, OSError) as e:
+            if retries == 0: raise
+            retries -= 1
+
+
 class HttpArchive(BaseArchive):
     def __init__(self, spec):
         super().__init__(spec)
@@ -878,13 +888,7 @@ class HttpArchive(BaseArchive):
         return urllib.parse.urlunparse((url.scheme, getNetLoc(url), url.path, '', '', ''))
 
     def __retry(self, request):
-        retries = self._retries
-        while True:
-            try:
-                return request()
-            except (WebdavError, OSError) as e:
-                if retries == 0: raise
-                retries -= 1
+        return retryWebdavRequest(request, self._retries)
 
     def _canManage(self):
         return True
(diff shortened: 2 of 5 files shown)


mahaase:
i'm not sure, but the root cause of the failed windows build could be:
`Error: Failed to download: [Errno 28] Disk full`

jkloetzke:
Thanks. I'll have a look after my vacation...

doc/manual/configuration.rst
```diff
@@ -2602,6 +2602,14 @@ file        Use a local directory as binary artifact repository. The directory
             directory. The optional ``fileMode`` and ``directoryMode`` keys
             take the desired access modes as numeric value to override the
             default umask derived modes.
+gitea       Uses a `Gitea generic package registry`_ as binary artifact
+            repository. The base server URL is given in ``url``, the registry
+            owner (user or organization) in ``owner`` and the generic package
+            name in ``package``. Credentials are given explicitly in the
+            ``token`` key (a personal access token) or as ``user``/``password``
```
jkloetzke: Hmm, why not use the established way of encoding the username/password for HTTP basic authentication in the URL? The "http" backend already does it this way...

pym/bob/archive.py
```diff
@@ -1211,6 +1217,236 @@ def __upload(self):
...
+                if retries == 0:
+                    raise ArtifactError(str(e))
+                retries -= 1
+
+    def _openDownloadFile(self, buildId, suffix):
+        url = self._remoteName(buildId, suffix)
+        return self.__retry(lambda: self.__openDownloadFile(url))
+
+    def __openDownloadFile(self, url):
+        req = urllib.request.Request(url, headers=self.__headers(), method="GET")
```
jkloetzke: Instead of using urllib directly, the `Webdav` class should be used. If there is something missing in this class, feel free to extend it.

pym/bob/archive.py
```diff
@@ -1211,6 +1217,236 @@ def __upload(self):
...
+        return GiteaDownloader(rsp)
+
+    def __exists(self, url):
+        # Cheap existence/authentication probe before uploading a possibly large
+        # artifact. Only the first byte is requested. This surfaces an
+        # authentication problem as a clean 401 instead of an opaque connection
+        # reset that would otherwise occur when the server drops the connection
+        # while we are still sending the body.
+        req = urllib.request.Request(url,
+            headers=self.__headers({'Range': 'bytes=0-0'}), method="GET")
```
jkloetzke: This looks very odd. Don't they support the `HEAD` request?

pym/bob/archive.py
```diff
@@ -1211,6 +1217,236 @@ def __upload(self):
...
+                raise   # transient server error -> retryable
+            raise ArtifactError("GET {} {}".format(e.code, e.reason))
+
+    def _openUploadFile(self, buildId, suffix, overwrite):
+        url = self._remoteName(buildId, suffix)
+        if overwrite:
+            # Gitea's generic registry refuses to overwrite an existing file.
+            # Metadata files (build-id, fingerprint) must be replaced, so delete
+            # a possibly existing file first (a missing one is ignored).
+            self.__retry(lambda: self.__delete(url))
```
jkloetzke: I guess we can't do much about this but it's inherently racy. A racing upload can still fail because another upload hit the time frame between delete and upload.

jkloetzke (commented):
Can you split the change into multiple commits? At least one for the implementation, one for the documentation and one for the tests? Also, if you need to extend the `Webdav` class, that should be a separate commit as well.

doc/manual/configuration.rst
mahaase: You're right, that's the better way — the explicit keys are gone. `token`, `user` and `password` were dropped from the schema and the credentials are now taken from the URL, exactly like the http backend does.

pym/bob/archive.py
mahaase: Done — `GiteaArchive` is now built on top of `WebDav` and shrank from ~240 to
~70 lines. The hand-written urllib layer is gone, and with it the custom error
translation: `BaseArchive` already handles `WebdavNotFoundError`,
`WebdavAlreadyExistsError` and `WebdavError` everywhere. `GiteaDownloader` and
`GiteaUploader` are deleted too, since `HttpDownloader`/`HttpUploader` are
backend agnostic. The duplicated retry loop is now a shared function used by
both backends.

Two extensions to `WebDav` were needed (separate commit):

- the generic registry does not honour `If-None-Match` and reports an existing
  file with a 409, so that is now treated like a 412
- `deletePath()` for deleting by absolute path, because the gitea paths are
  built from the package API layout. `delete()` keeps its relative filename and
  delegates.

One behavioural change worth pointing out: the custom "authentication failed"
message is gone. It only existed because the old code had no preflight and hit
an opaque connection reset in the middle of the body upload. Now that a HEAD
(or DELETE, when overwriting) always precedes the PUT, a bad credential fails
cleanly as `HEAD 401 Unauthorized`, same as the http backend. The hint about the
required `write:package` token scope moved to the documentation.

pym/bob/archive.py
mahaase: They do — Gitea registers `r.Methods("HEAD,GET", "", generic.DownloadPackageFile)`
 for the generic package download route. I had not checked that properly. The
 Range hack is removed and the existence check is just `WebDav.exists()` now.

pym/bob/archive.py
mahaase: Agreed, and I don't see a way around it: the generic registry has no
conditional or atomic replace, so the file has to be deleted first. A racing
upload will lose with a 409, which maps to `WebdavAlreadyExistsError` and is
handled like any other "already exists" case.

Note this only affects the metadata files (build-id, fingerprint) that are
uploaded with `overwrite=True`; the artifact itself is never overwritten. I've
made the code comment state the race explicitly instead of glossing over it.

mahaase:
@jkloetzke Thanks for your review and feedback! 

Hopefully everything should be addressed. The change is now split into four commits:

1. `webdav: support conflict replies and deleting by path` — the WebDav
   extensions, isolated as requested
2. `archive: add gitea generic package registry backend` — the implementation
3. `doc: document the gitea archive backend`
4. `test: cover the gitea archive backend`

mahaase:
disk full issue again.

pym/bob/archive.py
```diff
@@ -1213,6 +1217,89 @@ def __upload(self):
...
+        name = super().getArchiveName()
+        if name:
+            return name
+        return urllib.parse.urlunparse((self.__url.scheme, self.__url.netloc,
+            self.__basePath(), '', '', ''))
+
+    def _canManage(self):
+        # Managed operations (scan/clean) would require the Gitea package list
+        # API. Not implemented yet.
+        return False
```
jkloetzke: I'm wondering it it would make sense to use one of the other supported repository options. For example, the "Conan Package Registry" seems to support to search for artifacts...

jkloetzke (commented):
This seems to have the same commits from #706. Maybe we should get that one merged first.

Please make the "webdav: fix pickling and support conflict replies and deletion by path" three commits. It contains three different things.

Also, "archive: strip URL credentials in the gitea backend too" should be squashed into the commit that creates the gitea backend.

jkloetzke:
> @jkloetzke Thanks for your review and feedback!
> 
> Hopefully everything should be addressed. The change is now split into four commits:
> 
>     1. `webdav: support conflict replies and deleting by path` — the WebDav
>        extensions, isolated as requested
> 
>     2. `archive: add gitea generic package registry backend` — the implementation
> 
>     3. `doc: document the gitea archive backend`
> 
>     4. `test: cover the gitea archive backend`

Please, don't let me talk to the AI. This actually feels kind of rude. In fact, it would be way faster if I'd do that myself directly. But the point of such a review is not just to get the code into shape. The more important part is to let the contributors know how things should be done. You should actually learn during the review so that future contributions need much less review comments.

If I'm just talking to the Claude directly, this knowledge is probably not generated. At least this is what I have to assume. My fear is that is not a sustainable approach in the long run. Claude can generate changes like this PR quickly. But I need you as a contributor to make the first review. At least that is my hope for the future to keep the maintenance manageable.

mahaase:
> Please, don't let me talk to the AI. This actually feels kind of rude. In fact, it would be way faster if I'd do that myself directly. But the point of such a review is not just to get the code into shape. The more important part is to let the contributors know how things should be done. You should actually learn during the review so that future contributions need much less review comments.
> 
> If I'm just talking to the Claude directly, this knowledge is probably not generated. At least this is what I have to assume. My fear is that is not a sustainable approach in the long run. Claude can generate changes like this PR quickly. But I need you as a contributor to make the first review. At least that is my hope for the future to keep the maintenance manageable.

That's fair and I take your point. Nothing here is posted by a bot. I send everything myself, but the last comment included a summary written by AI and I just pasted it. To be open about it: the code is AI-assisted. Without that I couldn't contribute PRs of this size and complexity at all. Everything I send is tested and working for our use-cases. What I can't always judge is whether a change is the right fit for bob. For example that the webdav class should be used instead of urllib directly. This is something I would never have found on my own. That's the part I take away from your reviews and it sticks. We build this for our own use and upstream it because we'd rather see bob improve. We intend to stay, so I'd like to find a way of working that's sustainable for you as well.

mahaase:
update: i tested the managed feature. it looks pretty well as far as i can evaluate it.

pym/bob/webdav.py
```diff
@@ -78,10 +78,12 @@ def _getHeaders(self):
                 userPass.encode("utf-8")).decode("ascii")
         return headers
 
-    def _getURL(self, path):
+    def _getURL(self, path, query=None):
         # remove username and password from URI
+        if query is None:
+            query = self.__url.query
```
jkloetzke, line 84: Should be moved before the above comment. The comment applies to `getNetLoc` below, not to the new `if` statement.

doc/manual/configuration.rst
```diff
@@ -2602,6 +2602,14 @@ file        Use a local directory as binary artifact repository. The directory
             directory. The optional ``fileMode`` and ``directoryMode`` keys
             take the desired access modes as numeric value to override the
             default umask derived modes.
+gitea       Uses a `Gitea generic package registry`_ as binary artifact
+            repository. Only the HEAD, GET, PUT and DELETE methods are used.
```
jkloetzke, line 2606: I think the enumeration of the used methods is just confusing and should be removed. I guess Claude did it like the HTTP backend. But there this is a crucial information because it puts a requirement on the HTTP server.

doc/manual/configuration.rst
```diff
@@ -2680,6 +2688,37 @@ the anonymous access to the container can be used like this::
...
+   The password will be part of the Jenkins job configuration. Anybody who can
+   read the jobs ``config.xml`` will be able to retrieve the password!
+
+.. note::
+   The registry cannot enumerate its content. The managed operations of
+   :ref:`manpage-archive` therefore use the package API of the server
+   (``{url}/api/v1/packages/...``) to list the artifacts. As with the ``http``
+   backend only the artifact itself is deleted, the build-id and fingerprint
+   files are left behind. The package version is dropped by the server as soon
+   as its last file is gone.
```
jkloetzke, line 2718: I would delete this note. It's describing an implementation detail and is not relevant to the user of the backend.

jkloetzke:
> > Please, don't let me talk to the AI. This actually feels kind of rude. In fact, it would be way faster if I'd do that myself directly. But the point of such a review is not just to get the code into shape. The more important part is to let the contributors know how things should be done. You should actually learn during the review so that future contributions need much less review comments.
> > If I'm just talking to the Claude directly, this knowledge is probably not generated. At least this is what I have to assume. My fear is that is not a sustainable approach in the long run. Claude can generate changes like this PR quickly. But I need you as a contributor to make the first review. At least that is my hope for the future to keep the maintenance manageable.
> 
> That's fair and I take your point. Nothing here is posted by a bot. I send everything myself, but the last comment included a summary written by AI and I just pasted it.

Thanks for the clarification.

> To be open about it: the code is AI-assisted. Without that I couldn't contribute PRs of this size and complexity at all. Everything I send is tested and working for our use-cases. What I can't always judge is whether a change is the right fit for bob. For example that the webdav class should be used instead of urllib directly. This is something I would never have found on my own. That's the part I take away from your reviews and it sticks. We build this for our own use and upstream it because we'd rather see bob improve. We intend to stay, so I'd like to find a way of working that's sustainable for you as well.

First of all: thanks for putting the effort into this and upstreaming things. This work is very much appreciated. :pray:

In the end, my only wish is that you try to dive into the generated code so that you can be able to maintain it. Especially with features that have initially no other user and that are hard to test by myself, I need to rely on you to address any issues in the future. As long as you are able to do that, I'm fine with merging AI generated code. We just have to prevent the situation where neither you nor I understand the code. :sweat_smile:

jkloetzke (commented):
This looks like it could be merged soon. Just some small comments. You also have to rebase because some commits of the #706 PR are still part of this PR...

pym/bob/webdav.py
mahaase: the comment was removed while rebase, because `getNetLoc()` has its own documentation.

doc/manual/configuration.rst
mahaase: done.

doc/manual/configuration.rst
mahaase: done.

jkloetzke:
Thanks!

PR #706, opened 2026-08-19
archive: do not leak URL credentials in user visible messages

mahaase:
## Problem

The HTTP basic authentication credentials of the `http` archive backend are
part of the URL, as documented. All three places that turn that URL back into
a string for display use the raw `netloc`, which still carries the
`user:password@` part:

| | leaks via | visible |
|---|---|---|
| `getArchiveName()` (`archive.py:877`) | `_namedErrorString()` (`archive.py:401`) | always, no flag needed |
| `_remoteName()` (`archive.py:902`) | `details` of the `DOWNLOAD`/`UPLOAD`/`MAP-SRC`/`CACHE-BID`/`CACHE-FPR`/`MAP-FPRNT` status lines | with `-v` |
| `getArchiveUri()` (`archive.py:966`) | `bob archive` (`cmds/archive.py:32`) | always |

The first one is the bad one. `_namedErrorString()` is used for *every* error
message, including the perfectly ordinary "artifact not found" that happens for
each package on a cache miss. So a plain `bob build` on a cold cache prints the
password once per package:

```
>> aarch64::linux/devel::sandbox/core::coreutils
   DOWNLOAD  work/core/coreutils/.../workspace .. https://user:s3cr3t@server.test/artifacts: artifact not found
```

These messages routinely end up in build logs, CI consoles and pasted-into-an-issue
snippets.

The optional `name` archive setting is not a workaround: `_remoteName()` and
`getArchiveUri()` never consult it.

Note this is a display-only issue. `WebDav._getURL()` already strips the
credentials before the URL goes on the wire; they are sent in the
`Authorization` header (`webdav.py:47-55`).

## Fix

Factor the stripping that `WebDav._getURL()` already does into a
`stripUserInfo()` helper and use it in the three places above as well.

It takes and returns a parsed URL and preserves the result type, so both the
`urlsplit()` result used in `webdav.py` and the `urlparse()` result used in
`archive.py` work.

### Drive-by: cut at the last `@`, not the first

The existing implementation used `netloc.split('@')[1]`. `urlparse()` delimits
the user info at the **last** `@`, so the two disagree when the password
contains an unencoded `@`:

```python
>>> urlparse('https://u:p@ss@host:8443/x').netloc
'u:p@ss@host:8443'
>>> _.split('@')[1]        # before: bogus host, and part of the password survives
'ss'
>>> _.rsplit('@', 1)[1]    # after
'host:8443'
```

Percent encoding is documented as required, so this was latent rather than a
live bug, but there is no reason to get it wrong.

## Tests

- `TestHttpBasicAuthArchive.testNoCredentialsInMessages` — asserts all three
  methods drop the password and the `@`, and still contain the host.
- `TestStripUserInfo` — no credentials (identity), removal, `@` in the
  password, IPv6 literal, `urlsplit()` type preservation.

Reverting `archive.py` alone makes the first test fail on all three subtests;
`test_archive` + `test_webdav` are green with the fix (71 tests).

### DIFF
--- a/pym/bob/webdav.py
+++ b/pym/bob/webdav.py
@@ -28,7 +28,9 @@ def getNetLoc(url):
     """
     netloc = url.netloc
     if url.username is not None:
-        netloc = netloc.split('@')[1]
+        # urlparse() delimits the user info at the *last* '@'. Cut at the same
+        # one, otherwise a password with an unencoded '@' yields a bogus host.
+        netloc = netloc.rsplit('@', 1)[1]
 
     return netloc
 
--- a/pym/bob/webdav.py
+++ b/pym/bob/webdav.py
@@ -18,6 +18,21 @@ class WebdavNotFoundError(WebdavError):
 class WebdavAlreadyExistsError(WebdavError):
     pass
 
+
+def getNetLoc(url):
+    """Get the network location of a parsed URL without the credentials.
+
+    The user name and password of the HTTP basic authentication are part of
+    the URL. They are sent in the Authorization header instead and must be
+    kept out of the request URL and of anything that is shown to the user.
+    """
+    netloc = url.netloc
+    if url.username is not None:
+        netloc = netloc.split('@')[1]
+
+    return netloc
+
+
 class WebDav:
 
     class PartialDownloader:
@@ -55,12 +70,7 @@ class WebDav:
         return headers
 
     def _getURL(self, path):
-        # remove username and password from URI
-        netloc = self.__url.netloc
-        if self.__url.username is not None:
-            netloc = self.__url.netloc.split('@')[1]
-
-        return urlunsplit((self.__url.scheme, netloc, path,
+        return urlunsplit((self.__url.scheme, getNetLoc(self.__url), path,
                            self.__url.query, self.__url.fragment))
 
     def exists(self, path):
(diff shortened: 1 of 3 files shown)


pym/bob/webdav.py
```diff
@@ -18,6 +18,25 @@ class WebdavNotFoundError(WebdavError):
...
+    They are sent in the ``Authorization`` header instead.
+
+    Takes and returns a :func:`urllib.parse.urlsplit` or
+    :func:`urllib.parse.urlparse` result. The type is preserved.
+    """
+    if url.username is None:
+        return url
+    # rsplit: urlparse() treats the *last* '@' as the delimiter, so a password
+    # with an unencoded '@' must not confuse us into cutting at the first one.
+    return url._replace(netloc=url.netloc.rsplit('@', 1)[1])
```
jkloetzke: AFAICT, this is an undocumented, private method. Just do it like `_getURL()` previously.

jkloetzke (commented):
Please make this two commits for the fixes. One that prunes the credentials from the logs and one that splits the username/password and host name differently. These are two independent changes that should get their own commit.

The documentation commit can be dropped. Users of Bob probably don't care how HTTP basic auth works.

test/unit/test_archive.py
```diff
@@ -25,7 +25,7 @@
 from bob.archive import DummyArchive, HttpArchive, getArchiver
 from bob.errors import BuildError
 from bob.utils import runInEventLoop, getProcessPoolExecutor
-from bob.webdav import WebdavError
+from bob.webdav import WebdavError, stripUserInfo
```
jkloetzke: In the first commit, the imported function is called `stripUserInfo`, which is not defined yet. That will potentially break git bisect. The last commit fixes the test. So the last commit should be squashed.

It's also better to separate tests into a dedicated commit. Test code should only be touched if a code change also requires adaptations in unit tests.

jkloetzke:
Thanks!

